PR #14, opened 2021-05-07
Handle attachments and reactions for messages

SarahIsWeird:
Discord sends a list of all attachments in the JSON when transmitting a message object. Similarly, under certain circumstances a list of all reactions is transmitted. This PR adds both infos to the `dpp::message` object.

Firstly, there's a new attachment struct. It closely follows the [official docs](https://discord.com/developers/docs/resources/channel#attachment-object) on what's present and what not.
Secondly, I've added one field to dpp::reaction, the emoji name. Discord sends it, and for non-custom reactions, this is often the only way to figure out which emoji was used to react.

Both are populated accordingly.

### DIFF
--- a/include/dpp/message.h
+++ b/include/dpp/message.h
@@ -288,6 +288,62 @@ struct reaction {
 	bool me;
 	/** ID of emoji for reaction */
 	snowflake emoji_id;
+	/** Name of emoji, if applicable */
+	std::string emoji_name;
+
+	/**
+	 * @brief Constructs a new reaction object.
+	 */
+	reaction();
+
+	/**
+	 * @brief Constructs a new reaction object from a JSON object.
+	 * @param j The JSON to read data from
+	 */
+	reaction(nlohmann::json* j);
+
+	/**
+	 * @brief Destructs the reaction object.
+	 */
+	~reaction() = default;
+};
+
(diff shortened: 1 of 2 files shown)


SarahIsWeird:
This is an example file that showcases both new features.
```cpp
#include <dpp/dpp.h>

int main() {
    dpp::cluster bot("token");

    bot.on_message_create([&bot](const dpp::message_create_t &event) {
        if (event.msg->attachments.size() == 0)
            return;

        bot.message_create(dpp::message(event.msg->channel_id, "Your message has attachments!"));
        // Do stuff with attachment url, width and height, ...
    });

    bot.on_message_reaction_add([&bot](const dpp::message_reaction_add_t &event) {
        bot.message_get(event.message_id, event.reacting_channel->id, [](const dpp::confirmation_callback_t &cb) {
            for (auto &reaction : ((dpp::message &) cb.value).reactions) {
                // Yes, this isn't that good of an idea. The emoji name doesn't actually need to be
                // populated. But it's just so you get the idea. :D
                std::cout << reaction.count << " people reacted with " << reaction.emoji_name << "." << std::endl;
            }
        });
    });

    bot.start(false);

    return 0;
}
```

Issue #18, opened 2021-05-17
Error build slashcommand json with sub-commands and options


d-nery:
**Describe the bug**
Trying to use sub-commands with options in slash commands causes the `build_json` method to behave unexpectedly, replicating the sub-command in it's options, causing nested `type: 1` options, which is not supported by discord.

**To Reproduce**
Example program:
```cpp
int main() {
    dpp::slashcommand command;

    command.set_name("cmd name")
        .set_description("cmd desc")
        .set_application_id(1234567)
        .add_option(dpp::command_option(dpp::co_sub_command, "sub1", "sub1 desc")
                        .add_option(dpp::command_option(dpp::co_string, "option 1", "sub1 option1 desc")
                                        .add_choice(dpp::command_option_choice("choice1", std::string("choice1")))
                                        .add_choice(dpp::command_option_choice("choice2", std::string("choice2"))))
                        .add_option(dpp::command_option(dpp::co_integer, "option 2", "sub1 option2 desc")))
        .add_option(dpp::command_option(dpp::co_sub_command, "sub2", "sub2 desc"));

    std::string json_dump = command.build_json(false);

    std::cout << json::parse(json_dump).dump(4) << std::endl;
}
```

Generated output:

```json
{
    "description": "cmd desc",
    "name": "cmd name",
    "options": [
        {
            "description": "sub1 desc",
            "name": "sub1",
            "options": [
                {
                    "description": "sub1 desc",
                    "name": "sub1",
                    "options": [],
                    "required": false,
                    "type": 1
                },
                {
                    "description": "sub1 desc",
                    "name": "sub1",
                    "options": [
                        {
                            "description": "sub1 desc",
                            "name": "sub1",
                            "options": [],
                            "required": false,
                            "type": 1
                        }
                    ],
                    "required": false,
                    "type": 1
                }
            ],
            "required": false,
            "type": 1
        },
        {
            "description": "sub2 desc",
            "name": "sub2",
            "required": false,
            "type": 1
        }
    ]
}
```

**Expected behavior**
This is the expected output:

```json
{
    "description": "cmd desc",
    "name": "cmd name",
    "options": [
        {
            "description": "sub1 desc",
            "name": "sub1",
            "options": [
                {
                    "choices": [
                        {
                            "name": "choice1",
                            "value": "choice1"
                        },
                        {
                            "name": "choice2",
                            "value": "choice2"
                        }
                    ],
                    "description": "sub1 option1 desc",
[...]

PR #22, opened 2021-05-21
Implement json deserializers for interaction events

d-nery:
This fixes a similar issue to #18 #19 but the other way around, received interaction events had their sub-options ignored and not parsed, this used the json's library built-in `from_json` to make this deserialization easier.

The original `::fill_from_json` methods are still there and just call the automatic conversion from the library, for compatibility, also, I onlyu implemented this for the structs involved in the interatcion_create event, the others are still parsed as they used to be.

### DIFF
--- a/include/dpp/discordevents.h
+++ b/include/dpp/discordevents.h
@@ -2,7 +2,7 @@
  *
  * D++, A Lightweight C++ library for Discord
  *
- * Copyright 2021 Craig Edwards and D++ contributors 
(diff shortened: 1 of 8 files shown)


webtax-gh (approved):
lgtm

braindigitalis (commented):
src/dpp/guild.cpp:102
not sure this brace is supposed to be here?
it isn't supposed to skip the rest if no guild id...?

PR #24, opened 2021-05-23
Fix embed for interaction responses

d-nery:
Interaction responses are not normal messages and has a difference on the embed field, this fixes it


janaSunrise:
LGTM 👍🏼

PR #25, opened 2021-05-24
Fix: timestamp not part of build_json output

d-nery:
Library was not adding the timestamp field to the json object


webtax-gh:
This seems to set timestamp as a string. Looking at the header file it seems that it should be `time_t`?

webtax-gh (approved):
fixes build_json, lgtm

PR #26, opened 2021-05-24
Fix OpenSSL for Windows

smhman:
This commit fixes errors about random missing symbols (or redefined) on OpenSSL headers. This has been tested and works on my machine.


janaSunrise:
LGTM 👍🏼

PR #30, opened 2021-06-04
interaction_create_t::get_parameter returns reference to value on the stack

Sirraide:
While working with long (>16 chars) slash command parameters, I ran into some issues with parts of the data getting corrupted, and after some examination the problem seems to be the following:

`std::get` in `command_interaction ci = std::get(...` returns a `const command_interaction&`, but the assignment to a non-reference type creates a local copy on the stack. The `const command_value&` returned by the function is therefore a reference to a local object. Changing the type of `ci` from `command_interaction` to `const command_interaction&` seems and ought to have fixed the problem, as it preserves the reference returned by `std::get` instead of creating a local copy.


braindigitalis:
thanks!

PR #33, opened 2021-06-10
added a guild_member_move method

mobergmann:
I added a `guild_member_move` method, which moves a user to another channel (if the user is connected to a channel).

### DIFF
--- a/include/dpp/cluster.h
+++ b/include/dpp/cluster.h
@@ -1184,10 +1184,11 @@ public:
 	/**
 	 * @brief Moves the guild member to a other voice channel, if member is connected to one
 	 * @param channel_id Id of the channel to which the user is used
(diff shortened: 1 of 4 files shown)


mobergmann:
I'm not happy with the methods header: `void cluster::guild_member_move(const snowflake channel_id, const guild_member &gm, command_completion_event_t callback)`. Maybe we could change the `const guild_member &gm` to two parameters (user_id and `guild_id`) or a `std::pair`? In the method, we don't need all information of a `guild_member`, but it semantically makes sence.

braindigitalis:
i agree it should be changed so that the signature is snowflake channel_id and snowflake user_id with optional callback. Please change to be like this as the signature is simpler and as you say the guild_member is unused apart from its user id and not required :)

PR #35, opened 2021-06-15
caching improvements and lazy-caching mode

braindigitalis:
This PR adds caching improvements plus lazy-caching mode `dpp::cp_lazy` which massively reduces memory usage, only caching users when they message the bot.


braindigitalis:
also adds guild_member_move from mobergmann pr #33

PR #37, opened 2021-06-19
Fixes for Visual Studio in Windows and others

luizfernandonb:
Now it is no longer necessary to run `build.bat` to create a solution for Visual Studio, just open the folder and have the C++ development tools installed in the Visual Studio Installer (`CMakeLists.txt` has been modified to accept use in Visual Studio 2019 and likely later (if nothing changes))

I changed from `_WIN32` to `WIN32`, it's more pleasing to the eye and also gives the same :p

Some POSIX standard functions give errors when called on Windows, not because it's not supported, but because the compiler imposes you to use a better one, which in this case is the `_close()` function instead of `close()`, it doesn't matter, however you can't disable it on Windows

All macro definitions have been stripped from the code and placed in `CMakeLists.txt` for the entire project to avoid future macro reset warnings

I changed some flags that are used for GCC/Clang and added a specific check for MSVC because it uses `/` as a prefix and most of the options that exist in GCC/Clang don't exist in it


braindigitalis (changes requested):
is cmake 3.15 needed? can we keep dependency on 3.12 otherwise we lock out a ton of Linux systems.

this will not work on Debian stable which has 3.13.4 and is pretty much one of the most common Linux environments (and is also my production system).

Also does build dynamic libraries still default to yes?

looking good so far!

luizfernandonb:
> is cmake 3.15 needed? can we keep dependency on 3.12 otherwise we lock out a ton of Linux systems.

Ok, I'll go back to 3.12, I changed it to 3.15 because there was an annoying warning when I was compiling with MSVC, [this ](https://stackoverflow.com/questions/58708772/cmake-project-in-visual-studio-gives-flag-override-warnings-command-line-warnin) one, but anyway, this warning doesn't hurt

> Also does build dynamic libraries still default to yes?

It was only for Windows, I had placed it by mistake and now I fixed it and it's enabled for all systems

braindigitalis (approved):

braindigitalis:
thanks, Ive merged this! how does this affect the "building on windows" tutorial please? does that now need redoing?

luizfernandonb:
> thanks, Ive merged this! how does this affect the "building on windows" tutorial please? does that now need redoing?

Yes, as you no longer need build.bat, you just need to clone the repository, open an instance of Visual Studio and click on this option

image

Which in English is: `Open a local folder`

And then a Windows Explorer window will open to select the folder you want and just select the `DPP` folder (which is the cloned repository) and click and `Select folder`

image

What if you have the C++ development tools installed via Visual Studio Installer

image

Along with CMake tools from C++ for Windows

image

Everything should work perfectly :)

PR #41, opened 2021-06-19
fix block code in documentation

Bricklou:
Block code have some rendering issue in the documentation (file: `docpages/02_build.md`):
image


webtax-gh:
lgtm

PR #42, opened 2021-06-26
[🚧 WIP 🚧] 🎉 DPP Threads Support

webtax-gh:
Threads are now partially added to DPP. They are basically channels, but with these new keys: 
image

Todo list:
[ ] Add thread specific gw events (currently threads are only imported through `guild_create`)
[ ] Add thread creation utils
[ ] Add thread REST utils

### DIFF
--- a/include/dpp/channel.h
+++ b/include/dpp/channel.h
@@ -33,28 +33,37 @@ namespace dpp {
 #define GUILD_CATEGORY	4	//!< an organizational category that contains up to 50 channels
 #define GUILD_NEWS	5	//!< a channel that users can follow and crosspost into their own server
 #define GUILD_STORE	6	//!< a channel in which game developers can sell their game on Discord
+#define GUILD_NEWS_THREAD 10 //!< a temporary sub-channel within a GUILD_NEWS channel
+#define GUILD_PUBLIC_THREAD 11 //!< a temporary sub-channel within a GUILD_TEXT channel
+#define GUILD_PRIVATE_THREAD 12 //!< a temporary sub-channel within a GUILD_TEXT channel that is only viewable by those invited and those with the MANAGE_THREADS permission
 #define GUILD_STAGE	13	//!< a "stage" channel, like a voice channel with one authorised speaker
 
 /** @brief Our flags as stored in the object */
 enum channel_flags {
 	/// NSFW Gated Channel
-	c_nsfw =		0b00000001,
+	c_nsfw =			0b00000001,
 	/// Text channel
-	c_text =		0b00000010,
+	c_text =			0b00000010,
 	/// Direct Message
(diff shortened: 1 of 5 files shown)


src/dpp/events/guild_create.cpp
```diff
@@ -83,6 +83,20 @@ void guild_create::handle(discord_client* client, json &j, const std::string &ra
 			g->channels.push_back(c->id);
 		}
 
+		/* Store guild threads */
+		g->threads.clear();
+		g->threads.reserve(d["threads"].size());
+		for (auto & channel : d["threads"]) {
+			dpp::channel* c = dpp::find_channel(SnowflakeNotNull(&channel, "id"));
```
braindigitalis, line 90: caching here needs to be optional, note without the caching we dont need the `new` at all

include/dpp/channel.h
```diff
@@ -123,6 +147,15 @@ class channel : public managed {
 	/** Permission overwrites to apply to base permissions */
 	std::vector<permission_overwrite> permission_overwrites;
 
+	/** Approximate count of messages in a thread (threads) */
+	uint8_t message_count;
```
braindigitalis, line 151: is this really big enough? perhaps uint32_t?

include/dpp/channel.h
```diff
@@ -123,6 +147,15 @@ class channel : public managed {
...
+
+	/** Approximate count of members in a thread (threads) */
+	uint8_t member_count;
```
braindigitalis, line 154: Is this big enough? does discord allow > 255 members in a thread? perhaps uint16_t?

include/dpp/channel.h
```diff
@@ -123,6 +147,15 @@ class channel : public managed {
 	/** Permission overwrites to apply to base permissions */
 	std::vector<permission_overwrite> permission_overwrites;
 
+	/** Approximate count of messages in a thread (threads) */
+	uint8_t message_count;
+
+	/** Approximate count of members in a thread (threads) */
```
braindigitalis, line 153: Remember to init both of these to 0 in channel::channel

braindigitalis (changes requested):

include/dpp/channel.h
webtax-gh: yes, it's big enough 
image

include/dpp/channel.h
webtax-gh: yes, it's big enough 
image

include/dpp/channel.h
webtax-gh: resolved

braindigitalis:
hi how far did this get, I was holding off on the next release until we had threads support

braindigitalis:
continued and merged as part of PR#46

PR #44, opened 2021-07-10
Multiple embeds in normal message AND New embed footer member functions 

Axyte:


include/dpp/message.h
```diff
@@ -250,6 +250,13 @@ struct embed_footer {
 	std::string icon_url;
 	/** Proxied icon url */
 	std::string proxy_url;
+
+	/**Set footer's text**/
+    embed_footer& set_text(const std::string& t);
```
smhman, line 255: not trying to be nippy picky, but erm.. these spaces

Axyte: Sorry for that, I noticed them after the pull request.

PR #46, opened 2021-07-16
Finished implementing editing of slash command interaction responses

Sirraide:
This updates `cluster::interaction_response_edit` to use the proper api endpoint (`/api/v8/webhooks`) and format in accordance with https://discord.com/developers/docs/interactions/slash-commands#edit-original-interaction-response. Some unnecessary parameters were removed. Also added `interaction_create_t::edit_response`.

### DIFF
--- a/include/dpp/cluster.h
+++ b/include/dpp/cluster.h
@@ -767,12 +767,11 @@ public:
 
 	/**
 	 * @brief Respond to a slash command
-	 * 
-	 * @param interaction_id Interaction id to respond to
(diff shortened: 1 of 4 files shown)


braindigitalis:
Merged - does this relate to https://github.com/discord/discord-api-docs/issues/2410 ?

Sirraide:
I think so. I implemented this because I wanted to be able to acknowledge an interaction and send a response later, which is now possible.

PR #48, opened 2021-07-22
Thread GW events and REST utils

Axyte:


include/dpp/channel.h
```diff
@@ -81,6 +90,34 @@ struct permission_overwrite {
...
+
+/**
+ * @brief represents membership of a user with a thread
+ */
+struct thread_member 
+{
+	snowflake thread_id;
+	snowflake user_id;
+	time_t joined;
+	int32_t flags;
```
braindigitalis, line 116: possibly uint32_t? no need for a sign bit on bitfields

include/dpp/channel.h
```diff
@@ -123,6 +160,15 @@ class channel : public managed {
 	/** Permission overwrites to apply to base permissions */
 	std::vector<permission_overwrite> permission_overwrites;
 
+	/** Approximate count of messages in a thread (threads) */
+	uint8_t message_count;
+
+	/** Approximate count of members in a thread (threads) */
+	uint8_t member_count;
```
braindigitalis, line 167: i already know these oddball sizes are correct - queried originally in webtax's version :)

include/dpp/cluster.h
```diff
@@ -747,6 +747,21 @@ class cluster {
 	 */
 	void on_integration_delete (std::function<void(const integration_delete_t& _event)> _integration_delete);
 
+	/**
```
braindigitalis, line 750: would you like for me to add comments for these events after its merged, or are you wanting to do this now? We'd need comments either way :)

src/dpp/events/guild_create.cpp
```diff
@@ -85,6 +85,20 @@ void guild_create::handle(discord_client* client, json &j, const std::string &ra
 			g->channels.push_back(c->id);
 		}
 
+		/* Store guild threads */
+		g->threads.clear();
+		g->threads.reserve(d["threads"].size());
+		for (auto & channel : d["threads"]) {
+			dpp::channel* c = dpp::find_channel(SnowflakeNotNull(&channel, "id"));
+			if (!c) {
+				c = new dpp::channel();
```
braindigitalis, line 94: still got this caching of threads in here, needs removal as we definitely dont want to cache them :)

include/dpp/dispatcher.h
```diff
@@ -715,6 +715,33 @@ struct integration_delete_t : public event_dispatch_t {
 	integration deleted_integration;
 };
 
+/** */
+struct thread_create_t : public event_dispatch_t {
+	/**
+	 */
+	thread_create_t(class discord_client* client, const std::string& raw);
+	guild* creating_guild;
+	channel* created;
```
braindigitalis, line 724: should these be channel* and not just channel? as you're putting 't' directly in them in the events, which is a non-ptr type?

include/dpp/guild.h
```diff
@@ -242,6 +242,9 @@ class guild : public managed {
 	/** List of channels on this server */
 	std::vector<snowflake> channels;
 
+	/** List of threads on this server */
+	std::vector<snowflake> threads;
```
braindigitalis, line 246: good catch, we're going to need to keep these snowflake ids around

src/dpp/cluster.cpp
```diff
@@ -1219,6 +1219,97 @@ void cluster::current_user_leave_guild(snowflake guild_id, command_completion_ev
 	 });
 }
 
+void cluster::current_user_join_thread(snowflake thread_id, command_completion_event_t callback) {
+	this->post_rest("/api/v9/channels",thread_id, "/thread-members/@me", m_put, "", [callback](json &j, const http_request_completion_t& http {
+		if (callback) {
+			callback(confirmation_callback_t("confirmation", confirmation(), http));
```
braindigitalis, line 1225: do any of these return the thread info instead? or just 204 empty content? if they return any meainingful content we should find a way to pass that content to the user

braindigitalis (changes requested):
looking good so far! 💯

src/dpp/cluster.cpp
Axyte: Returns a 204 empty response on success.

braindigitalis (approved):

PR #52, opened 2021-08-02
Emoji shortcuts and important fix for segmentation fault!

Lohkdesgds:
I know it's not cool to have feature and a fix at the same time. The fix is important, but the "features" are just a shortcut I think are great for some applications.

The important one is the `src/dpp/cluster.cpp`


braindigitalis:
"cool" or not i'll still accept it and merge it, many thanks! :)

PR #55, opened 2021-08-13
sticker REST and events support

braindigitalis:
Better (read as: full) support for stickers via REST calls for their creation and editing, plus sticker change events.

### DIFF
--- a/include/dpp/message.h
+++ b/include/dpp/message.h
@@ -621,7 +621,12 @@ struct sticker {
 	 * for standard stickers, a comma-separated list of related expressions.
 	 */
 	std::string     tags;
-	/// Deprecated previously the sticker asset hash, now an empty string
+	/**
+	 * @brief Asset ID
+	 * @deprecated now an empty string but still sent by discord.
+	 * While discord still send this empty string value we will still have a field
+	 * here in the library.
+	 */
 	std::string     asset;
 	/** The type of sticker */
 	sticker_type	type;
--- a/include/dpp/message.h
+++ b/include/dpp/message.h
@@ -587,28 +587,127 @@ struct attachment {
 	~attachment() = default;
 };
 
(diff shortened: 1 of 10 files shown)


include/dpp/message.h
```diff
@@ -587,28 +587,127 @@ struct attachment {
...
 	/** The name of the sticker */
-	std::string name;
+	std::string	name;
+	/// description of the sticker (may be empty)
+	std::string     description;    
+	/** for guild stickers, the Discord name of a unicode emoji representing the sticker's expression.
+	 * for standard stickers, a comma-separated list of related expressions.
+	 */
+	std::string     tags;
+	/// Deprecated previously the sticker asset hash, now an empty string
```
webtax-gh, line 624: Do we need to support this if it’s always an empty string?

src/dpp/message.cpp
```diff
@@ -886,4 +882,105 @@ message& message::fill_from_json(json* d, cache_policy_t cp) {
...
+sticker::sticker() : id(0), pack_id(0), guild_id(0), type(st_standard), format_type(sf_png), available(true), sort_value(0) {
+}
+
+sticker& sticker::fill_from_json(nlohmann::json* j) {
+	this->id = SnowflakeNotNull(j, "id");
+	this->pack_id = SnowflakeNotNull(j, "pack_id");
+	this->name = StringNotNull(j, "name");
+	this->description = StringNotNull(j, "description");
+	this->tags = StringNotNull(j, "tags");
+	this->asset = StringNotNull(j, "asset");
```
webtax-gh, line 894: Same here

include/dpp/message.h
braindigitalis: if its in the spec on discord api i'd say so. we'll only need to add it later when or if they decide to put a value in it

PR #62, opened 2021-09-08
QOL Changes 

Axyte:


src/dpp/dispatcher.cpp
```diff
@@ -43,6 +43,13 @@ void interaction_create_t::reply(interaction_response_type t, const std::string
 	this->reply(t, dpp::message(this->command.channel_id, mt, mt_application_command));
 }
 
+void interaction_create_t::get_original_response(command_completion_event_t callback) const
+{
+	from->creator->post_rest(API_PATH "/webhooks", std::to_string(command.application_id), command.token + "/messages/@original", m_get, "", [callback](json &j, const http_request_completion_t& http) {
+		callback(confirmation_callback_t("message", message().fill_from_json(&j), http));
```
braindigitalis, line 49: should probably check callback is valid before calling it, in case someone passes in an uninitialised `std::function`

include/dpp/dispatcher.h
```diff
@@ -28,6 +28,11 @@
 
 namespace dpp {
 
+#define API_PATH	"/api/v" DISCORD_API_VERSION
+
+struct CoreExport confirmation_callback_t;
```
braindigitalis, line 33: CoreExport is not needed on a forward declaration

include/dpp/dispatcher.h
```diff
@@ -28,6 +28,11 @@
 
 namespace dpp {
 
+#define API_PATH	"/api/v" DISCORD_API_VERSION
```
braindigitalis, line 31: if we have moved this definition to here, do we then need to take it out of the cpp file it came from? (iirc cluster.cpp)

.github/workflows/ci.yml
```diff
@@ -89,3 +89,41 @@ jobs:
...
+
+      - name: Install Packages
+        run: sudo apt-get install -y zlib1g-dev:arm64 libssl-dev:arm64 g++-9-aarch64-linux-gnu gcc-9-aarch64-linux-gnu ninja-build libopus-dev:arm64 libsodium-dev:arm64 
+
+      - name: Prepare System for CMake
+        run: export PATH=$PATH:/usr/include:/usr/lib/aarch64-linux-gnu/:/usr/bin && sudo mv /usr/include/x86_64-linux-gnu/openssl/opensslconf.h /usr/include/openssl/ && sudo rm /usr/lib/x86_64-linux-gnu/libssl.so /usr/lib/x86_64-linux-gnu/libcrypto.so /usr/lib/x86_64-linux-gnu/libz.so /usr/lib/x86_64-linux-gnu/libcrypto.a /usr/lib/x86_64-linux-gnu/libssl.a && sudo mv /usr/lib/aarch64-linux-gnu/pkgconfig/libsodium.pc /usr/lib/pkgconfig/
+
+      - name: Generate CMakeFiles
+        run: mkdir build && cd build && cmake -DCMAKE_TOOLCHAIN_FILE=cmake/ARM64ToolChain.cmake -G Ninja ..
```
braindigitalis, line 116: I like how you've put this in a separate toolchain file that only gets included if you are building this deb file

include/dpp/dispatcher.h
Axyte: It's just a work around to avoid including the whole cluster.h file, when we only really need some basic stuff from it. But yeah, we can move that away from there.

src/dpp/dispatcher.cpp
Axyte: I didn't want branching there because it mostly it's a fixed function that is passed and would error out, saving us from branching, but I get the point you make. If you want, I can add it.

PR #68, opened 2021-09-10
Explicitly remove the const qualifier from string literal.

iSLC:
Mainly gets rid of the compiler ISO warning. The issue still remains but the intention is now obvious.


braindigitalis:
thanks!

PR #70, opened 2021-09-12
Webhook Adjustments

mitchej123:
I was having a hard time using a webhook to both post messages and use a user provided username.  These changes worked for me.

* Use the webhook token, if provided otherwise fall back to previous behavior (bot token).  According to the API documentation this should be a webhook token, not the bot token
image

* Add "username" to messages if there is an author, currently only used for execute_webhook
image


braindigitalis:
thanks, these functions should never use the bot token and it looks to be picking up this->token instead of a parameter called token.
I'll merge this in when I have access to a pc!

PR #103, opened 2021-10-01
Add support for role icons

Linker-123:
Adds `unicode_emoji` and `icon` properties to the Role class.

### DIFF
--- a/include/dpp/role.h
+++ b/include/dpp/role.h
@@ -98,6 +98,10 @@ public:
 	snowflake integration_id;
 	/** Bot id if any (e.g. role is a bot's role created when it was invited) */
 	snowflake bot_id;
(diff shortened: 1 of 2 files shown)


Axyte (changes requested):
The library has a `dpp::utility::iconhash` for icon hashes. It should be used, instead of `std::string`.

Linker-123:
> The library has a `dpp::utility::iconhash` for icon hashes. It should be used, instead of `std::string`.

I see, thanks. Should the `icon` and `unicode_emoji` properties be included in the build_json function?

Axyte:
Yes, since endpoints such as these accept them.
https://discord.com/developers/docs/resources/guild#create-guild-role

braindigitalis (approved):

PR #111, opened 2021-10-03
guild_get_members fixes, add guild_search_members

Axyte:


Axyte:
Should these objects be cached as well?

braindigitalis:
it would make sense to cache these but only when the cache policy is not `cp_none`.

Axyte:
Only user objects though.

include/dpp/cluster.h
```diff
@@ -1435,9 +1435,11 @@ class CoreExport cluster {
 	 * @brief Get all guild members
 	 *
 	 * @param guild_id Guild ID to get all members for
+	 * @param limit max number of members to return (1-1000)
+	 * @param after the highest user id in the previous page
 	 * @param callback Function to call when the API call completes.
 	 */
-	void guild_get_members(snowflake guild_id, command_completion_event_t callback);
+	void guild_get_members(snowflake guild_id, uint16_t limit, snowflake after, command_completion_event_t callback);
```
braindigitalis, line 1442: these should be optional yes? (so, after the callback, ugh)

Axyte: Follows the scheme of `messages_get` I think.

PR #125, opened 2021-10-13
created an embed example page

Commandserver:
## changes
added a page with an embed example (inspired by discord.js)
also added a screenshot to demonstrate the example

hopefully i made everything correctly

I named the page "Sending Embeds". If someone have a better name, feel free to change it


braindigitalis:
thanks, ive merged this and it should be live in the docs in a few minutes 👍

PR #127, opened 2021-10-14
fix guild_get_member bug

uwuAnimelover:
path should be 'members', not 'member'


braindigitalis:
looks good to me, i'll push this in last-minute into todays 9.0.8 release - thanks!

PR #153, opened 2021-11-16
200 stars when?

Axyte:


braindigitalis:
this is failing right now because `const` is missing off the end of the function definition where the body is defined.
Also consider making the declaration like:

`void reply(const message& msg, command_completion_event_t callback = {}) const;`

because without the `const`, you can't in-place instantiate a dpp::message and can only pass in a non-const variable.

Axyte:
> this is failing right now because `const` is missing off the end of the function definition where the body is defined.

Yup, mis-matched it.

> Also consider making the declaration like:
> 
> `void reply(const message& msg, command_completion_event_t callback = {}) const;`
> 
> because without the `const`, you can't in-place instantiate a dpp::message and can only pass in a non-const variable.

Regarding that, we can't take const references since we change member variables.

PR #156, opened 2021-11-20
adding attachment example

Commandserver:
2 program examples:
- uploading a local file
- download a file via URL and attach it


Commandserver:
i hope there are not too many typos

Commandserver:
the only thing that's missing is multi files

PR #166, opened 2021-12-06
updated docs: added truncated-message

Commandserver:


Commandserver:
you could double check if the length of those are right 💯

PR #167, opened 2021-12-07
Update INDEX.md

1olipop:
Changed "Github page." to "Github repository." as the second one makes more sense and also, Github pages already is a thing existing which doesn't share a meaning with a github repo.


braindigitalis:
thanks, are you on the Discord server so I can grant you your contributor role?

1olipop:
Yea, Knightcicle#2838

PR #169, opened 2021-12-11
added messages to slashcommands

Commandserver:
added messages as it was missing for the context menus
**(idk if it'll work )grafik**


braindigitalis:
idea is sound and build passed, thanks gonna merge it

PR #173, opened 2021-12-12
small fix for the code in the context menu example

Commandserver:
the code in the example was not displayed correctly


braindigitalis:
just a little trick you can do, if you install doxygen and graphviz, you can run doxygen from your D++ dir and it'll generate local docs in the docs dir. you can then check locally that a PR looks fine before submitting it

PR #174, opened 2021-12-13
fixed things with context menu ig

Commandserver:
discord conversation related to that https://discord.com/channels/825407338755653642/919319123572449280/919319934062985216

first i tried this and nothing happend:
```c++
bot.on_interaction_create([&](const dpp::interaction_create_t &event) {
        if (event.command.type == dpp::ctxm_message) {
            cout << "message event fired" << endl;
        }
    });
```
later i was realising that `event.command.type` can only compared with enum `interaction_type`.
okay so i came up with this:
```c++
bot.on_interaction_create([&](const dpp::interaction_create_t &event) {
        if (event.command.type == dpp::it_application_command) {
            dpp::command_interaction cmd_data = std::get<dpp::command_interaction>(event.command.data);
            if (cmd_data.type == dpp::ctxm_message) {
                if (cmd_data.name == "testaction") {
                    cout << "testaction context menu fired" << endl;
                }
            }
        }
    });
```
of course that was also **not working**.
i was wondering how to handle the message context menu?!
so i looked over the code and noticed that something was wrong. somehow a struct weren't equal to the discord docs (i think lol).

i've made some changes and now this works fine for me:
```c++
bot.on_interaction_create([&](const dpp::interaction_create_t &event) {
        if (event.command.type == dpp::it_application_command) {
            dpp::command_interaction cmd_data = std::get<dpp::command_interaction>(event.command.data);
            cout << "Interaction type: " << cmd_data.type << endl;
            if (cmd_data.type == dpp::ctxm_message) {
                cout << "message id: " << cmd_data.target_id << endl;
                if (cmd_data.name == "test") {
                    cout << "message context menu fired" << endl;
                }
            }
        }
    });
```

more at the review :)


braindigitalis:
there are no guarantees in dpp that the structure of the lib will equal the structure of the raw API. we make changes and optimisations to make for more efficient code, we also assume that users don't know the Discord API itself or need to reference it, as many do not and only know DPP. 

Regardless this could do with some helper functions to make it easier. 

anything regarding slash commands goes into the command_interaction, as many other things are received via interaction_create that are not slash commands and are separated into different structures.

other libraries mimicked the discord api closer and this landed them in hot water if they were not dynamically typed languages as the objects don't directly map for inheritence, just stuffing of extra fields like you would in JavaScript.

I'll give some thought as to how the use of context menu can be made better but remember for some silly reason a context menu is a slash command in the API. perhaps this needs it's own event like we have for buttons etc? food for thought...

braindigitalis:
merged with thanks! does this count as a breaking change?
I still think we need some helper functions for making this even easier.

Commandserver:
> I'll give some thought as to how the use of context menu can be made better but remember for some silly reason a context menu is a slash command in the API. perhaps this needs it's own event like we have for buttons etc? food for thought...

yeah. Having it's on event for the context menus would be the best. but i think then it would also be important to have something like `event.command.data.message` which simply contains the message or user whom you clicked on instead of having `target_id` and `resolved` and let the user loop trough the resolved messages and find the one with target_id etc. u know what i mean?

Commandserver:
> merged with thanks! does this count as a breaking change?
i think so. but anyway i would remake the context menu and make an own event. this would be a breaking change ig...

PR #179, opened 2021-12-16
Improved attachment example

Commandserver:
- bugfix: absolute paths didn't work in the first code block. Issue with `relative_path()`
- added an example how to upload an image and use it in an embed


braindigitalis:
does this need special build instructions?
`std::filesystem` is silly and requires a separate lib to be linked at compile time on g++ for some ass backwards reason

Commandserver:
...changed

PR #192, opened 2021-12-22
Update presence.h activity types to match whats on the discord docs.

DragonFire101:
updated presences activity types ~~not including custom since bots cant access the emoji field to use it correctly.~~

https://discord.com/developers/docs/topics/gateway#activity-object-activity-types


DragonFire101:
I left out at_custom due to bots not able to use it correctly because they cant access the emoji field in the activity struct.
I can add it in if required.

PR #206, opened 2022-01-06
Interstellar build system

SirObby:
Added support for [Interstellar](https://github.com/SirObby/interstellar) although it makes no sense to support it.

Added files:
`build.inter`
Modified files:
`docpages/02_build.md`

### DIFF
--- a/build.inter
+++ b/build.inter
@@ -1,0 +1,14 @@
+VARIABLE, VALUE;
+$_CC, /usr/bin/g++;
+add_source_folder, ./src;
+add_source_folder, ./src/dpp;
+add_source_folder, ./src/dpp/cluster;
+add_source_folder, ./src/dpp/events;
+add_include, ./include;
+link_libraries, -lm;
+add_flags, -pthread;
+add_flags, -lzlib;
+add_flags, -lopenssl;
+add_flags, -fPIC;
+add_library, dpp;
+add_shared_library, libdpp.so;
--- a/docpages/02_build.md
+++ b/docpages/02_build.md
@@ -6,6 +6,7 @@ The way you build D++ varies from system to system. Please follow the guide belo
 * \subpage buildwindows "Building on Windows"
 * \subpage buildosx "Building on OSX"
 * \subpage buildfreebsd "Building on FreeBSD"
+* \subpage buildlinux "Building using Interstallar (on Linux)"
 
 \page buildlinux Building on Linux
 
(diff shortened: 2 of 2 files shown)


braindigitalis:
there's something missing, what about opus and sodium and detection of the libs that are used? it isn't safe just to assume we can do -l<lihname> is it?

SirObby:
Interstellar is a very new build system, as such it hasn't matured like CMake.
I think until I implement if statements or something to detect opus and sodium, Interstellar support could be voice-less.

SirObby:
I can add a disclaimer in the docs saying voice isn't supported when using interstellar.

SirObby:
> why confuse new users with this unknown build system, isn't this lib intended to be easy for new users and users coming from other languages?

Compiling using interstellar has one command, using cmake and make has 4.

SirObby:
Also interstellar is probably among the simplest build system right now.

Linker-123:
You might just drop the project in a few weeks, don't see what's the point of having it there.

SirObby:
> You might just drop the project in a few weeks, don't see what's the point of having it there.

I'm not going to drop the project in a few weeks.

SirObby:
@Linker-123 Interstellar is like among the simplest and smallest projects I have, not going to drop it just, because you said so.

NexAdn:
> Compiling using interstellar has one command, using cmake and make has 4.

So what? If you take that argument, then just use plain make. There is a reason more mature build systems are in use today.

Also, I am against the idea of supporting two build systems for now real reason. I think this project should stick with only a single build system (be it CMake or something else, as long as it has the required features and at least a bit of maturity). Supporting a second build system just for the sake of supporting it, is not a good argument. Build systems are responsible for producing builds reliably and if we have two build systems, we might end up with two systems producing different results on the same codebase, if there is just a slightly different configuration between those.

SirObby:
They both use G++ or CMake, there aint gon be difference between builds using CMake or Interstellar.

NexAdn:
Actually there can. Consider that all source files that need compiling have to be stated in CMake, whereas in Interstellar they don't have (from what I deduced from your commits). Now, if you have some uncommited sources lying around in your working directory, you _will_ get a different result than someone using the unmodified master. In CMake, one would notice this as CMake either doesn't include the files in the build process (the good outcome) or complains about the files missing (the bad outcome). In either way, you'll immediately see that there is something wrong with the files.

Also, consider the automatic decision in CMake, whether or not to enable voice support (I don't like that CMake decides that automatically, btw, but that's another topic and not relevant to this discussion). From what I've read in this thread, Interstellar doesn't have such checks and thus reliably produces a different build result in the non-default case.

Next, Interstellar doesn't seem to care about detection how the libraries are named and where the include paths are located. I already had to work with shitty software that required vendored forks of other libraries and thus broke a system-wide install (and in the process maybe even other software on my system) completely, if I didn't change the prefix to some non-standard directory (in this case, I changed it to a subdirectory in `/opt`). CMake can deal with this kind of situation, since one can provide a custom install prefix and CMake stores the install prefix in its module config file (which can be relocated, too, btw), while Interstellar can not. So, as soon as I am in the scenario that I want to use a custom version of some dependency which is not located in standard directories (consider some newer version which isn't available in your distribution's repositories, which you had to compile yourself), Interstellar produces a different build result or (in the best case) fails if there isn't a second (older) version of the depenency installed in the system directories.

So, apart from the question, why one would support two build systems: Why would one event _want_ to use Interstellar as build system at all? It just lacks features that _are_ needed in today's world of shitty vendored dependencies and people thinking their software is so special that it mustn't be installed in standard system directories. And I haven't even talked about the problems of integrating something like cmake module files so the library can be used as a dependency in other projects.

Don't get me wrong, though. I haven't had a deep look into Interstellar (although I had a quick glance at the codebase while writing this comment). Interstellar can become a good build system and even suitable for projects, but in its current stage I don't think it is suited for any larger project or any library which shall be reused by other software. Currently, it is probably just `make` with extra steps and less features.

SirObby:
I agree, Interstellar shouldn't yet be used in prod (I should've marked this as a draft) it's still very new and missing some features that are required for a good build system, if you'd like to you can PR to it. There are some bad code and stuff in the code which hopefully I'll get fixed soon.

SirObby:
I'm closing this PR due to the planned changes for the Interstellar build system.
I will make a new PR once the Interstellar build system has been redone.

Issue #212, opened 2022-01-08
guild_command_create does not edit the slashcommand permissions


Commandserver:
**Describe the bug**
the permissions of slash commands aren't working when using `guild_command_create`.

i found out that `guild_command_edit_permissions` inside the `cluster::guild_command_create` method always throw and error:

```
{"message": "Unknown application command", "code": 10063}
```
I found out that this happens because the given slash command needs an `id` but doesn't have one.

So #209 isn't fully fixed.

**To Reproduce**

add a lambda in `guild_command_create` like so:

```c++
void cluster::guild_command_create(const slashcommand &s, snowflake guild_id, command_completion_event_t callback) {
	this->post_rest(API_PATH "/applications", std::to_string(s.application_id ? s.application_id : me.id), "guilds/" + std::to_string(guild_id) + "/commands", m_post, s.build_json(false), [s, this, guild_id, callback] (json &j, const http_request_completion_t& http) mutable {
		if (callback) {
			callback(confirmation_callback_t("slashcommand", slashcommand().fill_from_json(&j), http));
		}

		if (http.status < 300 && s.permissions.size()) {
			guild_command_edit_permissions(s, guild_id, [](const confirmation_callback_t & e) {
				if (e.is_error()) {
					std::cout << e.http_info.body << std::endl;
				} else {
					std::cout << "success" << std::endl;
				}
			});
		}
	});
}
```

**Expected behavior**
the `guild_command_edit_permissions` method works

braindigitalis:
this won't ever work because "s" doesn't contain the application id. you need to std::get the slashcommand returned inside the callback not reference the one you originally passed in. 

slashcommand passed in should not be a non const ref. is this up to date with master or dev branch?

Commandserver:
> slashcommand passed in should not be a non const ref. is this up to date with master or dev branch?

ups, yeah it should be a const

Commandserver:
> this won't ever work because "s" doesn't contain the application id. you need to std::get the slashcommand returned inside the callback not reference the one you originally passed in.

what do you mean? geting the slashcommand from where?

braindigitalis:
see #215 to see what I mean

Commandserver:
yeah, i got it :)

Commandserver:
okay

PR #213, opened 2022-01-08
add fix to guild_command_create to solve the permissions edit issue

Commandserver:
seems like these lines were there before https://github.com/brainboxdotcc/DPP/blob/f8e48910d0586b028a9495bd28078b1fa911c577/src/dpp/cluster.cpp#L385-L387

The parameter `s` will change to a non-const parameter this could be a breaking change?!

Shouldn't we change that also in `global_command_create` and other methods @braindigitalis ?

#212

### DIFF
--- a/include/dpp/cluster.h
+++ b/include/dpp/cluster.h
@@ -1461,7 +1461,7 @@ public:
 	 * @param callback Function to call when the API call completes.
 	 * On success the callback will contain a dpp::slashcommmand object in confirmation_callback_t::value. On failure, the value is undefined and confirmation_callback_t::is_error() method will return true. You can obtain full error details with confirmation_callback_t::get_error().
 	 */
-	void guild_command_create(const slashcommand &s, snowflake guild_id, command_completion_event_t callback = {});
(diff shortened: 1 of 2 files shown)


Commandserver:
why does it fail? could you rerun the checks @braindigitalis ?

braindigitalis:
> seems like these lines were there before
> 
> https://github.com/brainboxdotcc/DPP/blob/f8e48910d0586b028a9495bd28078b1fa911c577/src/dpp/cluster.cpp#L385-L387
> 
> The parameter `s` will change to a non-const parameter this could be a breaking change?!
> 
> Shouldn't we change that also in `global_command_create` and other methods @braindigitalis ?
> 
> #212

this is massively out of date... we have split up cluster.cpp even since this was in here, it's no longer a reference, you're working from an ancient dpp?

braindigitalis:
this is the more recent code, using a constant, this is why it breaks, we need to take the copy of the created one inside the lambda using std::get

https://github.com/brainboxdotcc/DPP/blob/8c2cb99a4b7b5c022e6603a66a29a2672cdf78d9/src/dpp/cluster/appcommand.cpp#L111

PR #214, opened 2022-01-08
fixed timestamp parsing method for communication_disabled_until

### LINKED ISSUE #203: is_communication_disabled method sometimes not working
**Describe the bug**
`is_communication_disabled` in `dpp::member` returns false when the member is in timeout!

**To Reproduce**

It happens when i set `communication_disabled_until` to +1 hour.
But only with short timeouts. For example I've tried it with a user that i gave a 600 days timeout and this worked fine.

**Expected behavior**
returns true when the user really is in timeout.

**System Details:**
my timezone is `UTC +1`. maybe this is usefull?!

**Additional context**
I've added some debugging to the current code:
```c++
bool guild_member::is_communication_disabled() const {
    std::cout << "internal communication_disabled_until: " << std::to_string(communication_disabled_until) << " current timestamp: " << time(nullptr) << std::endl;
    return communication_disabled_until > time(nullptr);
}
```
I called the method on a user that have a one hour timeout, and it returns false.
the code above printed 
> internal communication_disabled_until: 1641250800 current timestamp: 1641335917

I think its because discord uses ISO8601 timestamp and not the normal unix timestamp.
https://discord.com/developers/docs/resources/guild#guild-member-object-guild-member-structure

Commandserver:
changed the first if condition so that timestamps without dots will be parsed right, like this one: `2022-01-12T13:56:05+00:00`.
this could be a thing to be added to unit tests.

hopefully it doesn't break something :)

fixes #203

### DIFF
--- a/include/dpp/cluster.h
+++ b/include/dpp/cluster.h
@@ -1461,7 +1461,7 @@ public:
 	 * @param callback Function to call when the API call completes.
 	 * On success the callback will contain a dpp::slashcommmand object in confirmation_callback_t::value. On failure, the value is undefined and confirmation_callback_t::is_error() method will return true. You can obtain full error details with confirmation_callback_t::get_error().
 	 */
-	void guild_command_create(slashcommand &s, snowflake guild_id, command_completion_event_t callback = {});
+	void guild_command_create(const slashcommand &s, snowflake guild_id, command_completion_event_t callback = {});
 
 
 	/**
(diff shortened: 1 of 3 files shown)


braindigitalis:
has this been tested and does it break other timestamp formats with the . part?

Commandserver:
i've tested it with `member.joined_at` and this has worked

Commandserver:
i've now tested it with:
```c++
json j;
	j["test1"] = "2022-01-12T13:56:05+00:00";
	j["test2"] = "2022-01-12T13:56:05.32409+00:00";
	time_t t;
	dpp::set_ts_not_null(&j, "test1", t);
	cout << "Parsed time 1: " << t << endl;
	dpp::set_ts_not_null(&j, "test2", t);
	cout << "Parsed time 2: " << t << endl;
	```
and they both print
```
Parsed time 1: 1641992165
Parsed time 2: 1641992165
```

braindigitalis:
looks good to me, thanks so much for the effort spent in identifying and fixing this!

PR #217, opened 2022-01-13
improvements in invite class & docs enhancements

Commandserver:
- fixed a bug where `invite.guild_id` is not set, because the json was wrong parsed
- added some attributes to `invite`
- docs

### DIFF
--- a/include/dpp/invite.h
+++ b/include/dpp/invite.h
@@ -69,7 +69,7 @@ public:
 	/** True if this invite should not replace or "attach to" similar invites
 	 */
 	bool unique;
-	/** Readonly amount of times this invite has been used
+	/** How many times this invite has been used
 	 *
 	 * @note Only set when using cluster::channel_invites_get
 	 */
--- a/include/dpp/invite.h
+++ b/include/dpp/invite.h
@@ -33,6 +33,9 @@ public:
 	/** Invite code
 	 */
 	std::string code;
+	/** Readonly expiration timestamp of this invite or 0 if the invite doesn't expire
+	 */
+	time_t expires_at;
 	/** Guild for the invite
 	 */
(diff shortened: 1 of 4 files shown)


Commandserver:
I've tested it both btw, `dpp::invite::uses` and `dpp::invite::expires_at`

src/dpp/cluster/invite.cpp
```diff
@@ -50,7 +50,7 @@ void cluster::invite_delete(const std::string &invitecode, command_completion_ev
 
 
 void cluster::invite_get(const std::string &invitecode, command_completion_event_t callback) {
-	this->post_rest(API_PATH "/invites", dpp::url_encode(invitecode) + "?with_counts=true", "", m_get, "", [callback](json &j, const http_request_completion_t& http) {
+	this->post_rest(API_PATH "/invites", dpp::url_encode(invitecode) + "?with_counts=true&with_expiration=true", "", m_get, "", [callback](json &j, const http_request_completion_t& http) {
```
Axyte, line 53: This should be optional if possible.

include/dpp/invite.h
```diff
@@ -66,6 +69,11 @@ class DPP_EXPORT invite {
 	/** True if this invite should not replace or "attach to" similar invites
 	 */
 	bool unique;
+	/** How many times this invite has been used
+	 *
+	 * @note Only set when using cluster::channel_invites_get
+	 */
+	uint32_t uses;
```
Axyte, line 76: Set to zero in constructor alongside other added fields in the PR?

src/dpp/cluster/invite.cpp
Axyte: To provide choice to the user, of course.

PR #226, opened 2022-01-18
added webhook example

Commandserver:


Commandserver:
@braindigitalis when i added the callback to `execute_webhook` id didnt get executed because the program exits

PR #228, opened 2022-01-19
added timestamp parser to unit tests

Commandserver:


Commandserver:
could it be a problem when `2022-01-19T17:18:14Z` isn't right parsed to a timestamp?

PR #233, opened 2022-01-23
Export include directory to external CMake targets

atomheartother:
This PR exports the `include/` directory to any other CMake project who may want to include `dpp` as a source-dependency instead of building it system-wide. This makes the project more compatible with modern CMake practices and also makes it possible to install out-of-the-box using package managers like [CPM](https://github.com/cpm-cmake/CPM.cmake).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -138,8 +138,6 @@ endif()
 
 include_directories(${OPENSSL_INCLUDE_DIR} ${ZLIB_INCLUDE_DIRS})
 
-include_directories("include")
-
 find_package(Git QUIET)
 
 if(NOT GIT_FOUND AND EXISTS "${PROJECT_SOURCE_DIR}/.git")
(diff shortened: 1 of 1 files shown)


atomheartother:
For the record and for any future folk interested (or if you all would like to add CPM to the supported ways to install `dpp`), here's a basic CMake which builds dpp using CPM.cmake on Linux. Note that there are 0 guarantees that this will work on windows as windows is its own can of worms.

```cmake
cmake_minimum_required(VERSION 3.0)
project(cppbot VERSION "0.0.1")

include(cmake/CPM.cmake)
set(CPM_SOURCE_CACHE ${CMAKE_CURRENT_SOURCE_DIR}/deps/)

# Currently adds my branch on my repo, you'd just need to change the repo and tag to match whatever you want, preferably a specific version like v9.0.18
CPMAddPackage(
  NAME dpp
  GITHUB_REPOSITORY atomheartother/DPP
  GIT_TAG feat/cmake-export-include
)

add_executable(cppbot)

target_sources(cppbot PRIVATE
  src/main.cpp
)

target_link_libraries(cppbot PRIVATE
  dpp
)
```

PR #242, opened 2022-01-27
fix: role_edit_position and emoji::format didn't return animated emojis

Commandserver:
role position endpoint: https://discord.com/developers/docs/resources/guild#modify-guild-role-positions

### DIFF
--- a/include/dpp/cluster.h
+++ b/include/dpp/cluster.h
@@ -2675,12 +2675,12 @@ public:
 	 * Fires multiple `Guild Role Update` Gateway events.
 	 *
 	 * @note This method supports audit log reasons set by the cluster::set_audit_reason() method.
-	 * @param guild_id Guild ID to change the role positions on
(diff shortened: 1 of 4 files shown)


Commandserver:
I've tested the added `role_edit_positions` method for functionality

include/dpp/cluster.h
```diff
@@ -2671,7 +2685,7 @@ class DPP_EXPORT cluster {
 	/**
 	 * @brief Edit a role's position in a guild
 	 * 
-	 * Modify the positions of a set of role objects for the guild. Requires the `MANAGE_ROLES` permission.
+	 * Modify the position of a role. Requires the `MANAGE_ROLES` permission.
```
braindigitalis, line 2688: instead of doing this the function should be fixed to accept multiple roles like the API call it connects to

braindigitalis (changes requested):

braindigitalis (approved):

PR #243, opened 2022-01-27
rain rain go away

Axyte:


src/dpp/channel.cpp
```diff
@@ -78,6 +78,61 @@ std::string channel::get_mention() const {
...
+	return *this;
+}
+
+channel& channel::remove_flag(const channel_flags flag) {
+	this->flags &= ~flag;
+	return *this;
+}
+
+channel& channel::set_nsfw(const bool is_nsfw) {
+	this->flags |= (is_nsfw) ? dpp::c_nsfw : 0;
```
braindigitalis, line 127: this won't work. you need to use ~ and ^ to invert the bit

src/dpp/channel.cpp
```diff
@@ -78,6 +78,61 @@ std::string channel::get_mention() const {
...
+	this->name = utility::validate(name, 1, 100, "name must be at least 1 character");
+	return *this;
+}
+
+channel& channel::set_topic(const std::string& topic) {
+	this->topic = utility::utf8substr(topic, 0, 1024);
+	return *this;
+}
+
+channel& channel::set_parent_id(const snowflake parent_id) {
```
braindigitalis, line 91: do we need methods for these,.when we can do c->parent_id = x ?

src/dpp/channel.cpp
Axyte: Oh thanks, missed that. And did you mean `&` instead of `^`?

src/dpp/channel.cpp
Axyte: I mean that logic can be applied to every other class but we still have setters, don't we?

PR #253, opened 2022-02-03
fixed cluster::guild_get_members and cluster::guild_search_members

### LINKED ISSUE #250: guild_get_members
**Describe the bug**
guild_get_members is not working properly.
when casting to guild_member_map size is always 1 and contains incorrect info.

Once called 
**To Reproduce**
```
bot->on_ready([&](const dpp::ready_t& event) {
	p.guild_get_members(GUILD_ID, 50, 0, [](const dpp::confirmation_callback_t& event) {
		printf("Get member = %s\n", event.http_info.body.c_str()); // print correctly every user
		auto members = (dpp::guild_member_map&)(event.value); // work fine
		printf("Members size = %i\n", members.size()); // return 1
	});
})
```
**Expected behavior**
to the vector to contains every user? 

**System Details:**
 - Latest version (v9.0.18)

**Additional context**

Temporary fix ? 

https://github.com/brainboxdotcc/DPP/blob/master/src/dpp/cluster/guild_member.cpp#L74 

from `guild_members[snowflake_not_null(&curr_member, "id")] = guild_member().fill_from_json(&curr_member, guild_id, user_id);
` 
to -> `guild_members[guild_members.size() + 1] = guild_member().fill_from_json(&curr_member, guild_id, user_id);`

after checking json, ["id"] only exist in ["user"]

Commandserver:
fix #250


Jerem584:
thanks !

Commandserver:
> thanks !

Thanks for reporting :D

PR #265, opened 2022-02-09
fixed role_edit

### LINKED ISSUE #264: role_edit() fails with "Invalid Form Body"
**Describe the bug**
When attempting to use cluster::role_edit() to change a role we throw an error in the callback.

`{"code": 50035, "errors": {"_errors": [{"code": "DICT_TYPE_CONVERT", "message": "Only dictionaries may be used in a DictType"}]}, "message": "Invalid Form Body"}`

**To Reproduce**
1. Call role_edit() while passing in a modified or unmodified role.
2. Check the error results.

Here is code that I am using to reproduce the bug:

```
bot.role_edit(role_to_edit.set_name("test"), [](const dpp::confirmation_callback_t& role) {
    if (role.is_error()) {
        std::cout << "error: " << role.get_error().message << std::endl;
        std::cout << "error: " << role.http_info.body << std::endl;
    }
    auto t = get<dpp::role>(role.value);
    std::cout << t.name << std::endl;
});
```

**Expected behavior**
I would expect the role name to be updated however instead we just simply error out.

**System Details:**
 - OS: macOS 12.2 and RaspberryPI latest 64bit
 - Discord Desktop PTB macOS

Commandserver:
fix #264

### DIFF
--- a/src/dpp/cluster/role.cpp
+++ b/src/dpp/cluster/role.cpp
@@ -49,7 +49,7 @@ void cluster::role_edit(const class role &r, command_completion_event_t callback
 	if (p != j.end()) {
 		j.erase(p);
 	}
-	this->post_rest(API_PATH "/guilds", std::to_string(r.guild_id), "roles/" + std::to_string(r.id) , m_patch, j.dump(), [r, callback](json &j, const http_request_completion_t& http) {
+	this->post_rest(API_PATH "/guilds", std::to_string(r.guild_id), "roles/" + std::to_string(r.id) , m_patch, j, [r, callback](json &j, const http_request_completion_t& http) {
 		if (callback) {
 			callback(confirmation_callback_t("role", role().fill_from_json(r.guild_id, &j), http));
 		}


Axyte:
Tested?

Commandserver:
yes

src/dpp/cluster/role.cpp
```diff
@@ -49,7 +49,7 @@ void cluster::role_edit(const class role &r, command_completion_event_t callback
 	if (p != j.end()) {
 		j.erase(p);
 	}
-	this->post_rest(API_PATH "/guilds", std::to_string(r.guild_id), "roles/" + std::to_string(r.id) , m_patch, j.dump(), [r, callback](json &j, const http_request_completion_t& http) {
+	this->post_rest(API_PATH "/guilds", std::to_string(r.guild_id), "roles/" + std::to_string(r.id) , m_patch, j, [r, callback](json &j, const http_request_completion_t& http) {
```
braindigitalis, line 52: this must be j.dump(), you cant implicitly convert json to string like this reliably

braindigitalis (changes requested):

Issue #218, opened 2022-01-13
scheduled events in invites


Commandserver:
**Describe the solution you'd like**
adding scheduled events support in `dpp::invite`.
like documented here https://discord.com/developers/docs/resources/invite#invite-object-invite-structure.
maybe also the `target_application` field of the invite.

PR #273, opened 2022-02-13
added stage attribute and fixed stage_instance_get not returning stage instances

### LINKED ISSUE #272: stage_instance_get does not return a stage instance
**Describe the bug**
when calling `dpp::cluster::stage_instance_get`, it throws
```
terminate called after throwing an instance of 'std::bad_variant_access'
  what():  std::get: wrong index for variant
```
when i get the stage instance in the callback.
Same with `stage_instance_create` and `stage_instance_edit`.

**Expected behavior**
having the stage instance in it.

**System Details:**
- DPP 9.0.18

Commandserver:
fix #272 

couldn't tested the new stage attribute but looks okay to me

invite object: https://discord.com/developers/docs/resources/invite#invite-object-invite-structure


braindigitalis:
also fixes #218

PR #278, opened 2022-02-15
update build docs

es183923:
(previous PR was to the wrong branch)
This replaces using:

```
mkdir build
cd build
cmake ..
make -j8
```

with just:

```
cmake -B ./build
cmake --build ./build -j8
```

It does the same thing, but uses `cmake` instead, has less steps, and shows color.

It also specifies that Visual Studio 2022 is allowed, not just Visual Studio 2019.

### DIFF
--- a/docpages/02_build.md
+++ b/docpages/02_build.md
@@ -106,24 +106,11 @@ For voice support, additional dependencies are required:
     
 Replace the number after -j with a number suitable for your setup, usually the same as the number of cores on your machine. `cmake` will fetch any dependencies that are required for you and ensure they are compiled alongside the library.
 
-## 4. Optional: Run test cases
-
-Run `./test` for unit test cases. 
-
-Set the following vars to values that suit your system. Create a test server for your test bot, where the bot has administrator. There really should be nobody but you and your test bot on this server, you'll have to set the server as a community server and make an event, at least one voice channel, one text channel. The user id can be the bot's user id.
-
-    export DPP_UNIT_TEST_TOKEN="ODI2ZSQ4CFYyMzgxUzkzzACy.HPL5PA.9qKR4uh8po63-pjYVrPAvQQO4ln"
-    export TEST_GUILD_ID="907951970017480704"
-    export TEST_TEXT_CHANNEL_ID="907951970017480707"
-    export TEST_VC_ID="907951970017480708"
-    export TEST_USER_ID="826535422381391913"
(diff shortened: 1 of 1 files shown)


docpages/02_build.md
```diff
@@ -103,16 +101,23 @@ For voice support, additional dependencies are required:
...
-    make -j8
+    cmake -B ./build
+    cmake --build ./build -j8
     
 Replace the number after -j with a number suitable for your setup, usually the same as the number of cores on your machine. `cmake` will fetch any dependencies that are required for you and ensure they are compiled alongside the library.
 
 ## 4. Optional: Run test cases
 
-Run `./test` for unit test cases. You will need to create a `config.json` file in the directory above the executable file with a valid bot token in it. See the example file `config.example.json` for an example of the correct format.
+Run `./test` for unit test cases.
```
braindigitalis, line 111: running unit tests is not really encouraged, this should probably be its own page under the advanced section, as only developers/contributors to the library need to do this step to verify their PR hasnt broken anything. The new page should probably also detail how to add new unit tests to test.cpp.

docpages/02_build.md
```diff
@@ -103,16 +101,23 @@ For voice support, additional dependencies are required:
...
 Replace the number after -j with a number suitable for your setup, usually the same as the number of cores on your machine. `cmake` will fetch any dependencies that are required for you and ensure they are compiled alongside the library.
 
 ## 4. Optional: Run test cases
 
-Run `./test` for unit test cases. You will need to create a `config.json` file in the directory above the executable file with a valid bot token in it. See the example file `config.example.json` for an example of the correct format.
+Run `./test` for unit test cases. 
+
+Set the following vars to values that suit your system. Create a test server for your test bot, where the bot has administrator. There really should be nobody but you and your test bot on this server, you'll have to set the server as a community server and make an event, at least one voice channel, one text channel. The user id can be the bot's user id.
+
+    export DPP_UNIT_TEST_TOKEN="ODI2ZSQ4CFYyMzgxUzkzzACy.HPL5PA.9qKR4uh8po63-pjYVrPAvQQO4ln"
```
braindigitalis, line 115: this is an expired or fake token, yes (like the one pinned in #library-development)?

braindigitalis (changes requested):

docpages/02_build.md
es183923: I'll remove the testing part from here, and I'll create a separate PR for testing.

PR #279, opened 2022-02-16
fix warning by using fallthrough annotation

es183923:
This PR prevents clang from warning while compiling by adding the `[[fallthrough]]` annotation.


src/dpp/slashcommand.cpp
```diff
@@ -376,9 +376,9 @@ void from_json(const nlohmann::json& j, command_data_option& cdo) {
 				break;
 			case co_attachment:
 				cdo.value = snowflake_not_null(&j, "value");
+				[[fallthrough]];
```
braindigitalis, line 379: this works but technically line 379 should be a `break;`, not fall-through, this is a harmless bug :)
If you can please change it to break, i'll accept it...
Thanks for the PR!

braindigitalis (changes requested):

PR #289, opened 2022-02-18
added guild_current_member_edit

Commandserver:
added `guild_current_member_edit` that will replace the old `guild_set_nickname`.

Discord just changes the route of it.

https://discord.com/developers/docs/resources/guild#modify-current-member


src/dpp/cluster/guild.cpp
```diff
@@ -23,6 +23,20 @@
 
 namespace dpp {
 
+void cluster::guild_current_member_edit(snowflake guild_id, const std::string &nickname, command_completion_event_t callback) {
+	std::string o;
+	if (nickname.empty()) {
+		o = "{\"nick\": null}";
```
braindigitalis, line 29: you can't use null here.
you must use the nlohmann json null type instead, this will crash.

PR #293, opened 2022-02-20
added new webhook constructors

DJOetzi:
one adds the ability to directly use the webhook URL provided by discord, and the other one takes the webhook ID and webhook URL


include/dpp/webhook.h
```diff
@@ -56,6 +56,24 @@ class DPP_EXPORT webhook : public managed {
 	 */
 	webhook();
 
+	/**
+     * @brief Construct a new webhook object using the Webhook URL provided by Discord
+    */
```
braindigitalis, line 61: need to document the url parameter:
`@param webhook_url a fully qualified web address of an existing webhook`

include/dpp/webhook.h
```diff
@@ -56,6 +56,24 @@ class DPP_EXPORT webhook : public managed {
...
+	webhook(std::string webhook_url)
+	{
+		token = webhook_url.substr(webhook_url.find_last_of("/") + 1); //find token
+		id = std::stoull(webhook_url.substr(sizeof("https://discord.com/api/webhooks/") - 1, 18)); //18 is the amount of characters a snowflake possesses
+	}
+
+	/**
+     * @brief Construct a new webhook object using the webhook ID and the webhook token
```
braindigitalis, line 69: same here, must document the parameters

braindigitalis (changes requested):

include/dpp/webhook.h
```diff
@@ -56,6 +56,24 @@ class DPP_EXPORT webhook : public managed {
 	 */
 	webhook();
 
+	/**
+     * @brief Construct a new webhook object using the Webhook URL provided by Discord
+    */
+	webhook(std::string webhook_url)
+	{
+		token = webhook_url.substr(webhook_url.find_last_of("/") + 1); //find token
+		id = std::stoull(webhook_url.substr(sizeof("https://discord.com/api/webhooks/") - 1, 18)); //18 is the amount of characters a snowflake possesses
```
es183923, line 65: A snowflake doesn't necessarily be 18 characters. There are 17 digit ids, and soon there will be 19 digit ids.

DJOetzi: Noted that, thanks for pointing it out!

braindigitalis: also bear in mind std::stoull can throw... if you don't want to bubble that exception to the user you should catch it, and probably re throw it as `dpp::logic_exception`

include/dpp/webhook.h
```diff
@@ -56,6 +56,24 @@ class DPP_EXPORT webhook : public managed {
...
+	}
+
+	/**
+     * @brief Construct a new webhook object using the webhook ID and the webhook token
+    */
+	webhook(snowflake webhook_id, std::string webhook_token)
```
braindigitalis, line 71: constructor body should be in webhook.cpp not .h

braindigitalis (changes requested):

include/dpp/webhook.h
braindigitalis: also ctor body belongs in the cpp or you can end up with it being defined more than once causing a compile error

src/dpp/webhook.cpp
```diff
@@ -33,6 +33,25 @@ webhook::webhook() : managed(), type(w_incoming), guild_id(0), channel_id(0), us
 {
 }
 
+webhook::webhook(std::string webhook_url)
+{
+	token = webhook_url.substr(webhook_url.find_last_of("/") + 1);
+	try
+	{
+		id = std::stoull(webhook_url.substr(sizeof("https://discord.com/api/webhooks/") - 1, webhook_url.find_last_of("/")));
```
braindigitalis, line 41: this isn't doing what you think it is, sizeof() on a std::string will return the size of the object not the string length.
use webhook_url.length(), or for a const char*, strlen.

src/dpp/webhook.cpp
```diff
@@ -33,6 +33,25 @@ webhook::webhook() : managed(), type(w_incoming), guild_id(0), channel_id(0), us
...
+	}
+	catch (std::exception& e)
+	{
+		throw dpp::logic_exception(std::string("Your webhook's URL failed to parse! ID invalid.\n") + e.what());
+	}
+}
+
+webhook::webhook(snowflake webhook_id, std::string webhook_token)
```
braindigitalis, line 49: string parameters should be `const std::string&`

DJOetzi: the ID can be made a constant too, right?

Axyte: Yes.

src/dpp/webhook.cpp
```diff
@@ -33,6 +33,25 @@ webhook::webhook() : managed(), type(w_incoming), guild_id(0), channel_id(0), us
...
+{
+	token = webhook_token;
+	id = webhook_id;
```
Axyte, line 52: Consider using initialiser lists, if assigning is all you do in a constructor.

include/dpp/webhook.h
```diff
@@ -56,6 +56,21 @@ class DPP_EXPORT webhook : public managed {
 	 */
 	webhook();
```
Axyte, line 57: Maybe mark it as default?

src/dpp/webhook.cpp
Axyte: I think you also need to initialise the other data members, as in the original constructor?

PR #299, opened 2022-02-23
fix interaction_modal_response

ErenOzt:
Added DPP_EXPORT to interaction_modal_response


braindigitalis:
thanks for the PR, i fixed this based on what you reported, as you submitted this PR i'll count you as the fixer 😄

PR #300, opened 2022-02-23
updated mentioned version

DJOetzi:
updated the mentioned DPP version in the guides


Commandserver:
is there no automation for this process?

braindigitalis:
no there is no automation as the documentation says "for example", doesnt mean the user should literally use this url, but should find the latest version :)
keeping them updated every so often cant hurt though.

PR #311, opened 2022-03-05
use a red warning box

es183923:
Use a huge red warning box that will hopefully allow readers to know that they should read the stuff inside it, and use the template, instead of blindly following the instructions.


es183923:
Looks like this now:
Screen Shot 2022-03-05 at 4 35 18 PM

PR #312, opened 2022-03-05
use a red warning box

es183923:
Use a huge red warning box that will hopefully allow readers to know that they should read the stuff inside it, and use the template, instead of blindly following the instructions.

Looks like this now:
Screen Shot 2022-03-05 at 4 35 18 PM

### DIFF
--- a/docpages/02_build.md
+++ b/docpages/02_build.md
@@ -59,7 +59,7 @@ To build on windows follow these steps *exactly*. The build process depends on s
 
 ## Wait a minute! Read this first!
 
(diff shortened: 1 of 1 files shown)


braindigitalis:
i took this out of a warning box, because people werent reading it...

es183923:
I think it's because it was yellow. Red is a lot more contrasting.

braindigitalis:
let's leave this open for a while as it seems since the last change people are following the docs to building the template, as this is the path of least resistance...

PR #318, opened 2022-03-11
fixed return types of several cluster methods

Commandserver:
- doc updates
- fixed return types of some cluster methods


braindigitalis:
you can't use map types with the non list template, this is why all the errors.

Commandserver:
changed it back

PR #323, opened 2022-03-19
doc updates

Commandserver:
refactored example program intents. make some method params optional


braindigitalis:
I can't accept this pr, the file you have edited is auto generated as it says in the comment at the top.

PR #328, opened 2022-03-21
Text inputs have max 100 char placeholders, not 150 chars

es183923:
According to the Discord API, text inputs have [100](https://discord.com/developers/docs/interactions/message-components#text-inputs-text-input-structure) characters as the palceholder limit, while select menus have [150](https://discord.com/developers/docs/interactions/message-components#select-menu-object-select-menu-structure) characters as the placeholder limit.

This PR sets the limit for text inputs to 100 chars instead of 150.

### DIFF
--- a/include/dpp/message.h
+++ b/include/dpp/message.h
@@ -237,7 +237,7 @@ public:
 	 */
 	std::string url;
 
-	/** Placeholder text for select menus and text inputs (max 150 characters)
+	/** Placeholder text for select menus and text inputs (max 150 characters for select menus, 100 characters for text inputs)
 	 */
 	std::string placeholder;
 
--- a/src/dpp/message.cpp
+++ b/src/dpp/message.cpp
@@ -317,9 +317,8 @@ select_option& select_option::set_animated(bool anim) {
 	return *this;
 }
 
-
 component& component::set_placeholder(const std::string &_placeholder) {
-	placeholder = dpp::utility::utf8substr(_placeholder, 0, 150);
+	placeholder = dpp::utility::utf8substr(_placeholder, 0, (this->type == cot_text) ? 100 : 150);
 	return *this;
 }
 


es183923:
Actually, there are more issues with the current implementation. Select menus have max 100 chars for the label, while the code forces it to be 80.

es183923:
How should the type automatically be set then?

braindigitalis:
the max of 150 is  set as the maximum of all types. There's no other sane way to do it.

es183923:
I've decided to add the string length trimming to the set_type function. Is that okay?

src/dpp/message.cpp
```diff
@@ -199,6 +224,9 @@ void to_json(json& j, const component& cp) {
 		j["label"] = cp.label;
 		j["required"] = cp.required;
 		j["style"] = int(cp.text_style);
+		if (std::holds_alternative<std::string>(cp.value) && std::get<std::string>(cp.value).empty()) {
```
Axyte, line 227: I think you missed a `!` on the second condition.

braindigitalis:
adding a check in set_type is only going to work if you set the description before the type otherwise it will do nothing

es183923:
I also have the check in the `set_label` and `set_placeholder` functions, not just in `set_type`.

include/dpp/message.h
```diff
@@ -341,6 +342,16 @@ class DPP_EXPORT component : public json_interface<component> {
...
 
+	/**
+	 * @brief Set the value of the modal component.
+	 * For action rows, this field is ignored. Setting the
+	 * value will auto-set the type to dpp::cot_text.
+	 *
+	 * @param value Value text to set. It will be truncated to the maximum length of 4000 UTF-8 characters.
+	 * @return component& Reference to self
+	 */
+	component& set_value(const std::string &val);
```
braindigitalis, line 353: the value is not just string, see the value member, there should be overloaded versions for int64 and double too if you're adding this

braindigitalis (changes requested):

include/dpp/message.h
es183923: But, the API specifies that the json only accepts a string.

braindigitalis: where though? this is in dpp component which is used in a few places?

es183923: This is just for setting the default value of the modal, not anywhere else. [source](https://discord.com/developers/docs/interactions/message-components#text-inputs).

braindigitalis: no  a modal can only use text inputs but there are other kinds of input, and elsewhere a text input component may contain text, integers or double:
https://discord.com/developers/docs/interactions/message-components#component-object-component-types

braindigitalis (approved):

PR #329, opened 2022-03-22
add set_guild_id to channel

es183923:
You can't set the guild_id using functions, so this PR adds it.

### DIFF
--- a/src/dpp/channel.cpp
+++ b/src/dpp/channel.cpp
@@ -87,6 +87,11 @@ channel& channel::set_topic(const std::string& topic) {
 	return *this;
 }
 
+channel& channel::set_guild_id(const snowflake guild_id) {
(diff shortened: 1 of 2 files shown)


src/dpp/channel.cpp
```diff
@@ -87,6 +87,11 @@ channel& channel::set_topic(const std::string& topic) {
 	return *this;
 }
 
+channel& channel::set_guild_id(const snowflake guild_id) {
+	this->guild_id = guild_id;
```
braindigitalis, line 91: the parameter name guild_id shadows the member variable (they have the same name)

braindigitalis (changes requested):

es183923:
This is consistent with what the other functions have. That's why its using `this->guild_id` to specify which `guild_id` it is referring to.

PR #331, opened 2022-03-24
break: Add parameter dpp::timer to timer_callback_t

rgnter:
For when you don't want to manage a lifetime of object which holds `dpp::timer` handle such as `dpp::oneshot_timer` and still be able to stop the timer without convoluted pointers to it's ID.

### DIFF
--- a/include/dpp/cluster.h
+++ b/include/dpp/cluster.h
@@ -3421,7 +3421,7 @@ public:
 		/* Attach event */
 		listener_handle = ev(listener);
 		/* Create timer */
-		th = cl->start_timer([this]() {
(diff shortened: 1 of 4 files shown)


es183923:
You should make this an optional parameter so you can still use the existing implementation, and it doesn't become a breaking change. Also, I think you might need to update or add some documentation on this.

rgnter:
> You should make this an optional parameter so you can still use the existing implementation, and it doesn't become a breaking change. Also, I think you might need to update or add some documentation on this.

Making this an optional parameter only helps when calling this function. This does not make things easier for the user who implements this function.

PR #334, opened 2022-03-25
faster and more reliable thread blocking in dpp::sync

rgnter:
relying on timer doesn't look right.


braindigitalis:
looks good, been meaning to do this, will merge this now!
thanks

PR #336, opened 2022-03-25
select_option fill from message's component was missing

Lohkdesgds:
on a select menu event, options were missing. This fixes that


braindigitalis:
select_option shoule now derive from `json_interface<select_option>`

braindigitalis (changes requested):
derive from json_interface template class

braindigitalis (approved):

PR #340, opened 2022-03-26
api: Ability to wait for cluster termination

rgnter:
Merge it if you like it.


rgnter:
Notifying the cv doesn't actually terminate the cluster, but that's for another API to implement.

PR #345, opened 2022-03-31
add even more missing default callbacks

es183923:


braindigitalis:
no no no

these dont have defaults as they're "get" methods, you'd never call them with no lambda, what would be the point?

PR #348, opened 2022-04-01
refactored dpp::guild & get_banner_url

Commandserver:
fixed `guild::get_icon_url` wasn't working correctly


src/dpp/guild.cpp
```diff
@@ -29,7 +29,7 @@
 
 using json = nlohmann::json;
 
-const std::map<std::string, dpp::guild_flags> featuremap = {
+const std::map<std::string, std::variant<dpp::guild_flags, dpp::guild_flags_extra>> featuremap = {
```
braindigitalis, line 32: does this work? is it only working on the guild_flags being uint32_t and guild_flags_extra being uint8_t right now?

PR #350, opened 2022-04-03
add set_channel_id and set_guild_id to message

es183923:
`set_channel_id` and `set_guild_id` were missing in `dpp::message`


Linker-123:
Maybe inline the functions?

es183923:
Why?

braindigitalis:
> Maybe inline the functions?

we don't inline them as they cause bloat of the executable, and make headers bigger which means slower compiles, the compiler can decide if to inline but bear in mind these functions live inside a dll/so which means they usually can't be inlined

PR #352, opened 2022-04-04
deps: BoringSSL fallback

Wykerd:
Fallback to building BoringSSL from source if OpenSSL is not installed on the system. BoringSSL was chosen as a fallback since it uses CMake as its build system, making it easy to integrate in this project, and it is fully compatible with the existing codebase.

I found this patch very useful on macOS since CMake never seems to finds the correct OpenSSL install on my system.

Feel free to close this PR if you do not find it useful.


braindigitalis:
why bundle boringssl with the lib though? it's massive...

es183923:
If you installed it through brew, make sure that the path to the openssl is in the `$PATH` variable. Also, if you are on an Apple silicon device, you may need to manually set the `CMAKE_APPLE_SILICON_PROCESSOR` variable, as Cmake doesn't set it properly.

PR #363, opened 2022-04-27
Properly set deaf and mute properties on member

es183923:
Before, it wouldn't get unset bc it wouldn’t be put into the JSON. This PR fixes it.


braindigitalis:
this is failing multiple build checks

D:\a\DPP\DPP\main\src\dpp\guild.cpp(202,14): error C3861: 'is_mute': identifier not found [D:\a\DPP\DPP\main\build\dpp.vcxproj]

can you fix please?

PR #365, opened 2022-04-30
Added ability to pass reaction emoji to all reaction methods without url encoding it

xXTurnerLP:
I had problems with passing emoji as unicode character to reaction methods (either discord API or D++ didn't like it), so I added the ability to disable url encoding to be able to pass raw value (for example the already url-encoded value for given emoji).

This is done via additional parameter before callback which is `bool url_encode`
I wanted to make it with default value `true` so you can omit it, but some reaction methods dont have default callback handler in order to force users to provide one, thus `url_encoded` needs to be provided in each call to these methods with either `true` or `false`


Linker-123:
What are the errors you are getting from `confirmation_callback_t::http_info::body` when sending url encoded emojis?

PR #371, opened 2022-05-03
Switch synchronized calls from a condition variable to a promise.

kar8uncle:
## Issue

The synced functions sometimes indefinitely block. All other threads are waiting for work to do, so it appears the condition variable is waited after the rest request is completed. The lambda probably needed to lock the mutex before signaling to ensure the calling thread started the wait.

Anyway, thought that promises and futures are cleaner so switched to using those. Calling thread is no longer indefinitely blocked.

Also refactored `json_interface` since I encountered some warnings and thought those are valid things to fix.

## Commit Messages

```txt
Switch synchronized calls from a condition variable to a promise.

Occasionally the synchronized call blocks indefinitely: the request
has been fulfilled before the calling thread starts waiting. This
can in theory happen in certain thread scheduling order. Switched to
using promises and delegated all the management of locking to the
implementation of std::promise.
```

```txt
Fix compiler warnings about virtual destructors.

This change fixes the valid warning -Wdelete-non-abstract-non-virtual-dtor
from clang.

json_interface<T> contains a virtual method yet has no virtual destructors.
Some of its subclasses do not inherit from other classes, so these
subclasses have no virtual destructors, which can in theory lead to
undefined behavior when those objects are destructed through the pointer
to their base class (json_interface<T>).

In practice, json_interface is a CRTP. The subclasses are never referenced
through the base class, so virtual methods aren't necessary; only method
overloading is needed. This change switches json_interface<T>::build_json
from virtual to nonvirtual. The subclass declares an overload under the
same name.

The destructor of json_interface is also explicitly marked protected now,
since subclasses should never be destructed through a pointer to
json_interface.
```

## Remarks

`dpp::guild_widget` in `guild.h` does not inherit from `json_interface<guild_widget>` yet has implemented the `build_json` method. Suggest taking a look at whether this is expected.

### DIFF
--- a/include/dpp/appcommand.h
+++ b/include/dpp/appcommand.h
@@ -341,7 +341,7 @@ struct DPP_EXPORT interaction_response : public json_interface<interaction_respo
 	 *
 	 * @return std::string JSON string
 	 */
-	virtual std::string build_json(bool with_id = false) const;
+	std::string build_json(bool with_id = false) const;
 
 	/**
 	 * @brief Add a command option choice
@@ -445,7 +445,7 @@ public:
 	 *
 	 * @return std::string JSON string
 	 */
-	virtual std::string build_json(bool with_id = false) const;
+	std::string build_json(bool with_id = false) const;
 
 	/**
 	 * @brief Destroy the interaction modal response object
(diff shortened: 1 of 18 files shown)


braindigitalis:
are you on the official discord? if so you should get the contributor, role ping me :)

kar8uncle:
> are you on the official discord? if so you should get the contributor, role ping me :)

Just joined!

PR #376, opened 2022-05-08
Improve voice receive handling.

kar8uncle:
## Issue

When there are multiple speakers, the voice data received from `on_voice_receive` callback was wrong. After lots of trial and error, it turns out Opus decoders are stateful, so using the same decoder to decode for all speakers jams the audio data together in a weird way. The main fix is to use a unique decoder for each speaker.

This change also contains a number of refactors as a result of the trial and error. I think even better we can handle packet loss by passing a null buffer to the Opus decoder now that we know it is stateful, but haven't gotten time to actually try it out yet.

## Commit Message

```txt
1. Allocate an Opus decoder for each speaker. Opus decoders are stateful,
   so when there are multiple speakers, using one decoder will mess up the
   voice data of every speaker.

2. Add some handling of out-of-order UDP packets. UDP packets can arrive
   out-of-order so it is up to the application to handle such cases. We
   now does reordering for packets and flush them to voice handlers every
   100ms. When the bot or client disconnects, unflushed packets are
   immediately flushed.

3. Improved the parsing of RTP packets, especially around the so-called
   "undocumented Discord byte", which is actually misinterpretation of
   the header length property. Other changes in the parsing are mostly
   refactoring.
```

### DIFF
--- a/.cspell.json
+++ b/.cspell.json
@@ -73,7 +73,8 @@
 		"wstype",
 		"ssrc",
 		"clion",
-		"followup"
+		"followup",
+		"gifv"
 	],
 	"flagWords": [
 		"hte"


include/dpp/discordvoiceclient.h
```diff
@@ -151,11 +154,100 @@ class DPP_EXPORT discord_voice_client : public websocket_client
...
+		 * @brief libopus decoder
+		 */
+		std::unique_ptr<OpusDecoder,
+		                void(*)(OpusDecoder*)
+		                > decoder{nullptr, nullptr};
+#endif
+	};
+	/**
+	 * @brief Voice buffers to be reported to handler, grouped by speaker.
+	 * Buffers are parked here and flushed when the parking lot hits 500 elemenets.
```
kar8uncle, line 219: Oh this had a typo and the comment is also wrong (old comment). I've changed it to flush every 100ms.

kar8uncle:
The Codacy Static Analysis is a false positive. `creator` is a member of `this` so to use it with the same name I would have to either capture `this`, or as is done here, explicitly capture the result of the expression `creator` and name it `creator`, which is this allegedly problematic syntax.

braindigitalis:
looks like it doesn't build on windows right now though, any chance you can fix that please?

braindigitalis:
Constructor is deleted:

```
C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Tools\MSVC\14.31.31103\include\xmemory(680,47): error C2280: 'dpp::discord_voice_client::voice_payload::voice_payload(const dpp::discord_voice_client::voice_payload &)': attempting to reference a deleted function (compiling source file D:\a\DPP\DPP\main\src\dpp\discordvoiceclient.cpp) [D:\a\DPP\DPP\main\build\dpp.vcxproj]
```

kar8uncle:
I am trying to make it handle packet loss. Hang tight.

PR #378, opened 2022-05-09
Improve voice receive handling.

kar8uncle:
## Issue

When there are multiple speakers, the voice data received from `on_voice_receive` callback was wrong. After lots of trial and error, it turns out Opus decoders are stateful, so using the same decoder to decode for all speakers jams the audio data together in a weird way. The main fix is to use a unique decoder for each speaker, but this change also contains a number of refactors and fixes as a result of the trial and error.

## Commit Message

```txt
1. Allocate an Opus decoder for each speaker. Opus decoders are stateful,
   so when there are multiple speakers, using one decoder will mess up the
   voice data of every speaker.

2. Add handling for out-of-order UDP packets. UDP packets can arrive
   out-of-order so it is up to the application to handle such cases. We
   now does reordering for packets and flush them to voice handlers every
   100ms.

3. Add handling for lost UDP packets. Opus decoder allows passing nullptr
   to indicate a packet loss. The decoder can insert some correction, e.g.
   fade out, in place of an abrupt stop that creates popping sounds. When we
   detect a packet loss, we let Opus know and handle that for us.

4. Improve the parsing of RTP packets, especially around the so-called
   "undocumented Discord byte", which is actually misinterpretation of
   the header length property. Other changes in the parsing are mostly
   refactoring.
```

### DIFF
--- a/src/dpp/discordvoiceclient.cpp
+++ b/src/dpp/discordvoiceclient.cpp
@@ -18,7 +18,7 @@
  * limitations under the License.
  *
  ************************************************************************************/
-#include <string>
+#include <string_view>
 #include <iostream>
 #include <fstream>
 #ifndef WIN32
@@ -29,7 +29,6 @@
 #include <dpp/discordvoiceclient.h>
 #include <dpp/cache.h>
 #include <dpp/cluster.h>
-#include <thread>
 #include <dpp/nlohmann/json.hpp>
 #include <dpp/fmt-minimal.h>
 
@@ -44,6 +43,8 @@ struct OpusRepacketizer {};
 
 namespace dpp {
 
+constexpr int32_t opus_sample_rate_hz = 48000;
+constexpr int32_t opus_channel_count = 2;
 std::string external_ip;
 
 /**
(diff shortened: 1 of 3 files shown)


src/dpp/discordvoiceclient.cpp
```diff
@@ -44,6 +43,8 @@ struct OpusRepacketizer {};
 
 namespace dpp {
 
+constexpr opus_int32 opus_sample_rate_hz = 48000;
```
braindigitalis: reference to opus_int32 must be wrapped in #HAVE_VOICE
same for any other references to opus or sodium or the lib won't build if voice support is not available

braindigitalis (changes requested):

src/dpp/discordvoiceclient.cpp
kar8uncle: Yeah I realized. Changed to `int32_t`.

kar8uncle:
Doing some more testing.

kar8uncle:
Found an issue where the voice receive event from a lost packet didn't have the `voice_client` property set.

braindigitalis (approved):

PR #382, opened 2022-05-13
Support gain control per user.

kar8uncle:
## Issue

The voice I receive has rather low amplitude. Didn't want to do audio processing myself. Found that Opus decoder supports native gain control, so made this change to allow clients to set the gain, just like how the Discord client allows setting individual volumes for each speaker. This also affects the combined audio, which means clients can use this API to mute a certain user in the combined audio as well.

## Commit Message

```txt
Similar to the User Volume slider in the Discord client, the bot
can choose the volume that it wants to hear a specific user at.
Uses native Opus decode gain control.
```

### DIFF
--- a/src/dpp/discordvoiceclient.cpp
+++ b/src/dpp/discordvoiceclient.cpp
@@ -29,6 +29,7 @@
 #include <iostream>
 #include <fstream>
 #include <algorithm>
+#include <cmath>
 #include <dpp/exception.h>
 #include <dpp/discordvoiceclient.h>
 #include <dpp/cache.h>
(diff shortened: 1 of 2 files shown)


src/dpp/discordvoiceclient.cpp
```diff
@@ -821,6 +830,49 @@ void discord_voice_client::error(uint32_t errorcode)
...
 }
 
+void discord_voice_client::set_user_gain(snowflake user_id, float factor)
+{
+#ifdef HAVE_VOICE
+	int16_t gain;
+
+	if (factor < 0.0f) {
+		/* Invalid factor; must be nonnegative. */
+		return;
```
braindigitalis, line 840: should these invalid inputs throw dpp::exception derivatives instead? can't decide myself here.

kar8uncle: I am okay either way. Personally prefer more graceful error handling but throwing exceptions seems to be the common pattern in dpp.

braindigitalis: let's keep it as it is for now... it seems to me that programmatically specifying a negative volume simply by floating point error is a possibility and we shouldn't throw exceptions for that, in this case graceful failure is better.

PR #395, opened 2022-05-26
permissions rework

Commandserver:
- **renamed `dpp::role_permissions` to `dpp::permissions`**, but added `using role_permissions = permissions;` so it shouldn't be too much breaking. I also moved it to a new file `permissions.cpp`.

- **Added constructor for `dpp::permission_overwrite`**

- **The has_-getters in `dpp::role` are now using the utility methods of `dpp::permission`**

### Added new `dpp::permission` class
The class holds an `uint64_t` and has some utility methods to check, remove and add flags.
This new class is now used for permission bitmasks and i replaced the old uint64_t with it on various places.

### DIFF
--- a/include/dpp/appcommand.h
+++ b/include/dpp/appcommand.h
@@ -468,7 +468,7 @@ struct DPP_EXPORT command_resolved {
 	/**
 	 * @brief Resolved total guild member permissions in the channel, including overwrites
 	 */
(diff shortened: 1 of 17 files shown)


Commandserver:
`permission::has` now uses `(mask & flag) == flag`

Commandserver:
Ready to merge

PR #401, opened 2022-05-28
added missing include in permissions.h

Commandserver:
it didn't compile when i just include <dpp/permissions.h> into my project


braindigitalis:
you shouldnt just include permissions.h into your project.
adding this include here may drastically increase compile time of dpp, as nlohmann json is a big header only lib and we try to avoid it being included in headers itself... does it have any negative effects?

PR #404, opened 2022-05-28
rewrite permission::has

Commandserver:


include/dpp/permissions.h
```diff
@@ -117,19 +117,23 @@ class DPP_EXPORT permission {
...
-	 * bool is_mod = permission.has(dpp::p_kick_members | dpp::p_ban_members);
+	 * bool is_mod = permission.has(dpp::p_kick_members, dpp::p_ban_members);
 	 * // Returns true if the permission bitmask contains p_kick_members and p_ban_members
 	 * ```
 	 *
-	 * @return True if it has the permission
+	 * @return bool True if it has all the given permissions
 	 */
-	bool has(uint64_t p) const;
+	template <typename... T>
```
braindigitalis, line 133: doesn't this need to use enable_if like the other methods do now?

PR #413, opened 2022-06-09
breaking: removed reason parameter from cluster::guild_ban_add method 

Commandserver:


src/dpp/cluster_sync_calls.cpp
```diff
@@ -216,8 +216,8 @@ auditlog cluster::guild_auditlog_get_sync(snowflake guild_id) {
 	return dpp::sync<auditlog>(this, &cluster::guild_auditlog_get, guild_id);
 }
 
-confirmation cluster::guild_ban_add_sync(snowflake guild_id, snowflake user_id, uint32_t delete_message_days, const std::string &reason) {
-	return dpp::sync<confirmation>(this, &cluster::guild_ban_add, guild_id, user_id, delete_message_days, reason);
+confirmation cluster::guild_ban_add_sync(snowflake guild_id, snowflake user_id, uint32_t delete_message_days) {
+	return dpp::sync<confirmation>(this, &cluster::guild_ban_add, guild_id, user_id, delete_message_days);
```
braindigitalis, line 220: I'm not sure I would have done this.
instead id have set the audit reason in this method and kept the parameter, how often do you ban with no reason?

PR #419, opened 2022-06-17
Added voice action bitflag for guild_member fixed #417

### LINKED ISSUE #417: cluster::guild_edit_member throws `Target user is not connected to voice`
**Git commit reference**
7e61fdcb02b5f7a021cf3fcdd10e8c6134d61a96

**Describe the bug**
when calling `cluster::guild_edit_member`, discord tells
> Target user is not connected to voice

**To Reproduce**
call the method and pass a member who is not connected to voice

**Expected behavior**
edits the member (e.g. changes the nickname).

**Additional context**
seems to be the `mute` and `deaf` [json params](https://discord.com/developers/docs/resources/guild#modify-guild-member) that causes this issue because they're set even if the member is not in voice.

unsigned-ru:
set_deaf and set_mute now set a bitflag, guild_member_update only sets deaf and mute parameters if the flag is on.
Provides a fix for the bug i found and that is described in #417

### DIFF
--- a/src/dpp/guild.cpp
+++ b/src/dpp/guild.cpp
@@ -108,11 +108,13 @@ guild_member& guild_member::set_nickname(const std::string& nick) {
 
 guild_member& guild_member::set_mute(const bool is_muted) {
 	this->flags = (is_muted) ? flags | gm_mute : flags & ~gm_mute;
+	this->flags |= gm_voice_action;
 	return *this;
 }
 
 guild_member& guild_member::set_deaf(const bool is_deafened) {
 	this->flags = (is_deafened) ? flags | gm_deaf : flags & ~gm_deaf;
+	this->flags |= gm_voice_action;
 	return *this;
 }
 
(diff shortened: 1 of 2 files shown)


src/dpp/guild.cpp
```diff
@@ -200,8 +203,12 @@ std::string guild_member::build_json(bool with_id) const {
 			j["roles"].push_back(std::to_string(role));
 		}
 	}
-	j["mute"] = is_muted();
-	j["deaf"] = is_deaf();
+
+	if (flags & gm_voice_action) {
+		j["mute"] = is_muted();
+		j["deaf"] = is_deaf();
```
braindigitalis, line 209: you should probably clear gm_voice_action here after setting the values

Commandserver: Why would it be harmful?

unsigned-ru: Think it would be actually the best idea to clear it there yeah

Commandserver: Well then clear it in guild_edit_member

unsigned-ru: Can't, the guild member is passed as const reference

PR #428, opened 2022-06-21
fix: make execute_webhook supports avatar

yuto0214w:
1. It wasn't possible to edit webhook to change avatar, so edited it to use `wh.avatar`
2. This will override message.author.username if `wh.name` present

### DIFF
--- a/src/dpp/cluster/webhook.cpp
+++ b/src/dpp/cluster/webhook.cpp
@@ -78,12 +78,12 @@ void cluster::execute_webhook(const class webhook &wh, const struct message& m, 
 		{"thread_id", thread_id},
 	});
 	std::string body;
-	if (!thread_name.empty()) { // only use json::parse if thread_name is set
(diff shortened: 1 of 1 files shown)


yuto0214w:
Thanks for the feedback.
Fixed code style.
But for support of uploading avatar, as `avatar_url` only accepts http(s), it isn't possible to do like so other than modifying webhook.

braindigitalis:
no pretty sure you can send a base64 encoded image in the image data, rather than a remote URL, same as when modifying webhook

yuto0214w:
I tested dozen times, and read documentation but I couldn't seem to upload avatar without modifying webhook.
But as webhook avatar url format is `https://cdn.discordapp.com/avatars/:id/:hash.png`, I still don't think it's possible to do

braindigitalis:
but those are user avatars, how do you get one uploaded to use it?

PR #436, opened 2022-07-01
fix: delete_webhook_with_token should send DELETE

yuto0214w:


braindigitalis:
👀 no iea what went off here. it should have merged it and closed it. its done something odd!
thanks anyway for the contribution

PR #437, opened 2022-07-02
docs: some typos

Commandserver:


include/dpp/channel.h
```diff
@@ -409,7 +409,6 @@ class DPP_EXPORT channel : public managed, public json_interface<channel>  {
 	 * Does not return reliable information for voice channels, use
 	 * dpp::channel::get_voice_members() instead for this.
 	 * @return A map of guild members keyed by user id.
-	 * @note If the guild this channel belongs to is not in the cache, the function will always return 0.
```
Commandserver, line 412: guilds are always cached, so this note is kinda useless isn't it?

braindigitalis: no this isn't useless, it may be called before the on_guild_create for the guild, if you give it a channel object you've constructed yourself with a hard coded id, or if the guild is unavailable when the bot starts due to an outage, even a threading race condition could cause it if you attach your own threads to on_guild_create with a worker pool.
worth keeping as its an outcome that can happen non the less.

docpages/03_example_programs.md
```diff
@@ -594,10 +594,10 @@ To add a slash command you should use the dpp::cluster::global_command_create me
 or dpp::cluster::guild_command_create to create a local command (available only to one guild).
 
 When a user issues these commands the reply will arrive via the `on_slashcommand` event which you can hook, and take action
-when you see your commands. It is possible to reply to an interaction by using either the dpp::interaction_create_t::reply method,
+when you see your commands. It is possible to reply to an interaction by using either the dpp::slashcommand_t::reply method,
```
braindigitalis, line 597: technically the reply method is in the parent class interaction_create_t so this will properly ref link to the method docs. changing this might break it as the reply method is inherited into slashcommand_t

Commandserver: alright, ill revert it

include/dpp/channel.h
Commandserver: gotcha, changed back

PR #443, opened 2022-07-14
Adding support for Vcpkg.

RealTimeChris:
No need to do that stuff any more!


braindigitalis:
the one conflict in this file should be resolved by accepting the change from the dev branch which moves the #endif below the `close()` call.

braindigitalis:
is this ready now? you keep pushing changes...

PR #449, opened 2022-07-23
added a check if the channel_id in the guild_member_move method is 0...

mobergmann:
…to disconnect the user


braindigitalis:
thanks, merged!

PR #454, opened 2022-07-29
This is a pr

Axyte:


braindigitalis:
just a passing thought should it be _co not coro? because the keywords start co, e.g. co_await, co_return...

PR #459, opened 2022-08-07
added some unit tests for dpp::user and dpp::utility

Commandserver:


src/test.cpp
```diff
@@ -207,6 +207,57 @@ Markdown lol \\|\\|spoiler\\|\\| \\~\\~strikethrough\\~\\~ \\`small \\*code\\* b
...
+		dpp::user user1;
+		user1.id = 189759562910400512;
+		user1.discriminator = 0001;
+		user1.username = "brain";
+
+		set_test("USER.GET_MENTION", false);
+		set_test("USER.GET_MENTION", user1.get_mention() == "<@189759562910400512>");
+
+		set_test("USER.FORMAT_USERNAME", false);
+		set_test("USER.FORMAT_USERNAME", user1.format_username() == "brain#0001");
```
braindigitalis, line 220: this should really fail... usernames are case sensitive, it's possible to have both brain#0001 and Brain#0001 at the same time (I know because I met the lower case me!)

braindigitalis (changes requested):

PR #460, opened 2022-08-08
Added get_user() shrotcut for dpp::guild_member

AA1999:


include/dpp/guild.h
```diff
@@ -259,6 +260,14 @@ class DPP_EXPORT guild_member {
 	 */
 	guild_member& set_nickname(const std::string& nick);
 
+	/**
+	 * @brief Get the dpp::user object for this member
+	 * @return dpp::user user object
```
braindigitalis, line 265: needs to indicate that if the user isn't in cache it returns nullptr

braindigitalis (changes requested):

include/dpp/guild.h
AA1999: Added

braindigitalis (approved):

PR #475, opened 2022-08-23
Fix small issue in guild::build_json()

BaderEddineOuaich:
json field `afk_timeout` addition was depending on another field's validity.


braindigitalis:
thanks for the contribution!

PR #480, opened 2022-08-26
Added a position comparison operator for dpp::role

AA1999:
Used so that if someone wants to sort a vector of roles they can include the < > operators


AA1999:
Wait mb this was supposed to be done into the dev branch.

PR #489, opened 2022-08-29
breaking: delete message count on `cluster::guild_ban_add` is now scaled in seconds instead of days

Commandserver:
https://github.com/discord/discord-api-docs/commit/7e3e3a3de46a909c9a5bc04404f0df9a806e7271

### DIFF
--- a/include/dpp/cluster.h
+++ b/include/dpp/cluster.h
@@ -2340,11 +2340,11 @@ public:
 	 * @note This method supports audit log reasons set by the cluster::set_audit_reason() method.
 	 * @param guild_id Guild ID to add ban to
 	 * @param user_id User ID to ban
-	 * @param delete_message_days How many days of their user's messages to also delete (0-7). Defaults to 0
+	 * @param delete_message_seconds How many seconds to delete messages for, between 0 and 604800 (7 days). Defaults to 0
 	 * @param callback Function to call when the API call completes.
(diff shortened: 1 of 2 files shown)


Commandserver:
hm should we add something in the method to log a warning message when a value of 0 to 7 is passed to `delete_message_seconds`? So that the users know they should take a look at it

braindigitalis:
> hm should we add something in the method to log a warning message when a value of 1 to 7 is passed to `delete_message_seconds`? So that the users know they should take a look at it

yes sounds.sensible

Commandserver:
We can also multiply the value, if its 1 - 7, by 86400 seconds. So basically it will calculate the value to seconds. I mean who tf will ban a player and delete the messages of lets say the last 5 seconds. Think we can asume then, that they've not updated their code.

Commandserver:
Or both, logging and multiplying

braindigitalis:
definitely just the warning - changing the value is confusing and what if you actually want a 5 second delete range?

Commandserver:
can you do the logging stuff for me @braindigitalis ?

Commandserver:
i dont really know how to do that

braindigitalis:
```cpp
this->log(ll_warn, "Oopsy woopsy you did a fucky wucky!");
```
this, inside any cluster method. not that literal text though lol

PR #493, opened 2022-09-01
Implement the Meson build system

DrBluefall:
This is intended as a replacement to the current CMake build system. Mostly because the alterations I had to make, however small they may be, currently break builds in CMake without manual intervention.

I'm also admittedly unsure how to handle VCPKG support, since... well, the origin of this effort was my own frustrations with CMake & VCPKG.

I attempted to make as few changes as possible, but some still had to be made:

- `src/unittest.cpp` now reads `Robot.pcm` from a different path. (Funnily enough, it's the path that's actually documented in `docpages/example_programs/music_and_audio/soundboard.md`.)

- `src/dpp/utility.cpp`, `src/dpp/discordclient.cpp`, and `src/dpp/discordvoiceclient.cpp` now include `config.h`, a file generated by Meson. (This is why CMake builds are broken.)

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1,0 +1,284 @@
+project('dpp',
+        'cpp',
+        version : '10.0.7',
+        default_options : ['warning_level=3', 'cpp_std=c++17'])
+
+# Filesystem support
+fs = import('fs')
+# Pkg-Config file generation
+pkgconf = import('pkgconfig')
+
+cfg = configuration_data()
+cfg.set('DPP_BUILD', 1)
+
+# Generated with fd . src/dpp -e cpp -x echo "'{}',"
+src = [
+  'src/dpp/cluster_sync_calls.cpp',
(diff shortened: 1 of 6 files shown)


DrBluefall:
There's probably (read: extremely likely) something I forgot in the build definition, so assistance in dotting those i's and crossing those t's would be appreciated.

braindigitalis:
You have opened a PR against the master branch. PRs must target the `dev` branch, as such this PR has been automatically closed. Please re-target your PR against the dev branch if you reopen it. Thank you for your contribution.

eli-schwartz:
> * now include `config.h`, a file generated by Meson. (This is why CMake builds are broken.)

Something I often see in autotools projects, is:

```c
#ifdef HAVE_CONFIG_H
    #include "config.h"
#endif
```

Then the build system will pass this define to make the config header used everywhere relevant.

PR #494, opened 2022-09-01
Implement Meson build system

DrBluefall:
This is intended as a replacement to the current CMake build system.
Mostly because the alterations I had to make, however small they may be,
currently break builds in CMake without manual intervention.

Also, VCPKG support is missing, mostly due to my inexperience on this
side of the equation, and the origins of this PR being my own
frustrations with VCPKG & CMake.

I attempted to make as few changes as possible, but some still had to be
made:

- `src/unittest.cpp` now reads `Robot.pcm` from a different
path. (Funnily enough, it's the path that's actually documented in
`docpages/example_programs/music_and_audio/soundboard.md`.)

- `src/dpp/utility.cpp`, `src/dpp/discordclient.cpp`, and
`src/dpp/discordvoiceclient.cpp` now include `config.h`, a file
generated by Meson. (This is why CMake builds are broken.)

*one cherry pick later...*

Now filed & retested against the dev branch.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1,0 +1,284 @@
+project('dpp',
+        'cpp',
+        version : '10.0.7',
+        default_options : ['warning_level=3', 'cpp_std=c++17'])
+
+# Filesystem support
+fs = import('fs')
+# Pkg-Config file generation
+pkgconf = import('pkgconfig')
+
+cfg = configuration_data()
+cfg.set('DPP_BUILD', 1)
+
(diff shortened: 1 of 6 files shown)


braindigitalis:
You have opened a PR against the master branch. PRs must target the `dev` branch, as such this PR has been automatically closed. Please re-target your PR against the dev branch if you reopen it. Thank you for your contribution.

PR #498, opened 2022-09-04
fix uninitialized warnings fields in dpp.h

Commandserver:


include/dpp/dpp.h
```diff
@@ -81,11 +81,11 @@ class collected_reaction : public managed {
 	/// Reacting user
 	user react_user;
 	/// Reacting guild
-	guild* react_guild;
+	guild* react_guild{};
```
braindigitalis, line 84: as these are pointers they should probably initialise to nullptr explicitly. {} doesn't do this right?

es183923: brace initialization initializes pointers to `nullptr`.

PR #500, opened 2022-09-06
Added unverified bots default intent.

AA1999:
Same as `i_default_intents` but also has `i_message_content` intent.

### DIFF
--- a/include/dpp/intents.h
+++ b/include/dpp/intents.h
@@ -79,7 +79,7 @@ enum intents {
 	i_privileged_intents		= dpp::i_guild_members | dpp::i_guild_presences | dpp::i_message_content,
 	/// Every single intent
 	i_all_intents			= dpp::i_default_intents | dpp::i_privileged_intents,
-        // Unverified bots default intents
+        /// Unverified bots default intents
         i_unverified_default_intents    = dpp::i_default_intents | dpp::i_message_content
 };
 
--- a/include/dpp/intents.h
+++ b/include/dpp/intents.h
@@ -78,7 +78,9 @@ enum intents {
 	/// Privileged intents requiring ID
 	i_privileged_intents		= dpp::i_guild_members | dpp::i_guild_presences | dpp::i_message_content,
 	/// Every single intent
-	i_all_intents			= dpp::i_default_intents | dpp::i_privileged_intents
+	i_all_intents			= dpp::i_default_intents | dpp::i_privileged_intents,
+        // Unverified bots default intents
+        i_unverified_default_intents    = dpp::i_default_intents | dpp::i_message_content
 };
 
 };


include/dpp/intents.h
```diff
@@ -78,7 +78,9 @@ enum intents {
 	/// Privileged intents requiring ID
 	i_privileged_intents		= dpp::i_guild_members | dpp::i_guild_presences | dpp::i_message_content,
 	/// Every single intent
-	i_all_intents			= dpp::i_default_intents | dpp::i_privileged_intents
+	i_all_intents			= dpp::i_default_intents | dpp::i_privileged_intents,
+        // Unverified bots default intents
```
Commandserver, line 82: Put a triple slash there, otherwise it's just a normal comment

Commandserver:
Does unverified bots have message content intent enabled by default? 🤔

braindigitalis:
no, this only applies to unverified bots added before the change not new ones created today 

the message content is still a privileged intent and should not be made default by dpp

include/dpp/intents.h
AA1999: Done

AA1999:
Any unverified bot that is below 100 server should have no problems using message intents tho, right?
Discord made it clear it's only a thing for verified bots anyways

braindigitalis:
> Any unverified bot that is below 100 server should have no problems using message intents tho, right? Discord made it clear it's only a thing for verified bots anyways (or the bots at 100 servers)

yes but it is still a privileged intent and you still have to go into developer page and turn it on... we would be introducing issues later when the bot grows and they then require approval to use the intent...

AA1999:
It is called `unverified bots intent` for a reason

braindigitalis:
still shouldn't be defaulted imho, it becomes something else later and people will get used to having it enabled

braindigitalis:
hmm this actually looks different to what I saw before.... I'll merge it later

Axyte:
> still shouldn't be defaulted imho, it becomes something else later and people will get used to having it enabled 

Agreed with this statement here.

Axyte:
It's really unnecessary and in-fact over-thought imo.

AA1999:
It's not replacing the `default_intents`
It's just a handy alternative for people making bots that are private/never gonna reach 100 servers

Axyte:
> It's not replacing the `default_intents` It's just a handy alternative for people making bots that are private/never gonna reach 100 servers

I'm aware it's not replacing that, but I still feel the same. Rest is up to @braindigitalis.

Commandserver:
> It's really unnecessary and in-fact over-thought imo.

I agree tbh

AA1999:
We already have a `all` intents that people might wrongly get used to.
We could maybe rename it better but I feel like it's not really a bad addition

Commandserver:
> We already have a `all` intents that people might wrongly get used to.
> We could maybe rename it better but I feel like it's not really a bad addition

I would suggest `i_defaults_and_message_intents`

es183923:
That is literally just as verbose as `i_default_intents | i_message_content`

AA1999:
Idk just feels like having aliases is fine


> > We already have a `all` intents that people might wrongly get used to.
> > We could maybe rename it better but I feel like it's not really a bad addition
> 
> I would suggest `i_defaults_and_message_intents`

Seems fine to me, we'll let brain decide

AA1999:
Maybe `i_private_bots_default_intents`?

PR #515, opened 2022-09-24
forum

Commandserver:
https://discord.com/developers/docs/resources/channel#message-object-message-flags


include/dpp/message.h
```diff
@@ -956,7 +956,9 @@ enum message_flags {
 	/// this message is only visible to the user who invoked the Interaction
 	m_ephemeral = 1 << 6,
 	/// this message is an Interaction Response and the bot is "thinking"
-	m_loading = 1 << 7
+	m_loading = 1 << 7,
+	/// this message failed to mention some roles and add their members to the thread
+	m_failed_to_mention_some_roles_in_thread = 1 << 8,
```
braindigitalis, line 961: maybe m_thread_mention_failed is better? m_failed_to_mention_some_roles_in_thread is a massive mouthful...

braindigitalis (changes requested):

include/dpp/message.h
Commandserver: okay. changed!

Commandserver:
hm. does it fit in the flags attribute of uint8_t?

braindigitalis:
> hm. does it fit in the flags attribute of uint8_t?

no it doesnt, may need to go to uint16_t

Commandserver:
okay. the flags attribute is now uint16. looks okay to me

Commandserver:
i've also improved the alignment of the message attributes, which saves 16 bytes (for my setup)

include/dpp/channel.h
```diff
@@ -158,6 +187,51 @@ struct DPP_EXPORT thread_member
...
 
+/**
+ * @brief Specifies the emoji to use as the default way to react to a forum post. Exactly one of `emoji_id` and `emoji_name` must be set.
+ */
+struct DPP_EXPORT default_reaction_emoji {
+	/// The id of a guild's custom emoji
+	snowflake emoji_id;
+	/// The unicode character of the emoji
+	std::string emoji_name;
+};
```
Commandserver, line 198: we could also add a variant here, as they're not both used at the same time iirc. ?

braindigitalis: > we could also add a variant here, as they're not both used at the same time iirc. ?

no, its both for a custom emoji and just the name for a built in emoji

include/dpp/channel.h
```diff
@@ -136,6 +149,22 @@ struct DPP_EXPORT thread_metadata {
 	bool invitable;
 };
 
+/**
+ * @brief Default duration threads will stop showing in the channel list after the specified period of inactivity
+ */
+enum default_auto_archive_duration_t : uint8_t {
```
Commandserver, line 155: how's the naming?

Commandserver: should a add `_t`?

include/dpp/channel.h
```diff
@@ -158,6 +187,51 @@ struct DPP_EXPORT thread_member
 	 thread_member& fill_from_json(nlohmann::json* j);
 };
 
+/**
+ * @brief Specifies the emoji to use as the default way to react to a forum post. Exactly one of `emoji_id` and `emoji_name` must be set.
+ */
+struct DPP_EXPORT default_reaction_emoji {
```
Commandserver, line 193: does this need a ctor?

braindigitalis: no it doesnt, its a simple struct with types that can be easily initialised

include/dpp/channel.h
braindigitalis: naming looks good to me

Commandserver:
How's the idea of storing the `default_sort_order` and the `default_auto_archive_duration` in the flags, like the channel type is stored. This would save 16 bytes, but is hacky to do :/

src/dpp/cluster/thread.cpp
```diff
@@ -67,6 +67,22 @@ void cluster::thread_members_get(snowflake thread_id, command_completion_event_t
...
+	json j({
+		{"name", forum_name},
+		{"auto_archive_duration", auto_archive_duration},
+		{"rate_limit_per_user", rate_limit_per_user},
+		{"message", json::parse(forum_message.build_json())},
+		{"applied_tags", applied_tags},
+	});
+	this->post_rest_multipart(API_PATH "/channels", std::to_string(channel_id), "threads", m_post, j.dump(), [this, callback](json &j, const http_request_completion_t& http) {
+		if (callback) {
+			callback(confirmation_callback_t(this, channel().fill_from_json(&j), http));
```
Commandserver, line 81: The discord docs says that it
> returns a channel, with a nested message object

**https://discord.com/developers/docs/resources/channel#start-thread-in-forum-channel**
How to do that here?

include/dpp/channel.h
```diff
@@ -274,6 +361,14 @@ class DPP_EXPORT channel : public managed, public json_interface<channel>  {
...
 	channel& set_topic(const std::string& topic);
 
+	/**
+	 * @brief Set type of this channel object
+	 *
+	 * @param type Channel type to set
+	 * @return Reference to self, so these method calls may be chained
+	 */
+	channel& set_type(const channel_type& type);
+
```
Commandserver, line 371: added a `set_type` setter to set the type without messing around with the flags

include/dpp/channel.h
```diff
@@ -185,26 +256,30 @@ class DPP_EXPORT channel : public managed, public json_interface<channel>  {
...
+	 * Contains either nothing, the id of a guild's custom emoji or the unicode character of the emoji
 	 */
-	utility::iconhash icon;
+	std::variant<std::monostate, snowflake, std::string> default_reaction;
 
 	/**
-	 * @brief Channel banner (boost level locked)
+	 * @brief Channel icon (for group DMs)
 	 */
-	utility::iconhash banner;
```
Commandserver, line 196: removed banner field. Seems like discord silently removed it. It shouldn't break anyone's code because discord didn't even released an endpoint to request these channel banners kek

src/dpp/cluster/thread.cpp
braindigitalis: our api endpoint should return dpp::thread, the 'channels' inside a forum are just threads. We could add the embedded message into dpp::thread, in fact i think it already has capability for this as threads can be created in exactly the same way.

Commandserver: okay, ill add a message field in dpp::thread for it

include/dpp/cluster.h
```diff
@@ -3053,6 +3053,22 @@ class DPP_EXPORT cluster {
...
+	 * @param msg The message to start the thread with
+	 * @param auto_archive_duration Duration in minutes to automatically archive the thread after recent activity. Can be set to: `60`, `1440` (default), `4320`, `10080`
+	 * @param rate_limit_per_user amount of seconds a user has to wait before sending another message (0-21600); bots, as well as users with the permission manage_messages, manage_thread, or manage_channel, are unaffected
+	 * @param applied_tags List of IDs of forum tags (dpp::forum_tag) to apply to this thread
+	 * @param callback Function to call when the API call completes.
+	 * On success the callback will contain a dpp::thread object in confirmation_callback_t::value. On failure, the value is undefined and confirmation_callback_t::is_error() method will return true. You can obtain full error details with confirmation_callback_t::get_error().
+	 */
+	void thread_create_in_forum(const std::string& thread_name, snowflake channel_id, message& msg, uint16_t auto_archive_duration = 1440, uint16_t rate_limit_per_user = 0, std::vector<snowflake> applied_tags = {}, command_completion_event_t callback = utility::log_error());
```
Commandserver, line 3070: should we make the `auto_archive_duration` param to the new type `auto_archive_duration_t`?

braindigitalis: yeah then you can't put an invalid value in it

braindigitalis:
does this work with the sync/coro auto generation, with the optional parameters before the callback?
need to check it doesnt break the generation.

include/dpp/cluster.h
Commandserver: the downside is that we have to adapt the value to minutes in the method, something like:
```c++
if (auto_archive_duration == arc_1_hour) {
		j["auto_archive_duration"] = 60;
	} else if (auto_archive_duration == arc_1_day) {
		j["auto_archive_duration"] = 1440;
	} else if (auto_archive_duration == arc_3_days) {
		j["auto_archive_duration"] = 4320;
	} else if (auto_archive_duration == arc_1_week) {
		j["auto_archive_duration"] = 10080;
	}
```

braindigitalis: whats wrong with `switch` and `case`?

Commandserver: maybe someone wants to store the auto_archive_duration in their database, and using an anum makes it complicated idk. but maybe im thinking too much and this would be an edge case.

braindigitalis: yeah i think youre over thinking that :D

Commandserver: yeah, ill change that to the enum

Commandserver: done

braindigitalis (approved):

Commandserver:
> does this work with the sync/coro auto generation, with the optional parameters before the callback? need to check it doesnt break the generation.

i didn't get it to work. Can you do it for me when it's merged. I've no clue why it doesn't generate the method when compiling it

include/dpp/channel.h
```diff
@@ -78,7 +80,19 @@ enum channel_flags : uint8_t {
  */
 enum discord_channel_flags : uint8_t {
 	/// Thread pinned in a forum (type 15) channel
-	dc_pinned_thread = 0b00000001,
+	dc_pinned_thread = 1 << 1,
+	/// whether a tag is required to be specified when creating a thread in a forum channel. Tags are specified in the thread::applied_tags field.
+	dc_require_tag =   1 << 4,
```
Commandserver, line 85: this flag isn't properly implemented yet, as it needs the channel_flags to be uint16. It is there, but nothing more

Commandserver: maybe someone else can do it

PR #516, opened 2022-09-26
Forums

Axyte:


include/dpp/channel.h
```diff
@@ -157,6 +168,16 @@ struct DPP_EXPORT thread_member
 	 */
 	 thread_member& fill_from_json(nlohmann::json* j);
 };
+/** An object that represents a tag that is able to be applied to a thread in a GUILD_FORUM channel. */
+struct DPP_EXPORT forum_tag : public managed, public json_interface<forum_tag>
+{
+    /** the name of the tag (0-20 characters) */
```
braindigitalis, line 174: these appear to be indented with spaces instead of tabs?

Axyte: Nope, indented with tabs as usual.

include/dpp/channel.h
```diff
@@ -91,6 +92,16 @@ enum overwrite_type : uint8_t {
 	ot_member = 1
 };
 
+/**
+ * @brief types to sort posts in a forum channel
+ */
+enum sort_order_type : uint8_t {
```
Axyte, line 98: @braindigitalis do you want `_t` or is `_type` fine?

braindigitalis: probably should be _t like the rest of these, but i know there are some that arent _t on them

Axyte: The rest of the ones in that header are `_type`

PR #518, opened 2022-10-01
Allow disabling install target

j883376:
This PR allows users to optionally disable the CMake install target which is helpful when building the library statically inside of another project and installing is unnecessary.


braindigitalis:
I can merge this in but why would you just not type `make install`? the install step is not mandatory after all?

PR #536, opened 2022-10-30
Stabilize Tick Rate in sslclient

Anotra:
ssl client has
```c
1:  if (last_tick != time(nullptr)) {
2:    this->one_second_timer();
3:    last_tick = time(nullptr);
4:  }
```
The problem is, is that the time could change between the condition on line 1, and setting it on line 3. This can't be changed, as it would cause bugs in other places. The poll syscall further down can be reworked to wake up at the beginning of the second to make it less likely for the time difference on lines 1, and 3 to occur. Other parts of the code rely on specific times to operate, and they may be skipped over regardless, this just makes it far less likely to occur.

I'm aware, that sslclient is running on its own thread, and there may be many running, it might be best to stagger the wake up at a random interval `(beginning of second + rand(0 to 100) milliseconds)`.

### DIFF
--- a/src/dpp/sslclient.cpp
+++ b/src/dpp/sslclient.cpp
@@ -65,6 +65,7 @@
 #include <string>
 #include <iostream>
 #include <unordered_map>
+#include <chrono>
 #include <dpp/sslclient.h>
 #include <dpp/exception.h>
 #include <dpp/utility.h>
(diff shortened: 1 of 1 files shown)


braindigitalis:
thanks for the pr, it appears to not compile on all systems due to a missing include. can you fix this please?
one question though, how much more does the poll loop run than before? if it means that lots more cpu is eaten to ensure we don't skip a second this could cause issues at scale.

Anotra:
> how much more does the poll loop run than before?
A decent amount more, yea. I've fixed that with this new change.
It will now poll 1000ms if in the lower 600ms of the second, otherwise it will poll (time_left + time_left / 3 + 1). That means if it's 700ms into the curent second, (300 + 300 / 3 + 1) and will wake up 401ms later and being 101ms into the next second. It also doesn't wake up that much more than the original version now.

braindigitalis:
thanks, ive merged this. if you're on the discord, let me know and i'll give you a contributor role.

PR #537, opened 2022-10-30
Fixed various C++23 issues

WerWolv:
This PR fixes a few issues when building in C++23 mode using GCC.

- One instance of undefined behaviour with `std::string` which now correctly causes a compile error
- A few ambiguous conversions from json objects to strings

### DIFF
--- a/src/dpp/integration.cpp
+++ b/src/dpp/integration.cpp
@@ -116,7 +116,7 @@ bool integration::expiry_kicks_user() const {
 	return flags & if_expire_kick;
 }
 
-connection::connection() : id(0), revoked(false), verified(false), friend_sync(false), show_activity(false), visible(false) {
+connection::connection() : id({}), revoked(false), verified(false), friend_sync(false), show_activity(false), visible(false) {
 }
 
 connection& connection::fill_from_json(nlohmann::json* j) {
--- a/src/dpp/events/ready.cpp
+++ b/src/dpp/events/ready.cpp
@@ -43,7 +43,7 @@ std::mutex protect_the_loot;
  */
 void ready::handle(discord_client* client, json &j, const std::string &raw) {
 	client->log(dpp::ll_info, "Shard id " + std::to_string(client->shard_id) + " (" + std::to_string(client->shard_id + 1) + "/" + std::to_string(client->max_shards) + ") ready!");
-	client->sessionid = j["d"]["session_id"];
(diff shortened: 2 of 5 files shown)


src/dpp/integration.cpp
```diff
@@ -116,7 +116,7 @@ bool integration::expiry_kicks_user() const {
 	return flags & if_expire_kick;
 }
 
-connection::connection() : id(0), revoked(false), verified(false), friend_sync(false), show_activity(false), visible(false) {
+connection::connection() : id({}), revoked(false), verified(false), friend_sync(false), show_activity(false), visible(false) {
```
braindigitalis, line 119: i'm not sure this is more readable? Wouldnt `0ULL` be nicer than `{}`?

braindigitalis (commented):
looking good, i'd only change one simple thing for readability :)

braindigitalis:
will leave it up to you to decide if `{}` is more readable - let me know, and when ready i'll merge this in. Thanks!

src/dpp/integration.cpp
WerWolv: Initializing a string with `0` is undefined behaviour since 0 gets implicitly converted to `nullptr` and initializing a std::string is UB.

This has always been an issue but C++23 finally now adds a compile time check for it. `{}` makes the string be default initialized instead. If you want, you can completely remove that initialization as well

braindigitalis: id isnt a string is it? its a dpp::snowflake which is container for uint64_t?

WerWolv: https://github.com/brainboxdotcc/DPP/blob/master/include/dpp/integration.h#L140

In the connection class it is a string

es183923:
For something like
```cpp
connection::connection() : id(0), revoked(false), verified(false), friend_sync(false), show_activity(false), visible(false) {
```

you can just make it (and i would suggest this format)
```cpp
connection::connection() : id{}, revoked{}, verified{}, friend_sync{}, show_activity{}, visible{} {
```

bools are default initialized to false, strings are default initialized to be empty

braindigitalis:
thanks for the pr, if you're on the Discord let me know and I'll give you the contributor role

WerWolv:
No problem!
Just joined it, my handle is WerWolv#1337 :)

PR #542, opened 2022-10-31
Update install-linux-deb.md

spy-ware:
Added `sudo` on `install-linux-deb.md` .
I wasn't able to run this command without `sudo` .


braindigitalis:
the tutorial already says to do it as root. debian doesn't come setup with sudo as standard like ubuntu does, how you get to root is down to your system setup...

Issue #545, opened 2022-11-08
Build fails with misleading/incomplete error messages when pkg-config is not installed


tuokri:
**Git commit reference**
Commit hash `22bea16758a822db0d5287cea35c3b5c4dff787e`

**Describe the bug**
Build fails with little information when `libsodium` is installed but `pkg-config` is NOT installed.

**To Reproduce**
1. Install all the dependencies listed in README.md.
2. Build the project according to instructions in https://dpp.dev/buildlinux.html
3. Build fails with misleading log output

**Expected behavior**
1. Build succeeds with voice chat disabled
OR
2. Build fails with an informative error message

**Screenshots**

Misleading build log:
![image](https://user-images.githubusercontent.com/16985292/200650633-22655ebf-47c6-43ff-9aa3-11ffe2515749.png)

After a bit of digging, the culprit is found:
```diff
+++ b/cmake/FindSodium.cmake
@@ -46,7 +46,7 @@ endif()
 # UNIX
 if(UNIX)
   # import pkg-config
-  find_package(PkgConfig QUIET)
+  find_package(PkgConfig)
   if(PKG_CONFIG_FOUND)
     pkg_check_modules(sodium_PKG QUIET libsodium)
   endif()
```

Better error message:
![image](https://user-images.githubusercontent.com/16985292/200651192-de2ab45e-ea53-4a81-b6a0-3ba05c9a51ce.png)

After installing pkg-config with `sudo apt install pkg-config`:
![image](https://user-images.githubusercontent.com/16985292/200651538-529387c5-62de-4d3d-9c5c-bb4e648c640c.png)

**System Details:**
Debian WSL:
Distributor ID: Debian
Description:    Debian GNU/Linux 11 (bullseye)
Release:        11
Codename:       bullseye

**Additional context**
Suggestion: list `pkg-config` in README.md as a dependency for building with voice chat support (on Linux)

PR #546, opened 2022-11-08
Print status message from find_package(PkgConfig)

tuokri:
Related to #545. Not sure if this is just a Debian related issue (or any distro without pkg-config installed by default), but not silencing `find_package` output should help spot the issue earlier if someone else runs into this.

### DIFF
--- a/cmake/FindSodium.cmake
+++ b/cmake/FindSodium.cmake
@@ -46,7 +46,7 @@ endif()
 # UNIX
 if(UNIX)
   # import pkg-config
-  find_package(PkgConfig QUIET)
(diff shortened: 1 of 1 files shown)


braindigitalis:
thanks, FindSodium is actually provided by sodium as their standard way of detection, so I've avoided changing this file. it doesn't matter if it's changed though, as we aren't going to be synchronising it with upstream and will merge this instead, thanks for the PR!

This being said though perhaps soldium may also welcome this PR into their codebase too? Their support for cmake has never seemed correct and is just a shim around pkg-config as a silent dependency.

PR #554, opened 2022-11-16
[Update]: zlib version `1.2.13`

harshfeudal:
I have just noticed that zlib has been updated!

When I built it, I have just noticed that we have `zdll.lib` and idk if we need it or not. If not please tell me so that I can fix it!

Thanks!


braindigitalis:
where is the 32 bit zlib dll?
Was this built through vcpkg?

harshfeudal:
well i think i should close this. We will make zlib later ...

PR #558, opened 2022-11-23
Adding a new send_audio_type to handle audio gaps and stutters on win…

gnjax:
…dows

Added a new satype_overlap_audio type which propose a different implementation of audio packet throttling. Due to inaccuracies of sleep_for on windows, we cannot reliably pause the thread for the exact amount of time needed before the next packet, resulting in a slight overshoot of the time frame to keep the audio stream from stuttering. This implementation reduces the sleep time into multiple smaller increments and sleep as much as possible before the risk of sleeping too much occurs. It then adds the left over sleep duration to be added to the next packet's sleep duration, ensuring that we do not increase the gap by which we're ahead with each packet sent


braindigitalis:
thanks for the fix!

PR #559, opened 2022-11-23
Small fix in satype_overlap_audio

gnjax:
Prevent the sleep_increment from increasing over time


gnjax:
Sorry for the slight oversight, this is a minor fix that should allow playing really long audio without issues

Issue #560, opened 2022-11-25
[Bug?] Cannot catch all badges


harshfeudal:
**Describe the bug**
The Dpp library seems like cannot get all the user badge

**Expected behavior**
It will show all user badges, while the actual result only show the HypeSquad only

**System Details:**
 - OS: Windows 10
 - Discord Client used for testing [mobile, desktop, web]: both trio

**Additional context**
- This bug was found and tested by me and [@luk](https://github.com/luktvpl)

**Code**
https://github.com/Harshfeudal-Projects/Raiden-Shogun/blob/main/src/information/userInfo.cpp

**Screenshot**
My friend's account:
![image](https://user-images.githubusercontent.com/87577447/203880126-7ce9bf9f-c393-449c-bdab-4caf95dc2f48.png)

Mine account:
![image](https://user-images.githubusercontent.com/87577447/203880202-f6202892-a759-43b5-b1da-255057b969b8.png)

braindigitalis:
can't duplicate, please share code.

braindigitalis:
by code I mean a standalone test case demonstrating just the issue not an entire bot.

braindigitalis:
can you please confirm if #561 fixes this?

Commandserver:
Your problem probably is that you're using `dpp::find_user`. as this gives you only what is cached, the flags may be not all set. Try using `cluster::user_get`

harshfeudal:
Solved until Dpp version `10.0.22` ayy

PR #561, opened 2022-11-26
fix(issue #560): The bot does not return the entire user badge.

pneb:


Commandserver:
How can that fix a bug hardcoding a magic bitshift

braindigitalis:
fixes #560

braindigitalis:
thanks for the fix, when did this become a thing, public flags and flags were always the same bit masks...

PR #565, opened 2022-11-27
tests: added unit tests for role cluster methods, cluster::guild_get_member and user flags parsing

Commandserver:


src/unittest/test.cpp
```diff
@@ -410,6 +410,52 @@ Markdown lol \\|\\|spoiler\\|\\| \\~\\~strikethrough\\~\\~ \\`small \\*code\\* b
...
+								u.is_house_bravery() == 			((raw_flags & (1 << 6)) != 0) &&
+								u.is_house_brilliance() == 			((raw_flags & (1 << 7)) != 0) &&
+								u.is_house_balance() == 			((raw_flags & (1 << 8)) != 0) &&
+								u.is_early_supporter() == 			((raw_flags & (1 << 9)) != 0) &&
+								u.is_team_user() == 				((raw_flags & (1 << 10)) != 0) &&
+								u.is_bughunter_2() == 				((raw_flags & (1 << 14)) != 0) &&
+								u.is_verified_bot() == 				((raw_flags & (1 << 16)) != 0) &&
+								u.is_verified_bot_dev() == 			((raw_flags & (1 << 17)) != 0) &&
+								u.is_certified_moderator() == 		((raw_flags & (1 << 18)) != 0) &&
+								u.is_bot_http_interactions() == 	((raw_flags & (1 << 19)) != 0) &&
```
braindigitalis, line 445: i dont think we should hard code 1<< 18 etc into the code here. we should use the flags in `dpp/user.h`, no?
Our flags dont always match discord's flags and we can and do shuffle them about, which would break the unit tests.

src/unittest/test.cpp
```diff
@@ -464,6 +510,63 @@ Markdown lol \\|\\|spoiler\\|\\| \\~\\~strikethrough\\~\\~ \\`small \\*code\\* b
...
+							return;
+						}
+						dpp::role role = std::get<dpp::role>(event.value);
+						if (role.id == role_id) {
+							set_test("ROLE_EDIT", true);
+						}
+						bot.role_delete(TEST_GUILD_ID, role_id, [](const dpp::confirmation_callback_t &event) {
+							if (!event.is_error()) {
+								set_test("ROLE_DELETE", true);
+							}
```
braindigitalis, line 565: what happens if the role edit fails, so the guild is left created on the test server, the next time the role create runs this will fail permenantly until there is human intervention to delete it... we need to split out these so they arent dependant upon each other, or account for if the role already exists in role create test.

braindigitalis (changes requested):
changes as in comment to account for failure to edit leaving the role not deleted

src/unittest/test.cpp
Commandserver: True. I didn't think of that. I'll change that tomorrow

PR #574, opened 2022-12-06
Add unused-parameter warning, and fix resulting errors.

degustaf:


src/dpp/discordvoiceclient.cpp
```diff
@@ -1062,6 +1068,9 @@ size_t discord_voice_client::encode(uint8_t *input, size_t inDataSize, uint8_t *
 		throw dpp::voice_exception("Invalid input data length: " + std::to_string(inDataSize) + ", must be n times of " + std::to_string(mEncFrameBytes));
 	}
 #else
+	(void)input;
```
braindigitalis, line 1071: don't these generate another warning about a statement that has no effect?

degustaf: Not that I'm seeing. I was wary of adding `[[maybe_unused]]` because the arguments are used if the voice_client is enabled.

Honestly, it might make more sense for the function to only exist if `HAVE_VOICE` is defined.

braindigitalis:
is there no way to make this look less ugly?
what even is the point in enabling this warning, unused parameters don't have any runtime penalty.

braindigitalis:
just realised this makes a warning an error, why?
out of all the things that a compiler allows that should be stricter, I don't think this is one of them...

PR #581, opened 2022-12-13
added convertions from hex and cmyk colors to decimal value

ruslan-ilesik:
same as rbg function gives abilities to use colors of such formats in embeds.

### DIFF
--- a/include/dpp/utility.h
+++ b/include/dpp/utility.h
@@ -292,7 +292,7 @@ namespace dpp {
 		 * @param k key (black) value, between 0 and 1 inclusive
 		 * @return uint32_t returned integer colour value
 		 */
-		uint32_t DPP_EXPORT cmyk(double c, double m, double y,double k);
+		uint32_t DPP_EXPORT cmyk(double c, double m, double y, double k);
 		
 		/**
 		 * @brief Convert ints to CMYK for sending in embeds
@@ -303,7 +303,7 @@ namespace dpp {
 		 * @param k key (black) value, between 0 and 255 inclusive
 		 * @return uint32_t returned integer colour value
 		 */
-		uint32_t DPP_EXPORT cmyk(int c, int m, int y,int k);
+		uint32_t DPP_EXPORT cmyk(int c, int m, int y, int k);
 
 		/**
 		 * @brief Output hex values of a section of memory for debugging
--- a/include/dpp/utility.h
+++ b/include/dpp/utility.h
@@ -264,14 +264,14 @@ namespace dpp {
 		};
 
 		/**
-		 * @brief Convert floats to RGB for sending in embeds
+		 * @brief Convert doubles to RGB for sending in embeds
 		 * 
 		 * @param red red value, between 0 and 1 inclusive
 		 * @param green green value, between 0 and 1 inclusive
 		 * @param blue blue value, between 0 and 1 inclusive
 		 * @return uint32_t returned integer colour value
 		 */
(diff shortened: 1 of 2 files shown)


src/dpp/utility.cpp
```diff
@@ -226,7 +227,35 @@ namespace dpp {
...
+			if (value[0] == '#'){
+				value = value.substr(1,value.length()-1);
+			}
+			if (value.length() < 6){
+				throw  dpp::length_exception("Hexadecimal color is to short.");
+			}
+			if (value.length() > 6){
+				value = utf8substr(value,0,6);
+			}
+			const std::regex r{R"(^[0-9a-fA-F]{6}$)"};
```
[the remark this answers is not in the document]
ruslan-ilesik: Yes and no at the same time. if it will get something like ```1Z1234``` it will return 1 (reading all until incorrect format symbol) but if you do ```Z12345``` it will throw exception. I believe my option is better because user will see error straight forward, not some unknown behavior or error from time to time.

include/dpp/utility.h
```diff
@@ -275,14 +275,36 @@ namespace dpp {
...
+		/**
+		 * @brief Convert ints to CMYK for sending in embeds
+		 *
+		 * @param c cyan value, between 0 and 255 inclusive
+		 * @param m magenta value, between 0 and 255 inclusive
+		 * @param y yellow value, between 0 and 255 inclusive
+		 * @param k key (black) value, between 0 and 255 inclusive
+		 * @return uint32_t returned integer colour value
+		 */
+		uint32_t DPP_EXPORT cmyk(int c, int m, int y,int k);
```
Axyte, line 306: `int` has a 4-byte size typically and a much larger range than 0-255. Choose another type.

Axyte (changes requested):

include/dpp/utility.h
ruslan-ilesik: Brain sad to do it like that, so i did

ruslan-ilesik: Probably i can ask for char, so okay, i will change it

Axyte: Why `char`? Use a n-bit `int` type.

ruslan-ilesik: Also, if thinking that way i will also change rgb function. Because it has just int and getting values from 0-255.

ruslan-ilesik: here is same problem as with rgb, it do not work if it not int, so let it be it.

ruslan-ilesik: image

i will try to explain my opinion. with uint8_t (which is 1 byte int to store 0-255) you cant do simple calls like on picture. Moreover we already have rgb function which is using int for same range of numbers in input. Also In my opinion having one function with int another with uint8t will look strange.

braindigitalis: this is why we use int, because it will confuse the user when they try and use constant values. We dont want them to have to explicitly cast every time, or put `.0` on the end of each number...

braindigitalis:
Please DONT use `.` as a commit message, and repeat commit messages! These show in our changelogs! :( In the future please use proper descriptive commit messages.

ruslan-ilesik:
> Please DONT use `.` as a commit message, and repeat commit messages! These show in our changelogs! :( In the future please use proper descriptive commit messages.

Sorry. in future I will not use "."

Commandserver:
Are you uo to make a unit test for them?

PR #584, opened 2022-12-15
feat: added support for application role connection metadata

Commandserver:
Only tested `application_role_connection_get` and `application_role_connection_get`. Couldn't test the others yet.


Commandserver:
didn't tested it yet!

Issue #590, opened 2023-01-04
Issues with installed files


NexAdn:
**Git commit reference**
* tag v10.0.21
* 3ea77753f1d5d07110f4230d0471c8174ceb89fb (current master)
... but applies to most of the recent releases.

**Describe the bug**
A bunch of files required by CMake and pkgconfig get installed into the wrong directories, preventing said tools from finding the files or building against DPP correctly:

1. All headers are installed twice (to `/usr/include` and to `/usr/include/dpp-10.0`). This does not cause any problems. However, the second installed header tree is completely ignored by build systems and thus completely redundant.
2. The installed CMake modules aren't installed to standard CMake directories (`/usr/lib/cmake/dpp`), but rather to some custom versioned directory (`/usr/lib/dpp-10.0`). CMake is unable to find the packages there and thus fails builds of DPP's reverse dependencies if the package is installed in this way.
3. The shared library gets installed twice: `/usr/lib/libdpp.so(.10.0.21)` and `/usr/lib/dpp-10.0/libdpp.so(.10.0.21)`. Again, the second library is completely ignored as it's not in the linker's search path. Again, this does not prevent building, but only causes completely redundant files being installed.
4. ~~The pkg-config file sets `prefix = /usr/local`. However, on all major systems, files installed by the package manager don't belong into /usr/local but into /usr. DPP respects that, but it isn't reflected in the .pc file. The .pc file should set the prefix according to `${CMAKE_INSTALL_PREFIX}`.~~ Nevermind this one. On a git install, this actually works if CMAKE_INSTALL_PREFIX is set properly. Still, the prefix is set incorrectly at least in the deb package.
5. libdpp-config.cmake includes `${SELF_DIR}/libdpp.cmake`, but this file doesn't exist. Instead, dpp.cmake is probably the correct file that should be included and including it instead of libdpp.cmake actually resolves this issue.

**To Reproduce**

When using a prepackaged release:
1. Retrieve any .deb package from the latest release
2. Unpack the deb
3. Unpack the data tarball
4. `tree usr` and look at said files inside that tree

When building from source:
1. `mkdir dst`
2. `DESTDIR=dst ninja install` (or `make DESTDIR=dst install`)
3. `tree dst` and look at said files inside that tree

**Expected behavior**
Adhering to standard system directories, I expect the following directory tree to be in the installation:

* usr
  * include
    * dpp
      * (all header files here, optionally structured in subdirs)
  * lib
[...]

PR #591, opened 2023-01-04
Fixes to the DPP installlation on Linux

NexAdn:
This PR contains fixes for the CMake installation on Linux, removing redundant dirs, fixing `include()` statements and adhering to standard system directories, thus allowing reverse dependencies to find this package easily without any special configuration.

Note: These are the same fixes used in a Gentoo ebuild I use locally on my systems to install DPP. A similar set of patches (although not in the form of patch files) is applied in the AUR to make the package work. Thus, we have proof of concept that this patchset does what it should.

Closes: #590

### DIFF
--- a/cmake/CPackSetup.cmake
+++ b/cmake/CPackSetup.cmake
@@ -1,17 +1,15 @@
 include(GNUInstallDirs)
 set(DPP_EXPORT_NAME dpp)
-set(DPP_VERSIONED ${DPP_EXPORT_NAME}-${PROJECT_VERSION_MAJOR}.${PROJECT_VERSION_MINOR})
+set(DPP_CMAKE_DIR ${CMAKE_INSTALL_LIBDIR}/cmake/${DPP_EXPORT_NAME})
 set(DPP_VERSION_FILE ${PROJECT_BINARY_DIR}/${DPP_EXPORT_NAME}-config-version.cmake)
-set(DPP_INSTALL_INCLUDE_DIR ${CMAKE_INSTALL_INCLUDEDIR}/${DPP_VERSIONED})
-set(DPP_INSTALL_LIBRARY_DIR ${CMAKE_INSTALL_LIBDIR}/${DPP_VERSIONED})
 
 ## Pack the binary output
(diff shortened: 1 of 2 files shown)


braindigitalis:
Thanks for the fixes, for the purposes of changelog is this a breaking change? Would we potentially have people's systems depending on the oddball dpp-10.0 directory for finding the library and/or headers?
The original intent of this was for people who wanted to run dpp-9 alongside dpp-10 but we quickly realised this was not viable as discord dont maintain api v9 alongside v10 like this.

NexAdn:
I've seen that you've already merged this, but since some of your questions are important, let's try answer them:

> Thanks for the fixes, for the purposes of changelog is this a breaking change?

Yes and no. People who don't use the .pc or .cmake files provided could have had a successful build by linking using `/usr/lib/dpp-10.0/libdpp.so` instead of `-ldpp`. Those are the only people I imagine are affected by this change. It is still noteworthy in the changelog though.

For everyone else, the unpatched build probably didn't work anyway (CMake) or works as before (pkgconfig).

> Would we potentially have people's systems depending on the oddball dpp-10.0 directory for finding the library and/or headers?

As written above, this is theoretically possible, although highly unlikely. Still, I'd think most people would have just used -ldpp or `/usr/lib/libdpp.so` in the linker if they saw there was the correct lib. `/usr/lib` is the standard library dir, so it's just natural for people to look for the lib there.

> The original intent of this was for people who wanted to run dpp-9 alongside dpp-10 but we quickly realised this was not viable as discord dont maintain api v9 alongside v10 like this.

I don't know if the approach in the build system would have worked in this case.
The headers would be no problem (although you would have to take care of not duplicating the headers in `/usr/include/dpp`, since then you would get conflicting files), but the libraries still are a problem, since they are not in a standard library directory.
But as I said, this can be solved with `-Wl,-rpath`.
Still, in that case, just leaving out the libdpp.so symlink and letting SONAME do the magic would be easier.
For the CMake part, I don't know how that would've worked out.

But since that's the past now, we don't have to waste more brain time on this.

PR #592, opened 2023-01-08
feat: added setter for the channel::default_sort_order

Commandserver:


braindigitalis:
do we actually need a getter? seems redundant as its a simple variable which is public, not a bit value?

PR #603, opened 2023-01-22
misc: Guild flags updated

Commandserver:
- Removed the guild flags: `dpp::g_commerce` and `dpp::g_private_threads` and replaced them with the missing flags: `dpp::g_role_subscription_enabled` and `dpp::g_creator_store_page_enabled`
- Depreacted `dpp::g_seven_day_thread_archive`, `dpp::g_three_day_thread_archive` and `dpp::g_channel_banners` as discord removed them
- Added further guild flags: `dpp::g_no_role_subscription_notifications`, `dpp::g_no_role_subscription_notification_replies` and `dpp::g_role_subscriptions_available_for_purchase`
- I've also reordered the internal guild feature map to match discord's order


src/dpp/guild.cpp
```diff
@@ -436,7 +439,7 @@ guild& guild::fill_from_json(discord_client* shard, nlohmann::json* d) {
 	if (d->find("unavailable") == d->end() || (*d)["unavailable"].get<bool>() == false) {
 		/* Clear unavailable flag if set */
 		if (this->flags & dpp::g_unavailable) {
-			this->flags -= dpp::g_unavailable;
```
Commandserver, line 439: is this a feature or a bug?

braindigitalis: both do the same thing so long as the value you subtract is a power of two...

PR #605, opened 2023-01-25
V1

Commandserver:


src/dpp/cluster_sync_calls.cpp
```diff
@@ -317,10 +317,6 @@ confirmation cluster::guild_add_member_sync(const guild_member& gm, const std::s
 	return dpp::sync<confirmation>(this, &cluster::guild_add_member, gm, access_token);
 }
 
-guild_member cluster::guild_edit_member_sync(const guild_member& gm) {
```
braindigitalis, line 320: this breaks the sync methods, can't have overloaded versions as it confuses the template...

PR #610, opened 2023-02-12
Micro optimization

Neko-Life:
Not sure if anyone wanted this, just want to tidy some unused stuff and maybe save a few nanoseconds in the loop since I'm getting tired of hearing tiny gap in the audio Musicat streams and in the hope of fixing it by doing this.

FYI it doesn't fix it. I will appreciate any suggestion other than use better network since I live in a small village with no good network anywhere. I may gonna keep looking for any code that's responsible for sending voice packet and see if it can be optimized and optimized further. Maybe in the ssl read loop or tuning the sleep_time in the discord_voice_client::write_ready() method will be next.

These change is live in Musicat and no bug have been found.

### DIFF
--- a/src/dpp/discordvoiceclient.cpp
+++ b/src/dpp/discordvoiceclient.cpp
@@ -785,16 +785,22 @@ void discord_voice_client::write_ready()
 		std::lock_guard<std::mutex> lock(this->stream_mutex);
 		if (!this->paused && outbuf.size()) {
 			type = send_audio_type;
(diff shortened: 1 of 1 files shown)


braindigitalis:
how do you know this is faster, was it profiled?

I'm gonna be the one to say it, don't run a music bot in your home on domestic internet, get a VPS!

braindigitalis:
going to close this as you said it didnt improve anything...

Issue #611, opened 2023-02-13
Fix sodium_LIBRARY not being assigned when null


rube200:
I am working with this lib for a bot, currently im using two cmake env, one is in my raspberry witch take a long time to compile for the first time and the second is a WSL(Debian), the problem is after setting everything up `cmake` wouldn't work throwing me this message ```CMake Error: The following variables are used in this project, but they are set to NOTFOUND.
Please set them or make sure they are set and tested correctly in the CMake files:
sodium_LIBRARY_RELEASE (ADVANCED)```

Original log:
![image](https://user-images.githubusercontent.com/20555316/218363542-abe4e4eb-6cfc-4d84-806a-754de9712f0b.png)

Log after some debug and mod code:
![image](https://user-images.githubusercontent.com/20555316/218363682-492b5e1f-b979-47b4-98c7-b1f790bed5dc.png)

Debug code:
![image](https://user-images.githubusercontent.com/20555316/218363758-5413fcae-aaa8-4679-80e4-0398968aaaf8.png)

braindigitalis:
you need to read the output more closely, pkgconfig is missing, required by sodium

braindigitalis:
yeah this fix is wrong. it says could not find sodium then enables voice support anyway, this would fail to compile or fail to link to a program

rube200:
So i installed php since it was missing and also pkg-config and all got okay but
when i uninstall pkg-config again now this time gave me a diff output:
![image](https://user-images.githubusercontent.com/20555316/218414703-9b7558fa-b636-48a6-94c5-5ac5746c2473.png)
shouldn't it not enable voice if sodium depends of pkg-config??

Also sodium is installed but it didnt automatically install pkg-config:
![image](https://user-images.githubusercontent.com/20555316/218415484-e857beb0-fe97-4845-993e-a8e08e36f031.png)

rube200:
also after uninstalling php and pkg-config back to "original setup of packages"
![image](https://user-images.githubusercontent.com/20555316/218416684-ed27745d-93b5-4f16-9ec5-ecb99c8f4ea8.png)

NexAdn:
> So i installed php since it was missing and also pkg-config and all got okay but when i uninstall pkg-config again now this time gave me a diff output:

Looks to me like CMake just cached the details how to find sodium, so it's not surprising that it still considers it found after pkg-config is uninstalled. That's probably just CMake caching magic, not an issue with the CMakeLists.txt

[...]

PR #612, opened 2023-02-13
Fix sodium detection on wsl

rube200:
Check #611 for details.


rube200:
should i change this pr to dev?

braindigitalis:
yes, but I'm not sure if this pr belongs here or in sodium. the findsodium.cmake is from their repository originally. it's failing because you didn't install pkgconfig.

braindigitalis:
this fix will set sodium to found when  it isn't, reading the output in the linked issue. this would cause problems of failing to link or failing to compile.

PR #616, opened 2023-02-19
Added operator == to dpp::guild_member

AA1999:
Some std lib functions need a comparison operator so it's better if it's defined inside the class itself.

### DIFF
--- a/include/dpp/user.h
+++ b/include/dpp/user.h
@@ -312,6 +312,14 @@ public:
 	 * @return Formatted username and discriminator
 	 */
 	std::string format_username() const;
+	
+	/**
+	 * @brief Compare two user objects.
+	 * @param other_user Other user to compare to.
+	 * @return true if the user ids are equal, false otherwise.
+	 */
+	
+	bool operator == (dpp::user const& other_user) const;
 };
 
 /**
--- a/include/dpp/user.h
+++ b/include/dpp/user.h
@@ -319,7 +319,7 @@ public:
 	 * @return true if the user ids are equal, false otherwise.
 	 */
 	
-	bool operator == (dpp::user const& other_user) const;
+	bool operator == (user const& other_user) const;
 };
 
 /**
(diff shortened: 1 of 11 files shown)


src/dpp/guild.cpp
```diff
@@ -122,6 +122,10 @@ guild_member& guild_member::set_communication_disabled_until(const time_t disabl
 	this->communication_disabled_until = disabled_timestamp;
 	return *this;
 }
+	
+bool guild_member::operator == (dpp::guild_member const& other_member) const {
```
braindigitalis, line 126: this is missing a comparison on guild id, so the same person in two guilds will be equal and shouldnt be.

AA1999: fair enough, I'll implement right away

include/dpp/user.h
```diff
@@ -312,6 +312,14 @@ class DPP_EXPORT user : public managed, public json_interface<user>  {
...
 	 */
 	std::string format_username() const;
+	
+	/**
+	 * @brief Compare two user objects.
+	 * @param other_user Other user to compare to.
+	 * @return true if the user ids are equal, false otherwise.
+	 */
+	
+	bool operator == (user const& other_user) const;
```
braindigitalis, line 322: this is not needed. user derives from `managed` which already has `operator ==` and `operator !=` (see `dpp/managed.h`).

AA1999: Removed

src/dpp/guild.cpp
```diff
@@ -122,6 +122,10 @@ guild_member& guild_member::set_communication_disabled_until(const time_t disabl
 	this->communication_disabled_until = disabled_timestamp;
 	return *this;
 }
+	
+bool guild_member::operator == (guild_member const& other_member) const {
+	return this->user_id == other_member.user_id && this->guild_id == other_member.guild_id;
```
Commandserver, line 127: It should return false if the ids are 0. Because if someone is creating two guild_member objects, it returns true when compared because they have the default value 0.

AA1999: Alright, will edit accordingly.

PR #620, opened 2023-02-25
Updated m_suppress_notifications flag

CMDisme:
The m_suppress_notifications should be 4096, but it was 512.


braindigitalis:
thanks I already fixed this when I saw your message, I appreciate the pr!

PR #626, opened 2023-03-02
is_bot() shortcut for dpp::guild_member

AA1999:


src/dpp/guild.cpp
```diff
@@ -242,6 +243,10 @@ bool guild_member::is_muted() const {
 	return flags & dpp::gm_mute;
 }
 
+bool guild_member::is_bot() const {
+	return get_user()->is_bot();
```
Commandserver, line 247: This can cause an error as get_user csn return nullptr

AA1999: Shouldn't tho, if you have a member in your cache, you have the user in the cache too

braindigitalis: "if" you have the member in your cache.... what "if" you made a guild_member object manually? or "if" your cache is being emptied and you kept a copy of the guild member around? so many if's, never any reason to not chek for nulls.

braindigitalis: i dont see much point in adding this. if we're adding this, what about all the other is_ methods that are part of dpp::user (there are dozens). We definitely dont want a copy and paste job of all of them going into guild member, just call get_user() on the member, check for null and then call whatever method you want from there.

Just having is_bot() and no others seems weird and pointless.

AA1999: Checking for member being a bot is much more useful than checking if they're a discord staff/bug hunter etc.
But I can add the check

Commandserver: But that means the method always returns false when the member is not cached, even if the user actually is a bot. So in the end it's an unreliable result and people don't see where whether a false is a "real" false or the member was just not cached

AA1999: Well then we might have to change `get_user()` to use the `call from API if not in cache` method brain made

braindigitalis: i dont get why we need a shortcut to guild member having is_bot? there are practically no situations where you have a guild member but no user. For example in a message handler, event.msg.author.is_bot()...

AA1999: If you want to check a member in a guild is a bot or not.
For example for counting members that aren't bots.

braindigitalis: doesnt sound like a very common use case though, whats wrong with member->get_user()->is_bot()?

We can't use the 'call api if not in cache' here or the is_bot method you add would either be a blocking api call (ugh!!!) or would need a callback to return its data to you.

AA1999: Imagine you want to DM a member reacting to something, but you don't want it to be a bot
And while `member->get_user()->is_bot()` isn't bad at all, I feel like checking a guild member being bot directly would make sense.
However you are right about the issue for callback/blocking issue

AA1999: Also, you can get the nullptr issue with `get_user()` anyways
Even in that case you'll have no info on the user

braindigitalis: > Imagine you want to DM a member reacting to something, but you don't want it to be a bot And while `member->get_user()->is_bot()` isn't bad at all, I feel like checking a guild member being bot directly would make sense. However you are right about the issue for callback/blocking issue

... there is NO need to be doing anything with the member there for that. When you have a reaction event e.g. `on_message_reaction_add`, you have `event.reacting_user`, and this is filled from the event (it doesnt rely on cache) so you can always depend on it to be non-null, it isnt even a pointer.

PR #630, opened 2023-03-08
added many new colors and created alias for colors called colours

ruslan-ilesik:
All the colors was taken from https://www.computerhope.com/htmcolor.htm
Most of them are not in "html standard" but supported by all browsers so I think it could be helpful for some people who are familiar with web dev.

Also added alias for colors as both colors and colours are correct in English. Same as all gray colors have their version with grey  (also names was adjusted to style of colors written before, nothing has been removed from previous set of colors)

Unfraternally, spell checks do not know all the color names and gives a mistakes while there is not, idk how to fix it (add this words to it`s dictionary). Sorry for that.


Commandserver:
Add the words to `.cspell.json` ;)

PR #631, opened 2023-03-08
Convert http_request_completion_t headers to multimap

EonDigital:
Headers are allowed to have duplicate keys for a given request. A notable example is the set-cookie key.  Without support for this, the map will collapse these to the last entry provided.

This is an interface breaking change.

Instead of `headers[key] = value`
use `headers.emplace(key, value)`

Instead of accessing `headers[key]` or `headers.at(key)`, use:
```
auto range = headers.equal_range(key);
if ( range.first != headers.end() )
    for ( auto i = range.first; i != range.second; ++i ) {
        list.emplace_back(i->second);
    }
} else {
    // Empty
}
```


EonDigital:
Marking this as draft so I can make the change visible, but have time to add the unit testing requested.

braindigitalis (approved):

PR #633, opened 2023-03-14
fixed cluster::threads_get_active

Commandserver:
i also changed a bit of the documentation here and there e.g. documented some exceptions.

`cluster::threads_get_active` tested with:
```c++
bot.threads_get_active(ID, [](const dpp::confirmation_callback_t &event) {
			auto r = std::get<dpp::active_threads>(event.value);
			std::cout << "threads get active json: " << event.http_info.body << "\n" << r.thread_members.size() << ", " << r.threads.size() << std::endl;
		});
```

`cluster::thread_members_get` tested with:
```c++
bot.thread_members_get(1085266647419469894, [](const dpp::confirmation_callback_t &event) {
			std::cout << "thread_members_get json: " << event.http_info.body << std::endl;
		});
```


include/dpp/channel.h
```diff
@@ -809,5 +811,13 @@ typedef std::unordered_map<snowflake, channel> channel_map;
...
 typedef std::unordered_map<snowflake, thread> thread_map;
 
+/**
+ * @brief A group of threads and thread_members. returned from the cluster::threads_get_active method
+ */
+typedef struct {
+	thread_map threads;
+	thread_member_map thread_members;
+} active_thread_map;
+
```
Commandserver, line 821: i know this is c style but i wanted to stick with all the other maps. @Eremiell

PR #635, opened 2023-03-16
Added ease of access features

Crafter-san:
Added ease of access features for command creation and initialization. I'm not sure if I had to edit the cmake files but I assume I don't have to since I was just adding an hpp file


Axyte:
Feel free to request a link and mention of this as a separate wrapper on our discord #resources channel.

PR #637, opened 2023-03-24
Fix DPP install copying entire source tree root `include/` in its install directory when added as a subdirectory

Mishura4:
Currently, using DPP as a subdirectory through CMake's `add_subdirectory` command has DPP's install script resolve from the base of the entire source tree, resulting in grabbing the parent project's `include/` folder, as well as other subdirectories' `include/` folder, potentially causing errors.

This pull request changes references in the install script to the DPP directory from `${CMAKE_SOURCE_DIR}`, aka the top level of the entire source tree, to `${CMAKE_CURRENT_SOURCE_DIR}/..`, simply the parent directory of the script, resolving this issue.


Axyte:
Hi, thanks for contributing! Could you also replace other instances of `CMAKE_SOURCE_DIR`? I doubt there's any need for it to reference it from the top-level, though it will depend on the context still.

Thanks again

Mishura4:
Will try to over the weekend. Installation still works for me on Windows both using Ninja and MSVC.

Mishura4:
Here it is. I am however very unfamiliar with the process for ARM, nor do I have the ability to test it ; please double check that the changes on `cmake/ARM*ToolChain.cmake` didn't break anything, thank you!

Axyte:
Thanks!

Issue #604, opened 2023-01-25
Can't remove a member's nickname and roles with guild_edit_member


Commandserver:
**Git commit reference**
`d5f49c3fe22d1d14cadc1500f67fa9bdcffebca0`

**Describe the bug**
When using `cluster::guild_edit_member` to remove the nickname or roles of a member, it doesn't remove them. No error thrown. 

**To Reproduce**
e.g. this code:
```c++
auto m = dpp::find_guild_member(825407338755653642, 189759562910400512);
m.nickname = "";
m.roles = {};
bot.guild_edit_member(m);
```

**Expected behavior**
To remove the roles and nickname

**System Details:**
 - OS: Ubuntu
 - Discord Client used for testing [desktop]

**the fix**
The way the build_json method is made, the nickname and roles fields won't be send to discord when they're empty.

flankytingz:
Changing
```c++
if (!this->nickname.empty())
  j["nick"] = this->nickname;
if (!this->roles.empty()) {
  j["roles"] = {};
  for (auto & role : roles) {
    j["roles"].push_back(std::to_string(role));
  }
}
```
to
```c++
j["nick"] = this->nickname;
j["roles"] = {};
for (auto & role : roles)
  j["roles"].push_back(std::to_string(role));
```
in `guild.cpp` should fix the issue.

PR #640, opened 2023-03-26
Fixing issue #604

flankytingz:
Removed if statements to always update guild member's roles and nick when converting to JSON.

### DIFF
--- a/src/dpp/guild.cpp
+++ b/src/dpp/guild.cpp
@@ -209,10 +209,19 @@ std::string guild_member::build_json(bool with_id) const {
 		}
 	}
 
-    j["nick"] = this->nickname;
-    j["roles"] = {};
-    for (auto & role : roles)
-        j["roles"].push_back(std::to_string(role));
+    if (this->nickname.empty()) {
+        j["nick"] = nullptr;
+    } else {
+        j["nick"] = this->nickname;
+    }
+
+    if (this->roles.empty()) {
+        j["roles"] = nullptr;
+    } else {
+        j["roles"] = {};
+        for (auto & role : roles)
+            j["roles"].push_back(std::to_string(role));
+    }
 
 	if (flags & gm_voice_action) {
 		j["mute"] = is_muted();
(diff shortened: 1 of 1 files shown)


braindigitalis:
you now can't edit a user without setting a nickname if you do this?

similarly, you must always specificy the roles or it's going try to clear them...

not sure this fix will solve the problem without introducing more problems...?

flankytingz:
> you now can't edit a user without setting a nickname if you do this?
> 
> similarly, you must always specificy the roles or it's going try to clear them...
> 
> not sure this fix will solve the problem without introducing more problems...?

All parameters to this endpoint are optional and nullable. When moving members to channels, the API user must have permissions to both connect to the channel and have the MOVE_MEMBERS permission. 
https://discord.com/developers/docs/resources/guild#modify-guild-member

I would think that means if set to null, the roles and nickname will be removed.

braindigitalis:
yes, but you have no way to set them to null right now?

flankytingz:
> yes, but you have no way to set them to null right now?

Just realized that. They should be set to null, not "".

flankytingz:
This is assign nick and roles as null if they are empty.

```c++
if (this->nickname.empty()) {
    j["nick"] = nullptr;
} else {
    j["nick"] = this->nickname;
}

if (this->roles.empty()) {
    j["roles"] = nullptr;
} else {
    j["roles"] = {};
    for (auto & role : roles)
        j["roles"].push_back(std::to_string(role));
}
```

Axyte:
This stuff isn't going to work, user needs to decide whether the fields are to be cleared or they simply aren't present. This requires breaking API with either Booleans or introducing `std::optional`.

Commandserver:
Also, you should use `json::value_t::null`

Commandserver:
> This stuff isn't going to work, user needs to decide whether the fields are to be cleared or they simply aren't present. This requires breaking API with either Booleans or introducing `std::optional`.

Already tried the std::optional thing in #605
... looked too ugly. Maybe the booleans are a better idea

Axyte:
Uh, that isn't how I'd typically imagine `std::optional` here.

It would be the member of class so
```cpp
class guild_member {
    std::optional<std::string> nickname;
    /* ..... */
};
```
Although it would be much better, it's simply a breaking change in the API.

PR #650, opened 2023-04-08
refactor: modernize dpp::permission

Mishura4:
This PR modernizes dpp::permission by making it constexpr-compatible, making use of C++17 features and common practices.

Notable changes :
- Made value data member public, changed class to struct
- Made member functions constexpr and noexcept (except for conversion to json)
- Added unit tests for it

This effectively allows for a user to declare constexpr variables with the type `dpp::permission`, allowing to make use of the class's convenience member functions. As the type is effectively a convenience wrapper around the enum `dpp::permissions`, it allows to be used as an extension of it in a constexpr context, this is useful for example if a user wants to make a constexpr command table with default permissions, it allows them to use the dpp::permission convenience methods instead of raw enum values, while still retaining the performance benefits of constexpr variables, as the compiler is able to inline the calls very easily in a constant context. The data member was made public to make the struct a literal type to be able to be used in a constexpr context. It was directly modifiable through the implicit reference conversion anyways, so there is no reason to keep it private.

The PR also applies some modern C++14 and C++17 practices, such as applying noexcept, using `std::enable_if_t` instead of `std::enable_if`, `std::is_convertible_v` instead of `std::is_convertible`, which slightly speeds up compilation time.

### DIFF
--- a/docpages/example_programs/interactions_and_components/commandhandler.md
+++ b/docpages/example_programs/interactions_and_components/commandhandler.md
@@ -18,7 +18,7 @@ int main()
 {
 	dpp::cluster bot("token");
 
-        bot.on_log(dpp::utility::cout_logger());
+	bot.on_log(dpp::utility::cout_logger());
 
 	/* Create command handler, and specify prefixes */
 	dpp::commandhandler command_handler(&bot);
@@ -48,7 +48,7 @@ int main()
 			/* Command description */
 			"A test ping command",
 
-			/* Guild id (omit for a global command) */
+			/* Guild id (omit for a guild command) */
 			819556414099554344
 		);
 
--- a/include/dpp/appcommand.h
+++ b/include/dpp/appcommand.h
@@ -685,15 +685,16 @@ public:
 	std::variant<command_interaction, component_interaction, autocomplete_interaction> data; //!< Optional: the command data payload
 	snowflake guild_id;                                         //!< Optional: the guild it was sent from
 	snowflake channel_id;                                       //!< Optional: the channel it was sent from
+	dpp::channel channel;										//!< Optional: The partial channel object where it was sent from
 	snowflake message_id;					    //!< Originating message id for context menu actions
 	permission app_permissions;				    //!< Permissions of the bot in the channel/guild where this command was issued
 	message msg;						    //!< Originating message for context menu actions
-	guild_member member;                                        //!< Optional: guild member data for the invoking user, including permissions
+	guild_member member;                                        //!< Optional: guild member data for the invoking user, including permissions. Filled when the interaction is invoked in a guild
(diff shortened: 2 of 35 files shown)


Mishura4:
Didn't mean to put this on `master` so I'll reopen it on `dev`.

docpages/example_programs/interactions_and_components/commandhandler.md
```diff
@@ -48,7 +48,7 @@ int main()
 			/* Command description */
 			"A test ping command",
 
-			/* Guild id (omit for a global command) */
```
es183923, line 51: this change is incorrect-guild id needs to be there for guild commands, but not for global commands

include/dpp/appcommand.h
```diff
@@ -181,12 +181,12 @@ struct DPP_EXPORT command_option : public json_interface<command_option>  {
 	/**
 	 * @brief Add a localisation for this slash command option
 	 * @see https://discord.com/developers/docs/reference#locales
-	 * @param language Name of language, see the list of locales linked to above.
+	 * @param language Name of language, see the list of locales linked to above
 	 * @param _name name of slash command option in the specified language
-	 * @param _description description of slash command option in the specified language
+	 * @param _description description of slash command option in the specified language (optional)
```
es183923, line 186: why should this be optional?

include/dpp/appcommand.h
```diff
@@ -528,8 +528,8 @@ struct DPP_EXPORT command_data_option {
 	/**
 	 * @brief Get an option value by index
 	 * 
-	 * @tparam Type to get from the parameter
-	 * @param index index number of parameter
+	 * @tparam T Type to get from the parameter
+	 * @param index index of the option
```
es183923, line 532: what's this for? parameter should be kept as parameter

include/dpp/appcommand.h
```diff
@@ -582,8 +582,8 @@ struct DPP_EXPORT command_interaction {
...
```
es183923, line 586: Same as above

include/dpp/permissions.h
```diff
@@ -80,38 +82,41 @@ enum permissions : uint64_t {
 using role_permissions = permissions;
 
 /**
- * @brief Represents a permission bitmask (refer to enum dpp::permissions) which are hold in an uint64_t
+ * @brief Represents a permission bitmask (refer to enum dpp::permissions) which are held in an uint64_t
  */
-class DPP_EXPORT permission {
-protected:
```
es183923, line 86: The original inetntion of having this as protected is to prevent users from accessing the .value field, and only accessing it through its functions or implicit type conversion.

es183923:
oops, didn't see this was closed

PR #651, opened 2023-04-08
refactor: modernize dpp::permission

Mishura4:
This PR modernizes dpp::permission by making it constexpr-compatible, making use of C++17 features and common practices.

Notable changes :
- Made value data member public, changed class to struct
- Made member functions constexpr and noexcept (except for conversion to json which may throw)
- Added unit tests for it

This effectively allows for a user to declare constexpr variables with the type `dpp::permission`, allowing to make use of the class's convenience member functions. As the type is effectively a convenience wrapper around the enum `dpp::permissions`, it allows to be used as an extension of it in a constexpr context, this is useful for example if a user wants to make a constexpr command table with default permissions, it allows them to use the dpp::permission convenience methods instead of raw enum values, while still retaining the performance benefits of constexpr variables, as the compiler is able to inline the calls very easily in a constant context. The data member was made public to make the struct a literal type to be able to be used in a constexpr context. It was directly modifiable through the implicit reference conversion anyways, so there is no reason to keep it private.

The PR also applies some modern C++14 and C++17 practices, such as applying noexcept, using `std::enable_if_t` instead of `std::enable_if`, `std::is_convertible_v` instead of `std::is_convertible`, which slightly speeds up compilation time.

If this PR is accepted I will start a series of PRs updating other parts of the library to make use of C++17 features in a similar way.

### DIFF
--- a/include/dpp/permissions.h
+++ b/include/dpp/permissions.h
@@ -82,38 +82,41 @@ enum permissions : uint64_t {
 using role_permissions = permissions;
 
 /**
- * @brief Represents a permission bitmask (refer to enum dpp::permissions) which are hold in an uint64_t
+ * @brief Represents a permission bitmask (refer to enum dpp::permissions) which are held in an uint64_t
  */
-class DPP_EXPORT permission {
-protected:
+struct DPP_EXPORT permission {
+
 	/**
 	 * @brief The permission bitmask value
 	 */
-	uint64_t value;
+	uint64_t value{0};
 
-public:
 	/**
-	 * @brief Construct a permission object
-	 * @param value A permission bitmask
+	 * @brief Default constructor, initializes permission to 0
 	 */
-	permission(const uint64_t& value);
+	constexpr permission() = default;
 
 	/**
- 	 * @brief Construct a permission object
- 	 */
-	permission();
+	 * @brief Bitmask constructor, initializes permission to the argument
+	 * @param value The bitmask to initialize the permission to
+	 */
+	constexpr permission(uint64_t value) noexcept : value{value} {}
 
 	/**
 	 * @brief For acting like an integer
 	 * @return The permission bitmask value
 	 */
-	operator uint64_t() const;
+	constexpr operator uint64_t() const noexcept {
+		return value;
+	}
 
 	/**
 	 * @brief For acting like an integer
 	 * @return A reference to the permission bitmask value
 	 */
-	operator uint64_t &();
+	constexpr operator uint64_t &() noexcept {
+		return value;
+	}
 
 	/**
 	 * @brief For building json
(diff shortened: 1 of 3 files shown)


braindigitalis:
the only thing I would change here is the changing to a public data member for the value and making it a struct.

the value is private for a reason, to prevent users from directly trying to implicitly convert this type to and from int, this becomes an error instead of warning about possible loss of data. users would often accidentally convert permissions and snowflakes to int, or signed int64_t, which causes issues, this class and snowflake are to protect against that whilst providing an interface to behave somewhat like an int via the operator uint64_t().

Mishura4:
I'm confused. The member being public has no bearing whatsoever on the conversions, same for the user-defined implicit conversion, they just warn about the implicit conversion with `-Wconversion`, the same happens with using the value member. If you want an error to be generated you need to make the conversion explicit, or use `-Werror=conversion`. This has nothing to do with the member being public or not, especially as the user-defined condition is implicit.

Commandserver:
Have you tested the old unit test if they work with your changes? Because you put them in one commit

Commandserver:
> Yes. The old unit tests are still there, just moved into the lambda to be called in a constexpr context.

I would leave the old ones and just add the one for constexpr context below. Having a test for non constexpr context can't hurt

Mishura4:
I mean, it would have the same output either way 😅

Commandserver:
Also why do you use static_assert? You can use the test function

Mishura4:
So that the test can be checked at compile time, and refuse to compile if the test doesn't pass.

braindigitalis:
if the unit tests fail to compile this would stop all other tests running though so we can't measure coverage, I assume a failure to compile should never happen?

braindigitalis:
> I'm confused. The member being public has no bearing whatsoever on the conversions, same for the user-defined implicit conversion, they just warn about the implicit conversion with `-Wconversion`, the same happens with using the value member. If you want an error to be generated you need to make the conversion explicit, or use `-Werror=conversion`. This has nothing to do with the member being public or not, especially as the user-defined condition is implicit. Using the value member is exactly the same as using the implicit conversion, currently.
> 
> Example : https://godbolt.org/z/Tfbbrs9s6

if the member is public the user can bypass the protection provided by the class and access .value directly defeating the point. what would be the advantage of exposing the guts like this?

Mishura4:
Re unit tests: I mean, the purpose of unit tests is to maintain functionality through changes, is there ever a situation in which we want a build to pass while unit tests fail?

Re public member: There is virtually no protection in place due to the conversion operator to a reference anyways, the current design simply makes the type opaque, which is to be questioned what the benefits are, plus it does have drawbacks (e.g. the type cannot be used as a NTTP).

Mishura4:
Actually after I asked in a C++ community, I was given good arguments for why *not* to static_assert unit tests (e.g. it prevents debugging), so I will change that later today.

braindigitalis:
still, can you please make it a class again and make the internal type private? this was the intent of the design, it works for us and we don't want to change that. everything else is fine.

Mishura4:
Emoji of a frog doing a thumbs up expression

PR #653, opened 2023-04-09
fix: move cl compilation flags that clang-cl doesn't recognize to a "not clang-cl" clause

Mishura4:
When using clang-cl on windows which emulates MSVC and thus passes the `if (MSVC)` on CMake, some flags that cl uses aren't recognized. This produces warnings:
`warning : argument unused during compilation: '/MP' [-Wunused-command-line-argument]`
`warning : argument unused during compilation: '/Zc:preprocessor' [-Wunused-command-line-argument]`
`warning : argument unused during compilation: '/GL' [-Wunused-command-line-argument]` (in Release)

This PR makes it so that these flags are only set if an MSVC environment is used but the compiler is *not* clang-cl.


braindigitalis:
not sure we want to have to support clang-cl at present? it means yet another thing to put into ci, and something we dont all have access to (most of our contributors are not windows devs)

If we merge this, what other things are we committing to supporting? Just curious...

Mishura4:
You don't have to officially support a whole compiler and commit on maintaining it to accept contributions that do make it possible or improve their usage. 😄

PR #661, opened 2023-04-30
Updated information regarding WSL

wizard7377:
Removed opening vscode in separate window then navigating to WSL in favor of `code .` on the command line. Also, added a note on the Visual Studio 20xx page on the usefulness of WSL.


wizard7377:
Note, updated as to fix cstdint no longer being automatically included as of 13.1

PR #663, opened 2023-04-30
Updating to add the new unordered_set implementation.

RealTimeChris:
This should relatively significantly reduce memory usage as there is no additional storage being used for storing the keys to each of the items in the set.


Warpten:
May I suggest using `std::unordered_set` with heterogenous key lookup (`is_transparent` typedef on hasher and equality tester) if DPP is built for C++20?

Also, why is the actual set behind an unique_ptr?

braindigitalis:
this is apparently faster than an unordered map.

would love to see some profiling in real world use - not just filling a container with n items and comparing to map.

my concern is only that this hasn't been tested in production like unordered map is and we risk introducing some bugs in subtle places.

braindigitalis:
im closing this pr for now, i think its safer and more compatible and safer to stick with standard containers, especially as users can directly access the containers.

PR #665, opened 2023-05-08
added an include for cstdint. Fixes build issues for arch systems using clang + ninja. (uint16_t type not found)

Erarnitox:
as the title explains I had build issues on an arch system where the uint16_t type could not be found. On my Ubuntu it looks like this header is implicitly included. I made this pull request to explicitly include this header to avoid potential build errors.


Erarnitox:
This has already been fixed in the dev branch (didn't see it)

PR #676, opened 2023-05-10
added explicit casting to avoid warnings

nagilo:
Using a MacOS 13.2 with Xcode 14.2 I received the following warnings. 

Implicit conversion loses integer precision: 'unsigned long' to 'uint32_t' (aka 'unsigned int')
Implicit conversion loses integer precision: 'size_t' (aka 'unsigned long') to 'int'

Added casts solved the warnings and is still consistent with the intention of the code.

### DIFF
--- a/src/dpp/discordvoiceclient.cpp
+++ b/src/dpp/discordvoiceclient.cpp
@@ -398,12 +398,12 @@ int discord_voice_client::udp_send(const char* data, size_t length)
 	servaddr.sin_family = AF_INET;
 	servaddr.sin_port = htons(this->port);
 	servaddr.sin_addr.s_addr = inet_addr(this->ip.c_str());
-	return sendto(this->fd, data, (int)length, 0, (const sockaddr*)&servaddr, (int)sizeof(sockaddr_in));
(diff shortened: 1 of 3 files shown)


src/dpp/discordvoiceclient.cpp
```diff
@@ -398,12 +398,12 @@ int discord_voice_client::udp_send(const char* data, size_t length)
...
 	servaddr.sin_port = htons(this->port);
 	servaddr.sin_addr.s_addr = inet_addr(this->ip.c_str());
-	return sendto(this->fd, data, (int)length, 0, (const sockaddr*)&servaddr, (int)sizeof(sockaddr_in));
+	return (int) sendto(this->fd, data, (int)length, 0, (const sockaddr*)&servaddr, (int)sizeof(sockaddr_in));
 }
 
 int discord_voice_client::udp_recv(char* data, size_t max_length)
 {
-	return recv(this->fd, data, (int)max_length, 0);
+	return (int) recv(this->fd, data, (int)max_length, 0);
```
braindigitalis, line 406: why do you need this?
The correct return type for `recv` is int already according to the standards? same with `sendto`

PR #687, opened 2023-05-14
fix: fixed memory leak and potential use-after-free bug with voiceconn

Mishura4:
dpp::discord_voice_client does not free its active voice connections on destruction. This means that, when the cluster/voice client gets destroyed, any dpp::voiceconn objects as well as threads they have running won't be joined or freed. This causes not only a memory leak, but in the event of the cluster/voice client getting destroyed, and the still alive voiceconn thread trying to for example log a message, a use-after-free will occur and cause undefined behavior (usually a crash). This has happened to me about on about 50% of my executions of the unit tests on Windows.

This PR fixes this bug. However it does make use of std::unique_ptr, that as I understand it the library does not usually use, let me know if you would like me to change it to new/delete instead.

### DIFF
--- a/include/dpp/discordclient.h
+++ b/include/dpp/discordclient.h
@@ -360,7 +360,7 @@ public:
 	/**
 	 * @brief List of voice channels we are connecting to keyed by guild id
 	 */
-	std::unordered_map<snowflake, voiceconn*> connecting_voice_channels;
+	std::unordered_map<snowflake, std::unique_ptr<voiceconn>> connecting_voice_channels;
 
 	/**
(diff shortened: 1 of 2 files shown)


braindigitalis:
is the pointer now freed by removal of the iterator from the map?

Mishura4:
Pretty much, or when the map gets destroyed. The map owns the pointer, and when its elements are destroyed, the voiceconn is also destroyed and freed.

I didn't find any other code that made use of that map or grabbed/changed the pointer, so this change should be fine.

PR #700, opened 2023-05-21
fix, refactor, test, feat: added cluster::thread_edit, fix cluster::thread_members_get, refactor cluster::threads_get_active, add a bunch of tests for threads

Mishura4:
⚠️ WAITING TO MERGE #697 as this PR depends on it (I couldn't do anything otherwise as I am on Windows kekw)

### added cluster::thread_edit
Added `cluster::thread_edit`, as the only way to edit a thread was to use `cluster::channel_edit`, however it only takes a `dpp::channel` as a parameter, losing the extra fields from `dpp::thread`.

### fix cluster::thread_members_get
This function was broken, as the key from the `thread_member_map` it returns is the thread id, meaning, filling the list of users in the thread would just overwrite the same map element over and over. f9f01c161326352898ffaf5d779cc4e07fd1b37b changes it so the key is now the user_id.

### refactor cluster::threads_get_active
Following the changes in f9f01c161326352898ffaf5d779cc4e07fd1b37b, cluster::threads_get_active was refactored as it also depends on `thread_member_map` in its return type `active_threads`, in 47369c9c4801174f7755b6e4a01b70ebce56eeb7. Now the structure is a map of `active_thread`, the latter being a pair-ish of a thread and a `std::optional<dpp::thread_member>` (has_value() == the bot has joined the thread)
Sync tested, coro tested.

### add a bunch of tests for threads
Tests for creating a thread from a message, creating a thread on its own, editing a thread, deleting a thread, events for message creation, reaction, deletion, edition, in a thread.


src/unittest/test.cpp
```diff
@@ -806,62 +821,403 @@ Markdown lol \\|\\|spoiler\\|\\| \\~\\~strikethrough\\~\\~ \\`small \\*code\\* b
...
+			bool edit_tested = false;
+			bool members_tested = false;
+			bool messages_tested = false;
+			bool events_tested = false;
+			uint32_t events_tested_mask = 0;
+			uint32_t events_to_test_mask = 0;
+
+			void delete_if_done()
+			{
+				if (edit_tested && members_tested && messages_tested && events_tested) {
```
Mishura4, line 867: re: static analyser "Condition 'edit_tested' is always false"
it's not 😆

src/unittest/test.h
```diff
@@ -277,3 +277,45 @@ class message_collector : public dpp::message_collector {
...
+ */
+inline constexpr user_project_id_t get_user_snowflake;
+
+/**
+ * @brief Convenience lambda to check if a certain user is the owner of the test bot, mostly meant to be passed to standard algorithms
+ * @see get_user_snowflake
+ *
+ * @return bool whether the user is the test bot owner
+ */
+inline constexpr auto is_owner = [](auto &&user) noexcept -> bool {
```
braindigitalis, line 314: please don't do this! (return type auto with actual return type on the end of the definition using ->) it may be modern but it is unreadable to me, and doesn't follow the code style of the rest of dpp.

braindigitalis: (besides it's just extra typing and forces visual scanning all the way across the line, what's wrong with `constexpr bool is_owner...`

Mishura4: This is lambda, and it needs to be a lambda to be passed for example to std::find_if

Mishura4:
I believe this is finally ready to be merged, just merged dev into it as well. I was trying to rebase and clean up the commits a bit but it was bonking at 7e2ea124 and then ab7dc94a 😅 oh well

PR #717, opened 2023-06-27
[Docs/FAQ] Add Windows VS 2022 usage instructions

Andosius:
^


braindigitalis:
thinking about this, i'm not sure if we want to add this. It would encourage people back to the roll-your-own again instead of "use the template" which we have been pushing now successfully. We would then be back to dealing with people not setting it up to use the correct debug and build versions of the lib. The less we see "scary ptr" windows issues the better.

Andosius:
I can understand that. It is up to you, I have a well tested and working alternative for that:

Maybe I can change the page to build projects with my current DPP-CMake project? The latest update made it stable on Debian based distros (tested it on x64, arm64, armhf and auto builds libdpp if it can‘t be installed via dpkg) and on Windows.
It may even work with macOS via auto build.

For compatibility reasons (older CMake versions) I dropped FetchContent and rewrote the whole thing.
It uses curl on unix systems and PowerShell streams on Windows to get the desired version.
That makes using DPP extremly easy.

The only things left are:
- Improve the README and add dependencies for all OSes
- Explain how to trigger macOS build as it is experimental (I am sure macOS ist still supported, correct?)
- Adding proper macOS and Windows installation checks (not only testing for the download dir)

That will take one more day to finish and we will have a fully compatible solution that allows everyone to use and install DPP using CMake cross platform! And that with literally 3 lines of code and almost no overhead when configuring and everything is present.

You can test it yourself:
https://github.com/Andosius/DPP-CMake/tree/master

Let me know what you think.

PR #723, opened 2023-07-09
feat: cluster::interaction_followup_get_original

Mishura4:
This PR adds a method for the [Get Original Interaction Response](https://discord.com/developers/docs/interactions/receiving-and-responding#get-original-interaction-response) API endpoint, which is quite essential in order to use raw interactions. It also adds `interaction_response_get_original` as an alias for it, similar to `interaction_response_create`. Also adds sync and coro methods for both.


braindigitalis:
doesnt this do what `get_original_response` already does?
there are so many interaction endpoints, i get confused.

Mishura4:
Yes, from dpp::interaction_create_t ; however this is similar to the `cluster::interaction_followup_edit_original()` that already exists, which is the same as `dpp::interaction_create_t::edit_original_response`.

Having a method on the cluster notably allows for sync and coro convenience calls ; for example with PR #724 one can do `co_await cluster->co_interaction_get_original(...)` instead of `co_await dpp::awaitable{event, &dpp::interaction_create_t::get_original_response}`

Axyte:
There is no such thing as "original followup" in the API.

Mishura4:
So then should we remove `cluster::interaction_followup_edit_original` which is currently already in the lib?

Commandserver:
Im for removing the old interaction_followup_edit_original

Commandserver:
It's more confusing than helpful having this duplication anyway

Axyte:
> So then should we remove `cluster::interaction_followup_edit_original` which is currently [already in the lib](https://github.com/brainboxdotcc/DPP/blob/dev/include/dpp/cluster.h#L1313)?
 
Sure.

braindigitalis:
we can't just remove it, things are using it.

it would have to be deprecated, but there is a reason it's called "original":

https://discord.com/developers/docs/interactions/receiving-and-responding#followup-messages

the API endpoint contains the string "@original". it was originally called the "original response" and was renamed to "initial response". both mean the same thing.

definitely do not delete!

braindigitalis:
also it's in the API docs section about follow up messages.

the API docs on interactions are a mess.

Axyte:
That is not a followup message.

craigpipeandpiper:
no, bots are using that function

PR #724, opened 2023-07-10
feat: coroutines, still experimental but useable

Mishura4:
It's here! The rewrite of DPP's coroutines. Now supports :
- The task object going out of scope (e.g. starting a coroutine in an event callback and returning). The promise detaches itself and will handle its own destruction at the end of the coroutine
- Nested tasks, theoretically infinite levels of nesting (until out of memory)
- dpp::awaitable is now generic and can be used to wrap *any* function whose signature ends with a callback :
  - `co_await dpp::awaitable{interaction, &dpp::interaction_create_t::get_original_response};`
  - `co_await dpp::awaitable<http_request_completion_t>{[&](auto &&callback) { cluster->request(url, method, callback, postdata, mimetype, headers); }};`
- API calls are sent immediately on creation of the awaitable so a user can start a few and wait for all of them in one go later

Example :
```c++
_bot->on_message_create.co_attach([](const dpp::message_create_t& e) -> dpp::task<void> { // Tasks attached to an event are void return only
	// Make a nested task which will return a boolean (or any other type). Not necessary, this is to show nested tasks
	// Very advised to take any variables by parameter and not by capture - this way if the lambda goes out of scope but the task keeps running, the variables are still valid due to being on the coroutine stack frame
	auto nested_task = [](dpp::cluster *cluster, dpp::message event) -> dpp::task<bool> {
		// Create a message and wait for the API response
		dpp::confirmation_callback_t confirm = co_await cluster->co_message_create(dpp::message{"Retrieving emoji list"}.set_channel_id(event.channel_id));

		// Retrieve the dpp::message returned by the API
		dpp::message original = confirm.get<dpp::message>();

		// Get the emoji list from the guild and wait for the API response
		confirm = co_await cluster->co_guild_emojis_get(event.guild_id);

		// Get the dpp::emoji_map from the API response
		const auto &map = confirm.get<dpp::emoji_map>();
		int i = 0;
		for (auto it = map.begin(); it != map.end() && i < 5; ++i) { // List the 50 first emoji found, 10 per message
			dpp::message message{fmt::format("{}", fmt::join(
				std::ranges::views::iota(it, map.end()) // C++20 flex 😎
				| std::ranges::views::transform([](auto &&elem){ return (elem->second.get_mention()); })
				| std::ranges::views::take(10),
			" "))};

			// Create the message and wait for the API response
			co_await cluster->co_message_create(message.set_channel_id(event.channel_id));
			std::ranges::advance(it, 10, map.end());
		}
		co_return true;
	};
	// Wait for the whole thing to finish - for example you may want to do something else after with the result, maybe another nested task
	bool success = co_await nested_task(e.from->creator, e.msg);
}
```

Output:
Example output of the code above

✅Thoroughly tested on MSVC 19.37.32705 and g++ 13.1, works like a charm, no leaks, no crashes during intended usage.
⚠NOT tested on clang. Currently coroutines are still labelled as "experimental" on clang, and require using libc++. If anyone wants to test it, that would be really appreciated.


Mishura4:
Need to keep tweaking the docs, especially formatting, which I will do within the next few days

Mishura4:
👍

braindigitalis (approved):

PR #727, opened 2023-07-15
feat: Add flags and `is_remix()` to `dpp::attachment`

MCausc78:
Added flags and `is_remix()` method to `dpp::attachment`, `dpp::message`

### DIFF
--- a/include/dpp/guild.h
+++ b/include/dpp/guild.h
@@ -203,8 +203,6 @@ public:
 	time_t joined_at;
 	/** Boosting since */
 	time_t premium_since;
-	/** total permissions of the member in the channel, including overwrites, returned when in the interaction object */
-	permission permissions;
 	/** A set of flags built from the bitmask defined by dpp::guild_member_flags */
 	uint16_t flags;
 
--- a/include/dpp/guild.h
+++ b/include/dpp/guild.h
@@ -203,6 +203,8 @@ public:
 	time_t joined_at;
 	/** Boosting since */
 	time_t premium_since;
+	/** total permissions of the member in the channel, including overwrites, returned when in the interaction object */
+	permission permissions;
 	/** A set of flags built from the bitmask defined by dpp::guild_member_flags */
 	uint16_t flags;
 
--- a/include/dpp/guild.h
+++ b/include/dpp/guild.h
@@ -203,7 +203,7 @@ public:
 	time_t joined_at;
 	/** Boosting since */
 	time_t premium_since;
-	/** total permissions of the member in the channel, including overwrites, returned when in the interaction object */
+	/** total permissions of the member in the channel, including overwrites, returned when in the interaction object */
 	permission permissions;
 	/** A set of flags built from the bitmask defined by dpp::guild_member_flags */
 	uint16_t flags;
--- a/include/dpp/message.h
+++ b/include/dpp/message.h
@@ -767,6 +767,14 @@ struct DPP_EXPORT reaction {
 };
 
 /**
+ * @brief Bitmask flags for a dpp::attachment
+ */
+enum attachment_flags : uint16_t {
+	/// this attachment has been edited using the remix feature on mobile
+	a_is_remix = 1 << 2,
+};
+
+/**
  * @brief Represents an attachment in a dpp::message
  */
 struct DPP_EXPORT attachment {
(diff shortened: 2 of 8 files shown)


src/dpp/message.cpp
```diff
@@ -249,6 +249,8 @@ void to_json(json& j, const attachment& a) {
 		j["url"] = a.url;
 	}
 	j["ephemeral"] = a.ephemeral;
+	/* in future, Discord can allow bots send remixed attachment */
+	j["flags"] = a.flags;
```
braindigitalis, line 253: where do you hear this from? we can't just add support on heresay as poeple will try to use it, at worst case sending flags when it is not supported will generate an error.

braindigitalis (changes requested):

src/dpp/message.cpp
MCausc78: calm down, Discord API ignores 'flags' field if bot account, using user account would result attachment with flags

braindigitalis: still, if the docs don't have it we shouldn't send it

MCausc78: Ok fine

include/dpp/message.h
```diff
@@ -1523,6 +1540,13 @@ struct DPP_EXPORT message : public managed {
 	 * @return true if message is a DM
 	 */
 	bool is_dm() const;
+
+	/**
+	 * @brief Returns true if any of attachments is remixed
+	 * 
+	 * @return true if any of attachments is remixed
+	 */
+	bool is_remix() const;
```
Commandserver, line 1549: i dont see much point in adding this. Just having is_mixed() to check for attachments and nothing more seems pointless. It's a copy and paste from attachment and better placed only in dpp::attachment. Also the name of this method implies that "the message" "is remixed". If we're adding this, at least rename it to something like `has_remix_attachment`.

Commandserver (changes requested):

include/dpp/guild.h
```diff
@@ -203,6 +203,8 @@ class DPP_EXPORT guild_member {
 	time_t joined_at;
 	/** Boosting since */
 	time_t premium_since;
+	/** total permissions of the member in the channel, including overwrites, returned when in the interaction object */
+	permission permissions;
```
Commandserver, line 207: the permissions are stored in the interaction event itself to reduce storage as they're mostly empty because they're only retrieved through interaction events iirc

include/dpp/permissions.h
```diff
@@ -146,6 +146,25 @@ class DPP_EXPORT permission {
...
 
+	/**
+	 * @brief Check for permission flags set. It uses the Bitwise AND operator
+	 * @tparam T one or more uint64_t permission bits
+	 * @param values The permissions (from dpp::permissions) to check for
+	 *
+	 * **Example:**
+	 *
+	 * ```cpp
+	 * bool is_mod = permission.has(dpp::p_administrator, dpp::p_ban_members);
```
Commandserver, line 157: is this meant to be `has_eny()`? not just `has()`?

Commandserver (changes requested):

include/dpp/guild.h
MCausc78: not writing `interaction.resolved.member_permissions.find(interaction.usr.id)->second->has(dpp::p_ban_members)`, sorry

Commandserver:
i'll merge it once the builds are complete, okay?!

MCausc78:
> i'll merge it once the builds are complete, okay?!

yes

include/dpp/guild.h
braindigitalis: this would inflate the guild records (remember there are potentially hundreds of thousands of these cached in ram) for a value that is only used in an interaction. as commandserver said get_resolved_permissions is the way to do this, the permissions are per user not per guild.

include/dpp/permissions.h
braindigitalis: permission can take a list like this, if I remember correctly

MCausc78: > permission can take a list like this, if I remember correctly

`has` is check that you have all listed permissions, but `has(dpp::p_adminstrator | dpp::p_ban_members)` does

MCausc78:
> i'll merge it once the builds are complete, okay?!

build successful

include/dpp/utility.h
```diff
@@ -253,6 +253,13 @@ namespace dpp {
 		 */
 		bool DPP_EXPORT has_voice();
 
+		/**
+		 * @brief Returns true if D++ was built with coroutine support
+		 * 
+		 * @return bool True if coroutines are supported 
+		 */
+		bool DPP_EXPORT is_coro_enabled();
```
braindigitalis, line 261: why is this in this pr?
please do not combine various different features into one pr as they will be squash merged.

braindigitalis (changes requested):

braindigitalis (approved):

PR #734, opened 2023-07-17
feat: added functionality for guild onboarding and welcome screen

Commandserver:
This PR adds functionality to get and edit [guild onboarding](https://discord.com/developers/docs/resources/guild#guild-onboarding-object) and [welcome screen](https://discord.com/developers/docs/resources/guild#welcome-screen-object). A `welcome_screen_t` class already existed tho. Renamed it to `welcome_screen` because it's no more just a type. This refactoring shouldn't be breaking, as there was never a reason for anyone to use the old (`welcome_screen_t`) class. It was just a member of the `guild` class.

Implemented the following endpoints:
- [Get Guild Onboarding](https://discord.com/developers/docs/resources/guild#get-guild-onboarding)
- [Modify Guild Onboarding](https://discord.com/developers/docs/resources/guild#modify-guild-onboarding)
- [Get Guild Welcome Screen](https://discord.com/developers/docs/resources/guild#get-guild-welcome-screen)
- [Modify Guild Welcome Screen](https://discord.com/developers/docs/resources/guild#modify-guild-welcome-screen)


src/dpp/cluster/guild.cpp
```diff
@@ -150,4 +150,13 @@ void cluster::guild_sync_integration(snowflake guild_id, snowflake integration_i
 }
 
 
+void cluster::guild_get_onboarding(dpp::onboarding o, dpp::command_completion_event_t callback) {
```
braindigitalis, line 153: guild get onboarding should take guild id not onboarding as it's first parameter imho

Commandserver: I thought it can take the guild id from the onboarding object. But yeah ill change it.

include/dpp/guild.h
```diff
@@ -382,26 +384,94 @@ class DPP_EXPORT guild_member {
...
-	/// the server description shown in the welcome screen
+struct DPP_EXPORT welcome_screen: public json_interface<welcome_screen> {
+	/// The server description shown in the welcome screen
 	std::string description;
-	/// the channels shown in the welcome screen, up to 5
-	std::vector<welcome_channel_t> welcome_channels;
+	/// The channels shown in the welcome screen (max 5)
+	std::vector<welcome_channel> welcome_channels;
+	/// Whether the welcome screen should be enabled or disabled. It's empty for received welcome-screens. Only for modifying the welcome screen
+	std::optional<bool> enabled;
```
Commandserver, line 442: Not sure if this is the right place. It's only used when editing the welcome screen: https://discord.com/developers/docs/resources/guild#modify-guild-welcome-screen. It's stored in the guild class, so it would take more ram. I'm thinking if i should make this an extra parameter of `cluster::guild_edit_welcome_screen`. wdym?

braindigitalis: i think youre right, this is an excessive amount of extra storage, for something like the welcome channels which the majority of servers simply wont have. Also does it need an enabled flag to be optional? that seems like a tri-state bool which is just ugh. its enabled, or it isnt.

Commandserver: optional because it's an optional field in the endpoint

Commandserver: should i make two different structs of welcome screens? one `welcome_screen_t` as it was before, and `welcome_screen` which is derived from `welcome_screen_t` and has this extra boolean and is only used for the two welcome-screen-endpoints?

Commandserver: hm nah, sounds too complicated i guess. ill add the bool to `guild_edit_welcome_screen`

braindigitalis (approved):

PR #740, opened 2023-07-20
fix: add /bigobj to non-VCPKG MSVC CMake flags, fixing ninja build, and fix CMakeSettings.json

Mishura4:
- Adds /bigobj to Debug MSVC CMake flags for building the library, required to use nlohmann::json with Ninja (see https://github.com/nlohmann/json/issues/1114) (in `library/CMakeLists.txt` so it shouldn't affect vcpkg)
- Fixes the "x64-Debug" profile to actually build in Debug instead of currently Release
- Adds "x64-Release", "x64-Debug-Coro", "x64-Release-Coro" CMake profiles to use with Visual Studio


braindigitalis:
we don't support vs2015 though

Mishura4:
It's an error that happens since vs2015, it still happens on recent versions with Ninja

PR #743, opened 2023-07-26
fix: coroutines: better support for clang, remove experimental if unneeded, add missing headers

Mishura4:
This PR should improve compiler support for coroutines. Mainly, it adds clang with stdc++, as well as clang17, which removed experimental headers.

The list of supported compilers becomes as follows :
- g++: experimental support for g++10 (requires `-fcoroutines` on a consumer project), native support from g++11 onwards
- clang:
  - [9, 14): experimental support with libc++
  - [12, 14): experimental support with stdc++/glibcxx
  - 14+: native support with either
- no changes for MSVC (still "17.something")

Draft, I still need to test compilers.

### DIFF
--- a/include/dpp/coro.h
+++ b/include/dpp/coro.h
@@ -1,11 +1,21 @@
 #ifdef DPP_CORO
 #pragma once
 
-#if !(defined( _MSC_VER ) || defined( _CONSOLE ) || defined( __GNUC__ )) // if libc++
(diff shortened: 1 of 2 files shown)


Mishura4:
👍 Tested g++10, g++11, g++13, clang13 with libc++, clang16 with stdc++, clang16 with libc++.

PR #744, opened 2023-07-27
refactoring: Json parsing helper functions added to unify the way arrays are parsed

Commandserver:


include/dpp/discordevents.h
```diff
@@ -39,6 +40,30 @@ uint64_t DPP_EXPORT snowflake_not_null(const nlohmann::json* j, const char *keyn
...
+ */
+void DPP_EXPORT set_snowflake_array_not_null(const nlohmann::json* j, const char *keyname, std::vector<class snowflake> &v);
+
+/** @brief Sets an array of objects from a json field value, if defined, else does nothing
+ * @tparam T The class of which the array consists of. Must be derived from dpp::json_interface
+ * @param j nlohmann::json instance to retrieve value from
+ * @param keyname key name to check for the values
+ * @param v Value to change
+ */
+template<class T> inline void DPP_EXPORT set_object_array_not_null(nlohmann::json* j, const char *keyname, std::vector<T> &v) {
```
Mishura4, line 56: Shouldn't DPP_EXPORT here since it's a template. Might get errors on Windows because of it
Also `inline` is redundant here as well also because it's a template, though that's not important.

I'll request changes for the DPP_EXPORT

Mishura4 (changes requested):

Mishura4 (approved):

Mishura4 (approved):

braindigitalis (approved):

PR #745, opened 2023-08-01
Adding AVX implementation for mixing audio.

RealTimeChris:
Should significantly increase efficiency on supported platforms. Specify a CMake variable called AVX_TYPE that is set to either T_AVX512, T_AVX2, T_AVX, or T_AVX_FALLBACK to specify each of the respective architectures.


include/dpp/isa_detection.hpp
craigpipeandpiper, line 1: minor change, headers in dpp should be .h not .hpp

include/dpp/isa_detection.hpp
```diff
@@ -0,0 +1,256 @@
...
+ *
+ ************************************************************************************/
+#pragma once
+
+#if defined _MSC_VER || defined __GNU__ || defined __clang__
+	#include <immintrin.h>
+#endif
+
+#ifdef HAVE_VOICE
+	#include <opus/opus.h>
```
craigpipeandpiper, line 28: we shouldnt include this in the public header, or it makes the bot diretly depend on opus headers. This can cause huge issues. Can this be put inside the implementation without causing problems? opus_int16 etc are basically exactly like stdint variants of the same thing.

include/dpp/isa_detection.hpp
```diff
@@ -0,0 +1,256 @@
...
+	 * @param up_sampled_vector Pointer to the array of opus_int32 values.
+	 * @param decoded_data Pointer to the array of opus_int16 values.
+	 * @param x Index to select a specific set of elements to combine.
+	 */
+	inline void combine_samples(opus_int32* up_sampled_vector, const opus_int16* decoded_data, size_t x) {
+		for (size_t y = 0; y < 4; ++y) {
+			__m128i* up_sampled_ptr = reinterpret_cast<__m128i*>(up_sampled_vector + (x * 8 * y));
+			const __m128i* decoded_ptr = reinterpret_cast<const __m128i*>(decoded_data + (x * 8 * y));
+
+			__m128i current_up_sampled = _mm_loadu_si128(up_sampled_ptr);
```
craigpipeandpiper, line 203: can/should we 'using' these __m128 etc to make them more readable?

src/dpp/discordvoiceclient.cpp
```diff
@@ -54,6 +54,30 @@
 #endif
 
 namespace dpp {
+moving_averager::moving_averager(uint64_t collectionCountNew) {
```
craigpipeandpiper, line 57: variable names should be snake_case in dpp :)

win32/include/openssl/cms.h
craigpipeandpiper, line 1: not sure why these are changed in openssl? can they please be excluded from the PR

craigpipeandpiper (commented):
left a review from my work account

include/dpp/isa_detection.hpp
Mishura4: I'm pretty sure all those reinterpret_cast are UB as well? Doesn't this violate the strict aliasing rule of C++?

craigpipeandpiper: dont know, personally i'd use a C cast here, for clarity to indicate "we arent doing anything special, and we just treat this data as raw bits"... because thats essentially what AVX is doing, packing the raw bytes of multiple values into a native 512 or 256 bit register.

Mishura4: C casts are basically the same as reinterpret_cast in this case, that falls into UB territory as well
If the data was obtained as a different type, type aliasing needs to be done via memcpy or it is UB

craigpipeandpiper: pretty sure you cant memcpy these, they are intrinsics and are in registers not ram... it works fine in discordcoreapi, so something acceptable is at play (if its ub, it seems to work everywhere just out of dumb luck)

Mishura4: Yeah this is generally one of the UBs that work on most major hardwares but will break on anything unconventional

include/dpp/discordvoiceclient.h
```diff
@@ -37,6 +37,7 @@
 #include <dpp/dispatcher.h>
 #include <dpp/cluster.h>
 #include <dpp/discordevents.h>
+#include <dpp/isa_detection.hpp>
```
braindigitalis, line 40: this still includes the hpp instead of the .h

PR #747, opened 2023-08-03
D++ == CHAD++ // true

Axyte:


include/dpp/event_router.h
```diff
@@ -173,13 +173,13 @@ template<class T> class event_router_t {
 					from->creator->log(dpp::loglevel::ll_error, std::string{"Uncaught exception in event coroutine: "} + exception.what());
 			}
 		};
-		std::for_each(coroutine_container.begin(), coroutine_container.end(), [&](auto &ev) {
+		for (auto& ev : coroutine_container) {
```
Mishura4, line 176: Did this not work with `const auto&`?

braindigitalis: it wasn't const before, you can't cancel it if it's const either

Mishura4: It should be, that's not the actual event, but the event handler
Tbh I'd change "ev" to something more explicit here

Axyte: Added and renamed.

include/dpp/integration.h
```diff
@@ -169,5 +169,5 @@ typedef std::unordered_map<snowflake, integration> integration_map;
 /** A group of connections */
 typedef std::unordered_map<snowflake, connection> connection_map;
 
-};
+} // namespace dpp
```
braindigitalis, line 172: are these really needed when every sane IDE shows what namespace the curly braces belong to?
seems to have made the pr huge.

Neko-Life: well seems like the point of the PR is just "tidy up" stuff according to their coding style

include/dpp/cluster.h
```diff
@@ -232,6 +232,16 @@ class DPP_EXPORT cluster {
...
 
+	/**
+	 * @brief dpp::cluster is non-copyable
+	 */
+        cluster& operator=(const cluster&) = delete;
+
+	/**
+	 * @brief dpp::cluster is non-moveable
+	 */
+        cluster& operator=(const cluster&&) = delete;
```
Mishura4, line 243: The parameter shouldn't be const here

Mishura4 (changes requested):

include/dpp/integration.h
Mishura4: That's a bit unfair, I'd say. Aside from the closing namespace comments, these are common modern practices really, with clear improvements and no drawbacks.

include/dpp/cluster.h
Axyte: Why?

braindigitalis: isnt it being changed? therefore never const?

Mishura4: Because that's not a move assignment operator. A move assignment operator is a special function with a specific signature that takes a non-const rvalue.

Axyte: https://en.cppreference.com/w/cpp/language/move_assignment

Mishura4: Seems fine after all, though I don't think it makes a lot of sense but since there isn't any practical issues with it i'll mark it resolved

Mishura4: Nvm yeah that is quite the pog, must I say https://godbolt.org/z/6P7fPe1Gx 👍

braindigitalis:
this ready to merge now?

needs a more useful title for the pr though for when it is squash merged

Mishura4 (approved):

braindigitalis (approved):

PR #763, opened 2023-08-08
feat: coroutines "beta test"

Mishura4:
### dpp::task<R> - coroutine object
Hot start (starts on constructor), executes in parallel
Can be cancelled which will throw dpp::task_cancelled_exception **__after__** the next co_await
Cannot detach, task will cancel itself when destroyed
Can be co-awaited later to retrieve return value
Requires synchronization at the point of co_await and destruction (lock-free)

### dpp::coroutine<R> - coroutine object
Lazy start (starts on co\_await)
Has a return value, retrieved on co_await
Cannot detach or run in parallel by design of lazy start
No synchronization, light object

### dpp::job - coroutine object
Hot start, runs in parallel and detached always, destroys itself when done
Cannot be co\_awaited
No synchronization, no return value, very light object

### dpp::async<R> - awaitable API request (previously dpp::awaitable<R>)
Can be constructed from a function
e.g:
- returned by co\_\*: `dpp:async request = cluster->co_message_create(dpp::message{"foobar"});`
- from arbitrary function: `dpp::async request{cluster, &dpp::cluster::message_create, dpp::message{"foobar"}};`

Hot start (starts on constructor), executes in parallel
Lock-free synchronization on co_await


Mishura4:
Depends on #775 as one of the examples in the docs gets a SIGSEGV from the destructor of a copied dpp::emoji

wizard7377:
Please wrap all C++20 specific features in a `#if __cplusplus != 201703L` and `#endif`

Mishura4:
DPP_CORO covers that.

docpages/advanced_reference/coroutines.md
```diff
@@ -19,7 +19,7 @@ int main() {
 
     /* Message handler to look for a command called !file */
     /* Make note of passing the event by value, this is important (explained below) */
-    bot.on_message_create.co_attach([](dpp::message_create_t event) -> dpp::task<void> {
+    bot.on_message_create.co_attach([](dpp::message_create_t event) -> dpp::job {
```
Jaskowicz1, line 22: Not sure i'm right, but should this not be `const dpp::message_create_t& event` or does that not work with coroutines?

Axyte: This could cause a lifetime issue in coroutines, as reference parameters are not copied over.

Jaskowicz1: Ahhhh right, thanks for the info! :)

Mishura4: Indeed! dpp::job runs in parallel so reference parameters may become dangling. See the red box in https://deploy-preview-763--dpp-dev.netlify.app/structdpp_1_1job.html

Jaskowicz1: Thank you!! Learn something new everyday hahaha

include/dpp/cluster_coro_calls.h
```diff
@@ -233,7 +233,7 @@ awaitable<confirmation_callback_t> co_guild_commands_get(snowflake guild_id);
  * @return confirmation returned object on completion
  * \memberof dpp::cluster
  */
-awaitable<confirmation_callback_t> co_interaction_response_create(snowflake interaction_id, const std::string &token, const interaction_response &r);
+[[nodiscard]] async<confirmation_callback_t> co_interaction_response_create(snowflake interaction_id, const std::string &token, const interaction_response &r);
```
Jaskowicz1, line 236: Would `std::string_view` not be better here or is it not permitted?

Mishura4: It would! Sadly cluster_coro_calls.* and cluster_sync_calls.* are auto-generated from dpp::cluster methods, they grab the parameters from them, so that is not really the purview of this PR 😔

Jaskowicz1: Ahhhh, gotcha :)

Mishura4:
Btw I suggest merging this PR normally, without squash, so each changes gets listed in the changelog one by one

braindigitalis (approved):

PR #770, opened 2023-08-11
Message "stream"

wizard7377:
Here I've added a feature where you can create both messages from channels and direct messages from users with a cluster object using a "stream like" interface


wizard7377:
Example program:
`
#include <dpp/dpp.h>

int main()
{
    dpp::cluster bot("TOKEN");
 
    bot.on_log(dpp::utility::cout_logger());
 
   
    bot.on_slashcommand([&bot](const dpp::slashcommand_t & event) {
        /* Check which command they ran */
        if (event.command.get_command_name() == "rep") {
            dpp::channel_stream(bot,event.command.get_channel()) << "Test one" << dpp::component().add_component(
                        dpp::component().set_label("Click me!").
                        set_type(dpp::cot_button).
                        set_emoji(u8"😄").
                        set_style(dpp::cos_danger).
                        set_id("myid")
                    ) << dpp::end_msg() << "Test two" << dpp::end_msg();
            dpp::channel_stream(bot,event.command.get_channel().id) << "Test three" << dpp::end_msg();
            dpp::dm_stream(bot,event.command.get_issuing_user()) << "Test 4" << dpp::end_msg();         

            dpp::dm_stream(bot,event.command.get_issuing_user().id) << "Test 5" << dpp::end_msg();         
            
        }
    });
 
    bot.on_ready([&bot](const dpp::ready_t & event) {
        if (dpp::run_once<struct register_bot_commands>()) {
 
            /* Create a new global command on ready event */
            dpp::slashcommand newcommand("rep", "test", bot.me.id);
            
 
            /* Register the command */
            bot.global_command_create(newcommand);
        }
    });
 
    bot.start(dpp::st_wait);
 
    return 0;
}
`

CMakeFiles/CMakeDirectoryInformation.cmake
```diff
@@ -0,0 +1,16 @@
+# CMAKE generated file: DO NOT EDIT!
+# Generated by "Unix Makefiles" Generator, CMake Version 3.27
+
+# Relative path conversion top directories.
+set(CMAKE_RELATIVE_PATH_TOP_SOURCE "/home/wizard/openSource/dpp_tests/cs_test")
```
braindigitalis, line 5: what is this...? I don't have a dir /home/wizard

include/dpp/streams.h
```diff
@@ -0,0 +1,158 @@
...
+         */
+        base_stream& operator<<(const embed &e);
+        /**
+         * @brief Build message in stream
+         * @param f Pair of filename and filecontent to add
+         * @return This stream, for purposes of chaining
+         */
+        base_stream& operator<<(const std::pair<std::string,std::string> &f);
+        virtual void send(command_completion_event_t callback = utility::log_error()) {};
+        virtual message send_sync() { return message(); };
```
braindigitalis, line 75: function bodies should not be in the .h

include/dpp/streams.h
```diff
@@ -0,0 +1,158 @@
...
+ */ 
+class DPP_EXPORT channel_stream : public base_stream {
+	public:
+		/**
+		 * @brief Constructer
+		 * 
+		 * @param bot Cluster that will send messages
+		 * @param out_channel Channel where messages will be sent
+		 */
+		channel_stream(cluster& bot, channel out_channel);
```
braindigitalis, line 128: const channel& out_channel

src/dpp/streams.cpp
```diff
@@ -0,0 +1,92 @@
...
+
+#include <dpp/streams.h>
+#include <dpp/channel.h>
+
+
+#include <utility>
+
+namespace dpp {
+
+int end_msg() { return 0; }
```
braindigitalis, line 32: please don't bunch stuff up on the same line

src/dpp/streams.cpp
```diff
@@ -0,0 +1,92 @@
...
+channel_stream::channel_stream(cluster& bot, channel out_channel) {
```
braindigitalis, line 33: { should be on separate line for start or function

src/dpp/streams.cpp
```diff
@@ -0,0 +1,92 @@
...
+}
+message channel_stream::send_sync() {
+	this->msg.set_channel_id(this->out_channel_id);
+	return this->bot->message_create_sync(this->msg);
+}
+
+
+
+
+dm_stream::dm_stream(cluster& bot, user out_user) {
```
braindigitalis, line 53: const user&

include/dpp/streams.h
```diff
@@ -0,0 +1,158 @@
...
+         * @param msg Text to add
+         * @return This stream, for purposes of chaining
+         */
+        base_stream& operator<<(const std::string& msg);
+        /**
+         * @brief Send message in stream
+         * @param n Number
+         * @return This stream, for purposes of chaining
+         */
+        base_stream& operator<<(const int& n);
```
braindigitalis, line 54: why only int, why not uint64_t, what about snowflake?

braindigitalis (changes requested):

CMakeFiles/CMakeDirectoryInformation.cmake
wizard7377: Ah sorry I forgot to remove all the build artefacts!

include/dpp/streams.h
wizard7377: The purpose of this is not to put a int into the stream, rather it is to act as an end of message indicator

include/dpp/streams.h
wizard7377: It's just the virtual placeholder, not the actual definition

braindigitalis: no it's a function body

include/dpp/streams.h
braindigitalis: this makes no sense. so I can't send an int with a value of 0 to the stream or it ends the stream?

cmake_install.cmake
```diff
@@ -0,0 +1,49 @@
+# Install script for directory: /home/wizard/openSource/dpp_tests/cs_test/libs/DPP
```
Mishura4, line 1: Please remove this file from the PR, this is specific to your computer, created when you did `cmake .`. In the future you should make a subdirectory, e.g. `mkdir build && cd build && cmake ..`.

include/dpp/channel.h
```diff
@@ -862,6 +870,9 @@ struct active_thread_info {
  * @brief A map of threads alongside optionally the thread_member tied to the bot if it is in the thread. The map's key is the thread id. Returned from the cluster::threads_get_active method
  */
 using active_threads = std::map<snowflake, active_thread_info>;
+/**
+ * @brief A stream wrapper to send messages
```
Mishura4, line 874: What does this document? The rest of the block under it is empty.

include/dpp/cluster.h
```diff
@@ -339,6 +339,7 @@ class DPP_EXPORT cluster {
 	 */
 	void log(dpp::loglevel severity, const std::string &msg) const;
 
+
```
Mishura4, line 342: You can remove this empty line here.

include/dpp/dpp.h
```diff
@@ -31,7 +31,9 @@
 #include <string>
 #include <vector>
 #include <map>
+
 #include <functional>
+
```
Mishura4, line 36: Why those empty lines?

include/dpp/streams.h
```diff
@@ -0,0 +1,158 @@
...
+#include <dpp/cluster.h>
+#include <utility>
+
+
+namespace dpp {
+/**
+ * @brief Return end of message
+ * @return 0
+*/
+int end_msg();
```
Mishura4, line 35: What does this mean? Return end of message? an int? What does `dpp::end_msg()` do?

include/dpp/streams.h
Mishura4: You should pass small variables as a value parameter, not as a const reference.

Mishura4: As brain said, this also creates strange behavior, for example :
```cpp
stream << "I have found " << get_count_of_something() << " things" << 0;
```
Will have very different behavior based on whether `get_count_of_something()` returns 0 versus non-0.
I suggest using a tag, such as :
```cpp
class stream {
  struct send_tag {};

  static constexpr send = send_tag{};
  
  base_stream &operator<<(send_tag); // your function that sends
  
  /* rest of the stream class */
};
```

include/dpp/streams.h
```diff
@@ -0,0 +1,158 @@
...
+         * @return This stream, for purposes of chaining
+         */
+        base_stream& operator<<(const embed &e);
+        /**
+         * @brief Build message in stream
+         * @param f Pair of filename and filecontent to add
+         * @return This stream, for purposes of chaining
+         */
+        base_stream& operator<<(const std::pair<std::string,std::string> &f);
+        virtual void send(command_completion_event_t callback = utility::log_error()) = 0;
```
Mishura4, line 74: Why is this virtual? Do we really need polymorphism here? See comment on line 54, why not make `operator<<(send_tag);` the send function, present in `channel_stream`, and `dm_stream`, and all other similar types?

include/dpp/streams.h
```diff
@@ -0,0 +1,158 @@
...
+        virtual message send_sync() = 0;
```
Mishura4, line 75: Same issue as the comment on `send`

include/dpp/streams.h
```diff
@@ -0,0 +1,158 @@
...
+         * @param e Embed to add
+         * @return This stream, for purposes of chaining
+         */
+        base_stream& operator<<(const embed &e);
+        /**
+         * @brief Build message in stream
+         * @param f Pair of filename and filecontent to add
+         * @return This stream, for purposes of chaining
+         */
+        base_stream& operator<<(const std::pair<std::string,std::string> &f);
```
Mishura4, line 73: You forgot the MIME type, there are 3 parameters to `dpp::message::add_file`.
I don't really like using pairs/tuples for this. Which field in the tuple is the filename, the filecontent? Why not create a type for it, with well-defined names?

include/dpp/user.h
```diff
@@ -25,9 +25,9 @@
 #include <dpp/managed.h>
 #include <dpp/utility.h>
 #include <dpp/json_interface.h>
-
 namespace dpp {
-
+class cluster; //TODO Find a way to remove this forward declaration
```
Mishura4, line 29: I would recommend completing this TODO yourself or removing it entirely, presumably the feature you're willing to add is complete and not half-baked.

library/CMakeLists.txt
```diff
@@ -390,4 +390,4 @@ if(DPP_INSTALL)
 
 	# CPack initialization for distribution
 	include(CPack)
-endif()
+endif()
```
Mishura4, line 393: Why was this modified? By UNIX convention, a text file must have a newline at the end to be recognized as a text file. Please add the newline back to the file.

include/dpp/streams.h
braindigitalis: why would zero be a special value? the whole point of a stream is it can take lots of types and especially integers

Jaskowicz1 (commented):
Why does `base_stream& operator<<(const int& n);` use & for the int (see streams.h)? ints shouldn't be passed by a const ref (see [this for more info](https://belaycpp.com/2022/02/15/constant-references-are-not-always-your-friends/#:~:text=Types%20that%20should%20be%20passed%20by%20value%20and%20not%20by%20reference)) as far as I'm aware.

src/dpp/streams.cpp
```diff
@@ -0,0 +1,95 @@
...
+#include <utility>
+
+namespace dpp {
+
+int end_msg() 
+{ 
+	return 0; 
+}
+channel_stream::channel_stream(cluster& bot, const channel& out_channel) {
+	this->bot = &bot;
```
Mishura4, line 37: Two problems :
1. `bot` is semantically ambiguous here. Please give a different name to the field from the parameter.
2. Can you use constructor initializer lists? This is what we do for new code and what we want to update the library to use where it's missing.
E.g.
```cpp
channel_stream::channel_stream(cluster& bot_, const channel& out_channel) : bot(&bot_), out_channel_id(out_channel.id) {}
```

src/dpp/streams.cpp
```diff
@@ -0,0 +1,95 @@
...
+dm_stream::dm_stream(cluster& bot, const user& out_user) {
+	this->bot = &bot;
+	this->out_user_id = out_user.id;
+}
+dm_stream::dm_stream(cluster& bot, snowflake out_user_id) {
+	this->bot = &bot;
+	this->out_user_id = out_user_id;
+}
+void dm_stream::send(command_completion_event_t callback) {
+
```
Mishura4, line 65: Extra new line you could remove here

Mishura4 (changes requested):

include/dpp/streams.h
Mishura4: (It seems my review comments were added back in time after your replies, so i'm replying here to make sure you see it hehe)

Mishura4:
Though this not explicitly mentioned in the coding standard, functions in DPP in general have their braces start on the same line, and have a newline in between definitions, e.g.
```cpp
void foo {
  // stuff
}

void bar {
  // other stuff
}
```
Though I can't explicitly require that you change your code to fit that since it's not in the coding standard, it would be appreciated.

include/dpp/streams.h
wizard7377: Yeah that happened

cmake_install.cmake
wizard7377: Opps sorry but the thing is the way i was doing it was testing it with a test script but i defo need to do this in the future

Jaskowicz1:
> Though this not explicitly mentioned in the coding standard, functions in DPP in general have their braces start on the same line, and have a newline in between definitions, e.g.
> 
> ```c++
> void foo {
> 	// stuff
> }
> 
> void bar {
> 	// other stuff
> }
> ```
> 
> Though I can't explicitly require that you change your code to fit that since it's not in the coding standard, it would be appreciated.

This is mentioned in the coding standard https://dpp.dev/coding-standards.html

include/dpp/channel.h
wizard7377: Whoops! I originally had the streams for channels and users in their respective headers, and I forgot to remove the docs when I removed the actual declarations

include/dpp/streams.h
```diff
@@ -0,0 +1,207 @@
...
+	public:
+		/**
+		 * @brief Constructer
+		 * 
+		 * @param bot Cluster that will send messages
+		 * @param out_channel Channel where messages will be sent
+		 */
+		channel_stream(cluster& bot_, const channel& out_channel); 
+		/**
+		 * @brief Constructer
```
Mishura4, line 179: Also constructor*, on the one below too

wizard7377:
Here's a new and improved test with the new features:
```
#include <dpp/dpp.h>

int main()
{
    dpp::cluster bot("TOKEN");
 
    bot.on_log(dpp::utility::cout_logger());
 
   
    bot.on_slashcommand([&bot](const dpp::slashcommand_t & event) {
        /* Check which command they ran */
        if (event.command.get_command_name() == "rep") {
            dpp::channel_stream(bot,event.command.get_channel()) << "Test 0" << dpp::end_msg() << "Test 1"
                    << dpp::component().set_type(dpp::cot_selectmenu).set_placeholder("Pick something").add_select_option(dpp::select_option("1","2","3")).set_id("6") 
					<< dpp::component().set_label("hello").set_type(dpp::cot_button).set_id("1").set_style(dpp::cos_primary)  
					<< dpp::component().set_label("hello2").set_type(dpp::cot_button).set_id("2").set_style(dpp::cos_primary)    
					<< dpp::end_row()
					<< dpp::component().set_label("hello3").set_type(dpp::cot_button).set_id("3").set_style(dpp::cos_primary)   
					<< dpp::end_msg();
            dpp::channel_stream(bot,event.command.get_channel().id) << "Test 2" << dpp::end_msg();
            
            dpp::dm_stream(bot,event.command.get_issuing_user()) << "Test 3" << dpp::end_msg();         

            dpp::dm_stream(bot,event.command.get_issuing_user().id) << "Test 4" << dpp::end_msg();         
            
        }
    });
 
    bot.on_ready([&bot](const dpp::ready_t & event) {
        if (dpp::run_once<struct register_bot_commands>()) {
 
            /* Create a new global command on ready event */
            dpp::slashcommand newcommand("rep", "test", bot.me.id);
            
 
            /* Register the command */
            bot.global_command_create(newcommand);
        }
    });
 
    bot.start(dpp::st_wait);
 
    return 0;
}
```

Jaskowicz1:
I would love to see what this looks like in discord itself. Whilst I like this new example with adding components, I don't like that `dpp::end_msg()` is used mid-way. I would interpret that as the message is now done but I'm assuming it means `\n`. Either way, seeing it do more than just send message and actually simplify adding a component does make me like this. I feel this would benefit newer people.

wizard7377:
> I would love to see what this looks like in discord itself. Whilst I like this new example with adding components, I don't like that `dpp::end_msg()` is used mid-way. I would interpret that as the message is now done but I'm assuming it means `\n`. Either way, seeing it do more than just send message and actually simplify adding a component does make me like this. I feel this would benefit newer people.

It does indeed mean that the message is done

Jaskowicz1:
> > I would love to see what this looks like in discord itself. Whilst I like this new example with adding components, I don't like that `dpp::end_msg()` is used mid-way. I would interpret that as the message is now done but I'm assuming it means `\n`. Either way, seeing it do more than just send message and actually simplify adding a component does make me like this. I feel this would benefit newer people.
> 
> It does indeed mean that the message is done

Okay, so going back to your example's first message. It shows `dpp::channel_stream(bot,event.command.get_channel()) << "Test 0" << dpp::end_msg() << "Test 1"
                    << dpp::component().set_type(dpp::cot_selectmenu).set_placeholder("Pick something").add_select_option(dpp::select_option("1","2","3")).set_id("6") 
					<< dpp::component().set_label("hello").set_type(dpp::cot_button).set_id("1").set_style(dpp::cos_primary)  
					<< dpp::component().set_label("hello2").set_type(dpp::cot_button).set_id("2").set_style(dpp::cos_primary)    
					<< dpp::end_row()
					<< dpp::component().set_label("hello3").set_type(dpp::cot_button).set_id("3").set_style(dpp::cos_primary)   
					<< dpp::end_msg();` this, does this mean that it will send two separate messages (one being "Test 0", the other being "Test 1" with buttons)?

If so, I feel that this shouldn't be allowed as it becomes really weird looking, allowing more messages after an end message call.

Mishura4:
I don't really like the function model still. It's harder for the compiler to inline than a constexpr variable, it doesn't look like std::cout << "foo" << std::endl;`, also as a standalone `dpp::end_msg()` is confusing, what message is being ended? The name "end_msg" doesn't communicate that it's being sent very well. To me it sounds more like finishing a builder. Also why provide one function for a `message`, what if we add streams for other facilities, will each of them need a separate function?

wizard7377:
#general _ Trash 7 - Discord 8_12_2023 12_24_11 PM
#general _ Trash 7 - Discord 8_12_2023 12_24_27 PM

wizard7377:
> I don't really like the function model still. It's harder for the compiler to inline than a constexpr variable, it doesn't look like `std::cout << "foo" << std::endl;`, also as a standalone `dpp::end_msg()` is confusing, what message is being ended? The name "end_msg" doesn't communicate that it's being sent very well. To me it sounds more like finishing a builder. Also why provide one function for a `message`, what if we add streams for other facilities, will each of them need a separate function?

What do you mean, if we add streams for other facilities?

Mishura4:
For example if you want to stream a thread and << users to it, to add them to the thread. Making streams work with other stream than messages.

wizard7377:
> why provide one function for a `message`, what if we add streams for other facilities, will each of them need a

I don't get your point, are you saying that there are too many << overloads or too few?

Mishura4:
I'm not saying either of those.

What i'm saying is, when designing a feature that works on a large library, one needs to take into account how it interacts with the rest of the library too. If you can stream a channel and add messages to it to simplify calling cluster methods, then it makes sense that other cluster calls could be simplified with other things too. For example adding users to a thread. Then, because you chose `end_msg` as a name for a message, you'll have to add *another* function for that, `end_user_add`? Why not just unify calling conventions?

In fact now the more I think about it, the more i think it would make sense to have a message builder with your additions, that you can then stream to the channel stream you get on a channel.
```cpp
dpp::message m = message::stream() << "Message text" << some_embed << component1 << component2;

channel.stream(cluster) << m << dpp::message{"another message"};
```
etc.

include/dpp/streams.h
wizard7377: It does work that way, it is merely a different implementation for a direct message than a channel, which is why it must be virtual, as it still must be accessible from the base class.

Mishura4: Why does it have to be accessible from the base class?

Commandserver:
I'm closing this PR as we see no benefit in this and it doesn't fit with the style of the lib.

PR #774, opened 2023-08-14
perf: remove copying of audio buffer before encrypting

Fluiser:
Removing meaningless allocation and copying memory.

Communication with the system is expensive.

### DIFF
--- a/src/dpp/discordvoiceclient.cpp
+++ b/src/dpp/discordvoiceclient.cpp
@@ -1183,12 +1183,11 @@ discord_voice_client& discord_voice_client::send_audio_raw(uint16_t* audio_data,
 	}
 
 	opus_int32 encodedAudioMaxLength = (opus_int32)length;
-	std::vector<uint8_t> encodedAudioData(encodedAudioMaxLength);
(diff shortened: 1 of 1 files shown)


braindigitalis:
is this fully tested, with both send and receive of data, and mixing audio?

the data copies are not meaningless at all and are required to prevent buffers overwriting themselves inside opus and sodium - please ensure everything works correctly. 

I do not think this is safe.

braindigitalis (changes requested):
please ensure unit tests pass and that a sample program is written to show this. you removed the pull request template text so I have no way of knowing if this was done.

Mishura4:
According to the [sodium docs](https://libsodium.gitbook.io/doc/secret-key_cryptography/secretbox), the buffers can overlap. However yes it would be good to confirm everything still works.

Mishura4:
I've also renamed the PR to comply with the [coding style standards](https://dpp.dev/coding-standards.html).

Fluiser:
std::vector lives in scope, as does its argument. If we assume that the memory is not available by the label, then it may also be inaccessible at the time of copying to the vector.

PR #775, opened 2023-08-16
fix: fix various issues and bugs with dpp::emoji

Mishura4:
- [✅] My pull request is made against the `dev` branch.
- [❓ github actions will tell :kekw:] I have ensured that the changed library can be built on your target system. I did not introduce any platform-specific code.
- [✅] I have ensured that all methods and functions are **fully documented** using doxygen style comments.
- [✅] I tested my commits, by adding a test case to the unit tests if needed
- [✅] I have ensured that I did not break any existing API calls.
- [✅] My code follows the [coding style guide](https://dpp.dev/coding-standards.html) (if you are not sure, match the code style of existing files including indent style etc).
- [✅] I have not built my pull request using AI, a static analysis tool or similar without any human oversight. Where I have generated this pull request using a tool, I have justified why this is needed.
- [✅] I agree to the terms of the [DCO (Developer Certificate of Origin)]((https://dpp.dev/coding-standards.html))

This PR fixes various bugs involving dpp::emoji, including a double free after copying an emoji with image data, as well as dpp::base64_encode reading past the end of the data sent to it. It also modernizes (a little bit) dpp::emoji, and adds utility functions to get a mime type and file extension from a dpp::image_type or a dpp::sticker_format.

### DIFF
--- a/include/dpp/emoji.h
+++ b/include/dpp/emoji.h
@@ -55,38 +55,66 @@ public:
 	/**
 	 * @brief Emoji name
 	 */
-	std::string name;
+	std::string name{};
 	/**
 	 * @brief User id who uploaded the emoji
 	 */
-	snowflake user_id;
+	snowflake user_id{0};
 	/**
 	 * @brief Flags for the emoji from dpp::emoji_flags
 	 */
-	uint8_t flags;
+	uint8_t flags{0};
 	/**
 	 * @brief Image data for the emoji if uploading
 	 */
-	std::string* image_data;
-	
+	std::string image_data{};
+
 	/**
 	 * @brief Construct a new emoji object
 	 */
-	emoji();
+	emoji() = default;
 
 	/**
 	 * @brief Construct a new emoji object with name, ID and flags
-	 * 
-	 * @param n The emoji's name
-	 * @param i ID, if it has one (unicode does not)
-	 * @param f Emoji flags (emoji_flags)
+	 *
+	 * @param name The emoji's name
+	 * @param id ID, if it has one (unicode does not)
+	 * @param flags Emoji flags (emoji_flags)
 	 */
-	emoji(const std::string n, const snowflake i = 0, const uint8_t f = 0);
+	emoji(const std::string_view name, const snowflake id = 0, const uint8_t flags = 0);
+
+	/**
+	 * @brief Copy constructor, copies another emoji's data
+	 *
+	 * @param rhs Emoji to copy
+	 */
+	emoji(const emoji &rhs) = default;
+
+	/**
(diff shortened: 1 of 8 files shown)


include/dpp/emoji.h
```diff
@@ -55,38 +55,66 @@ class DPP_EXPORT emoji : public managed, public json_interface<emoji>  {
...
 	 * @brief Flags for the emoji from dpp::emoji_flags
 	 */
-	uint8_t flags;
+	uint8_t flags{0};
 	/**
 	 * @brief Image data for the emoji if uploading
 	 */
-	std::string* image_data;
-	
+	std::string image_data{};
```
braindigitalis, line 70: the reason we didn't have this as string is that a string is 12 bytes and a pointer is 8 or 4. this is only set when you are uploading an emoji, but, there are millions of emojis cached in a running bot at scale. this adds megabytes to the bot. can you possibly rethink this to not have a pointless empty std::string on every emoji instance?

include/dpp/emoji.h
```diff
@@ -55,38 +55,66 @@ class DPP_EXPORT emoji : public managed, public json_interface<emoji>  {
...
+	 * @param rhs Emoji to move from
+	 */
+	emoji(emoji &&rhs) noexcept = default;
 
 	/**
 	 * @brief Destroy the emoji object
 	 */
-	virtual ~emoji();
+	~emoji() override = default;
+
```
braindigitalis, line 104: if we do this we prevent anyone ever overriding emoji as this dtor is no longer virtual?

include/dpp/emoji.h
```diff
@@ -95,35 +123,35 @@ class DPP_EXPORT emoji : public managed, public json_interface<emoji>  {
...
 
 	/**
 	 * @brief Build the json for this object
-	 * 
+	 *
 	 * @param with_id include the id in the JSON
 	 * @return std::string json data
 	 */
-	std::string build_json(bool with_id = false) const;
+	std::string build_json(bool with_id = false) const override;
```
braindigitalis, line 142: does this no longer match the signature of json_interface?

braindigitalis (changes requested):

include/dpp/emoji.h
Axyte: Do you suggest removing this field entirely and making it a parameter on the request functions? Would help with the memory rather more than having pointers in each object.

Mishura4: I do like that idea

Commandserver (approved):

braindigitalis (approved):

PR #793, opened 2023-08-22
feat(build): add off-by-default DPP_USE_PCH option in CMake

Mishura4:
Adds an option to use private PCH when building DPP from source.
Mostly useful on MSVC, makes the build take 30 seconds instead of 5 minutes.
Useful for contributors mostly 👀 as this only affects DPP itself due to being private, and does not propagate to dependents.

This does not affect vcpkg.


Mishura4:
I *could* add it to the vcpkg CMake as well, I'm just not familiar with vcpkg so I'm not sure how I'd test that 🤔

braindigitalis:
> I _could_ add it to the vcpkg CMake as well, I'm just not familiar with vcpkg so I'm not sure how I'd test that 🤔

yeah lets leave vcpkg alone. it works, and its fragile af. Lets not poke the hornets nest.

braindigitalis (approved):

PR #808, opened 2023-08-26
docs: refactored page links

Commandserver:

Why? because doxygen uses the filenames to generate the links when we don't use `\page`, and that's ugly e.g. `md_docpages_04_building_a_bot.html` should be just `building-a-bot.html` shouldn't it?.

There're many ways to link a page e.g.: `[Shards](\ref clusters-shards-guilds)` or `\ref clusters-shards-guilds "Shards"`, which i've used for this pr.

I've tested all links to make sure they're working.


Jaskowicz1:
So, does \ref mean references will work when we run doxygen locally? I remember you mentioning "ref" in one of my PRs but I had no idea of it.

Commandserver:
> So, does \ref mean references will work when we run doxygen locally? I remember you mentioning "ref" in one of my PRs but I had no idea of it.
> 
> If so, I'm so all for this PR and I will be praising you forever 😆

Yes. they're better as we don't hard-code a filename. So when we rename a page, we don't have to change all links with it.

Jaskowicz1:
> Yes. they're better as we don't hard-code a filename. So when we rename a page, we don't have to change all links with it.

This is absolutely amazing then, thank you a lot!!

docpages/INDEX.md
```diff
@@ -36,10 +36,10 @@ You can find further releases in other architectures and formats or the source c
 * Efficient caching system for guilds, channels, guild members, roles, users
 * Sharding and clustering (Many shards, one process: specify the number of shards, or let the library decide)
 * Highly optimised ETF (Erlang Term Format) support for very fast websocket throughput (*no other C++ Discord library has this!*)
-* [Slash Commands/Interactions support](https://dpp.dev/slashcommands.html)
-* [Voice support](https://dpp.dev/soundboard.html) (sending **and** receiving audio)
+* \ref slashcommands "Slash Commands/Interactions support"
```
braindigitalis, line 39: if we are doing this please make sure the names of the links match what they used to be . usually when doing \ref, the title of the link is always the title of the page but this breaks the sentences where we put a link in a sentence.

docpages/INDEX.md
```diff
@@ -36,10 +36,10 @@ You can find further releases in other architectures and formats or the source c
...
+* \ref soundboard "Voice support" (sending **and** receiving audio)
 * The entire Discord API is available for use in the library
-* Stable [Windows support](https://dpp.dev/buildwindows.html)
+* Stable \ref buildwindows "Windows support"
 * Ready-made compiled packages for Windows, Raspberry Pi (ARM64/ARM7/ARMv6), Debian x86/x64 and RPM based distributions
```
braindigitalis, line 43: (yeah like this)

braindigitalis (approved):

Jaskowicz1:
Checked through every page to see if anything was missing and there isn't! I'm more than happy with this PR :)

PR #812, opened 2023-08-26
misc(coro): deprecate sync methods when DPP_CORO is enabled

Mishura4:


braindigitalis:
not sure I like this idea, as there are tons of bots using sync calls

Mishura4:
We actively discourage using sync already, so the point is that they shouldn't, isn't it?

braindigitalis:
that is true but we won't actually be removing them, the lib itself actively uses a couple of sync methods on startup to get number of shards etc.

Mishura4:
I'll change it in the next few days

Mishura4:
Wait i didn't mean to close it lmao

Mishura4:
(The plan is still to change the warning very soon)

PR #841, opened 2023-09-03
docs: added a setting status page.

Jaskowicz1:
This PR adds a new page for setting the bot's status.


Jaskowicz1:
Corrected the misc.md subpage creation, removing the `"<title>"` from the line.

Henonicks:
Too much cod, but it's beatiful!11!!1!1!11

braindigitalis (changes requested):
this would be better if we then show in a second example how to set the status on a timer e.g. with a guild count in it. pay attention though to the fact that the presence is set per shard so you'll need to loop all shards to set presence properly

Jaskowicz1:
> this would be better if we then show in a second example how to set the status on a timer e.g. with a guild count in it. pay attention though to the fact that the presence is set per shard so you'll need to loop all shards to set presence properly

I like that idea! I'll get onto that now.

Jaskowicz1:
> this would be better if we then show in a second example how to set the status on a timer e.g. with a guild count in it.

The suggested example has now been added and is ready for a review!

docpages/example_programs/misc/setting_status.md
```diff
@@ -0,0 +1,76 @@
...
+{
+    /* Create the bot */
+    dpp::cluster bot("token");
+
+    bot.on_log(dpp::utility::cout_logger());
+
+    bot.on_ready([&bot](const dpp::ready_t& event) {
+        /* We don't need the run_once here as we're not registering commands! */
+
+        std::thread status_thread([&bot]() {
```
braindigitalis, line 49: the thread needs to be in a run_once or the bot will spawn a new thread every time any shard reconnects

Jaskowicz1: Ah yes, good point.

braindigitalis (approved):

PR #845, opened 2023-09-06
fix: fixed discord_voice_client::send_audio_raw blocking thread when provided with invalid size

Neko-Life:
## Changes in this PR

- Fix `send_audio_raw` blocking thread when provided with invalid size
- Define `SEND_AUDIO_RAW_MAX_LENGTH` so users can choose not to hardcode the size in their code
- Update comment

## PR Checklist

### DIFF
--- a/include/dpp/discordvoiceclient.h
+++ b/include/dpp/discordvoiceclient.h
@@ -60,7 +60,7 @@ inline constexpr uint16_t AUDIO_TRACK_MARKER = 0xFFFF;
 
 inline constexpr int AUDIO_OVERLAP_SLEEP_SAMPLES = 30;
 
-inline constexpr size_t SEND_AUDIO_RAW_MAX_LENGTH = 11520;
+inline constexpr size_t send_audio_raw_max_length = 11520;
 
 using json = nlohmann::json;
 
@@ -676,7 +676,7 @@ public:
 	/**
 	 * @brief Send raw audio to the voice channel.
 	 * 
-	 * You should send an audio packet of `SEND_AUDIO_RAW_MAX_LENGTH` (11520) bytes.
+	 * You should send an audio packet of `send_audio_raw_max_length` (11520) bytes.
 	 * Note that this function can be costly as it has to opus encode
 	 * the PCM audio on the fly, and also encrypt it with libsodium.
 	 * 
@@ -695,15 +695,15 @@ public:
 	 * 
 	 * @param length The length of the audio data. The length should
 	 * be a multiple of 4 (2x 16 bit stereo channels) with a maximum
-	 * length of `SEND_AUDIO_RAW_MAX_LENGTH`, which is a complete opus
+	 * length of `send_audio_raw_max_length`, which is a complete opus
 	 * frame at highest quality. `dpp::voice_exception` will be thrown
 	 * if length is less than 4 or not divisible by 4.
 	 *
 	 * Generally when you're streaming and you know there will be
 	 * more packet to come you should always provide packet data with
-	 * length of `SEND_AUDIO_RAW_MAX_LENGTH`.
+	 * length of `send_audio_raw_max_length`.
 	 * Silence packet will be appended if length is less than
-	 * `SEND_AUDIO_RAW_MAX_LENGTH` as discord expects to receive such
+	 * `send_audio_raw_max_length` as discord expects to receive such
 	 * specific packet size. This can cause gaps in your stream resulting
 	 * in distorted audio if you have more packet to send later on.
 	 * 
--- a/include/dpp/discordvoiceclient.h
+++ b/include/dpp/discordvoiceclient.h
@@ -56,7 +56,7 @@ struct OpusRepacketizer;
 
 namespace dpp {
 
-// !TODO: change these to constexpr and rename every occurence across the codebase
+// !TODO: change these to constexpr and rename every occurrence across the codebase
 #define AUDIO_TRACK_MARKER (uint16_t)0xFFFF
 
 #define AUDIO_OVERLAP_SLEEP_SAMPLES 30
--- a/include/dpp/discordvoiceclient.h
+++ b/include/dpp/discordvoiceclient.h
@@ -56,9 +56,10 @@ struct OpusRepacketizer;
 
 namespace dpp {
 
-inline constexpr uint16_t AUDIO_TRACK_MARKER = 0xFFFF;
+// !TODO: change these to constexpr and rename every occurence across the codebase
+#define AUDIO_TRACK_MARKER (uint16_t)0xFFFF
 
-inline constexpr int AUDIO_OVERLAP_SLEEP_SAMPLES = 30;
+#define AUDIO_OVERLAP_SLEEP_SAMPLES 30
 
 inline constexpr size_t send_audio_raw_max_length = 11520;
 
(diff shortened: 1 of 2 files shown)


include/dpp/discordvoiceclient.h
```diff
@@ -50,7 +50,7 @@
 #include <functional>
 #include <chrono>
 
-
+#define SEND_AUDIO_RAW_MAX_LENGTH 11520
```
Mishura4, line 53: If you want to let users define their own size, then you should wrap this in `#ifndef`. Otherwise, this should be a constexpr variable, not a macro.

src/dpp/discordvoiceclient.cpp
```diff
@@ -1166,20 +1166,35 @@ discord_voice_client& discord_voice_client::set_send_audio_type(send_audio_type_
 
 discord_voice_client& discord_voice_client::send_audio_raw(uint16_t* audio_data, const size_t length)  {
 #if HAVE_VOICE
-	const size_t max_frame_bytes = 11520;
+	static const size_t max_frame_bytes = SEND_AUDIO_RAW_MAX_LENGTH;
```
Mishura4, line 1169: Why the addition of `static` here? I'm guessing as an "optimization"? You're going to get a cache miss which is much more costly than the initialization of a size_t in the register.
Also, if we're going to edit this line, can we make it constexpr?

src/dpp/discordvoiceclient.cpp
```diff
@@ -1166,20 +1166,35 @@ discord_voice_client& discord_voice_client::set_send_audio_type(send_audio_type_
...
+
+	if (length < 4) {
+		return *this;
+	}
+
+	if ((length % 4) != 0) {
```
Mishura4, line 1175: This can be `if ((length & 0b11) != 0)`, much better than a modulo

src/dpp/discordvoiceclient.cpp
```diff
@@ -1166,20 +1166,35 @@ discord_voice_client& discord_voice_client::set_send_audio_type(send_audio_type_
...
+		throw dpp::voice_exception("Raw audio packet size should be left with zero remainder when divided by 4");
```
Mishura4, line 1176: This is confusing. I suggest "should be divisible by 4" or "should be a multiple of 4"

src/dpp/discordvoiceclient.cpp
```diff
@@ -1166,20 +1166,35 @@ discord_voice_client& discord_voice_client::set_send_audio_type(send_audio_type_
...
 	if (length > max_frame_bytes) {
 		std::string s_audio_data((const char*)audio_data, length);
 		while (s_audio_data.length() > max_frame_bytes) {
 			std::string packet(s_audio_data.substr(0, max_frame_bytes));
-			s_audio_data.erase(s_audio_data.begin(), s_audio_data.begin() + max_frame_bytes);
-			if (packet.size() < max_frame_bytes) {
-				packet.resize(max_frame_bytes, 0);
-			}
-			send_audio_raw((uint16_t*)packet.data(), max_frame_bytes);
+			const auto packet_size = (long long)packet.size();
```
Mishura4, line 1183: Why the cast to long long here? And as a C cast nonetheless?

src/dpp/discordvoiceclient.cpp
```diff
@@ -1166,20 +1166,35 @@ discord_voice_client& discord_voice_client::set_send_audio_type(send_audio_type_
...
+			s_audio_data.erase(s_audio_data.begin(), s_audio_data.begin() + packet_size);
+
+			send_audio_raw((uint16_t*)packet.data(), packet_size);
 		}
 
 		return *this;
+	}
+
+	if (length < max_frame_bytes) {
+		std::string packet((const char*)audio_data, length);
```
Mishura4, line 1194: Do you really need to make a copy of the whole buffer and then truncate it? Can't you just send a maximum of max_frame_bytes to send_audio_raw? Note also that as of mid 2023, declaring storage as `const char*` and then casting it to any other type is UB

Mishura4 (changes requested):

src/dpp/discordvoiceclient.cpp
Mishura4: I'm also confused, you have this exception here but if we send a size of 3 it just silently doesn't do anything? Shouldn't it send an error too if length < 4?

src/dpp/discordvoiceclient.cpp
Mishura4: Nvm I see now that's what the code above is doing, I need to look into that

include/dpp/discordvoiceclient.h
Neko-Life: I don't want users to define their own size, I want users to have access to the size without hardcoding it in their codebase. What's the benefit of constexpr variable than macros here that is should be a constexpr variable?

src/dpp/discordvoiceclient.cpp
Neko-Life: clangd is complaining about different signedness at the code below. I don't think C-casting size_t to long long would be problematic and it is simple and instantly understandable by stupid people such as me and the complain went away

Neko-Life (commented):
I will update the code later after work

include/dpp/discordvoiceclient.h
```diff
@@ -50,14 +50,18 @@
...
 struct OpusEncoder;
 struct OpusRepacketizer;
 
 namespace dpp {
 
+inline constexpr uint16_t AUDIO_TRACK_MARKER = 0xFFFF;
+
+inline constexpr int AUDIO_OVERLAP_SLEEP_SAMPLES = 30;
+
+inline constexpr size_t SEND_AUDIO_RAW_MAX_LENGTH = 11520;
```
Neko-Life, line 63: Hold on, how the naming should be? `dpp::SEND_AUDIO_RAW_MAX_LENGTH` doesn't seems right. I guess I will change it to snake case for this variable only

include/dpp/discordvoiceclient.h
```diff
@@ -50,14 +50,18 @@
...
-
 struct OpusDecoder;
 struct OpusEncoder;
 struct OpusRepacketizer;
 
 namespace dpp {
 
+inline constexpr uint16_t AUDIO_TRACK_MARKER = 0xFFFF;
+
+inline constexpr int AUDIO_OVERLAP_SLEEP_SAMPLES = 30;
```
Jaskowicz1, line 61: This doesn't follow the [coding style standards](https://dpp.dev/coding-standards.html#:~:text=Class%20names%2C%20function%20names%20and%20method%20names). I am aware that it was previously all caps before which does confuse me :P

Neko-Life: I will revert those to macro as those are out of scope of this PR

Jaskowicz1: Sounds good!

include/dpp/discordvoiceclient.h
Mishura4: Awesome, thanks for changing it! To answer your question, constexpr variables have a type, which makes them less prone to subtle bugs introduced by like implicit conversions or integer promotions. They're also namespaced, so there is no risk of conflicting with a user's code.

src/dpp/discordvoiceclient.cpp
Mishura4: Hmm I suggest using a static_cast to ptrdiff_t then, that's the type recommended for pointer or iterator arithmetics
I usually just use size_t but I'm assuming clangd will like ptrdiff_t better here

Mishura4 (changes requested):

Mishura4 (approved):

PR #848, opened 2023-09-09
fix: corrected presence.cpp to allow custom statuses.

Jaskowicz1:
This PR changes presence.cpp to allow custom statuses. The way discord asks for a presence means we need to set the "state" in the json object, however, we don't do this (hence this PR!). This now means you can do the following:

```
bot.set_presence(dpp::presence(dpp::presence_status::ps_online, dpp::activity_type::at_custom, "Testing 123!"));
```
image

This PR also adds "cmake-build-debug" to the `.gitignore`.

I also have no idea why the indentation is weird, I simply added the code in CLion and it looks great on my end, but on GitHub and a text editor it does that. If it's okay to leave then that's great, otherwise, I'm happy for someone to give me a fix for it!

### DIFF
--- a/src/dpp/presence.cpp
+++ b/src/dpp/presence.cpp
@@ -73,14 +73,14 @@ presence::presence() : user_id(0), guild_id(0), flags(0)
 presence::presence(presence_status status, activity_type type, const std::string& activity_description) {
 	dpp::activity a;
 
-    /* Even if type is custom, a name is still required.
-     * We'll just set the name as this activity_description as it won't be used if custom either way. */
+	/* Even if type is custom, a name is still required.
+	 * We'll just set the name as this activity_description as it won't be used if custom either way. */
 	a.name = activity_description;
 
-    /* If the type is custom, set the state as "activity_description" */
-    if(type == activity_type::at_custom) {
-        a.state = activity_description;
-    }
+	/* If the type is custom, set the state as "activity_description" */
+	if(type == activity_type::at_custom) {
+	    a.state = activity_description;
+	}
 
 	a.type = type;
 	activities.clear();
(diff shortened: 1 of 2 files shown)


Jaskowicz1:
I do want to add, we can still set the state all the time and normal statuses (like `at_game`, `at_listening`) will ignore it. The reason I'm not doing that is to reduce the JSON string size so normal statuses aren't heavily increased in size when they don't need to be.

We also don't need to change `fill_from_json` because it already has code to check for state.

src/dpp/presence.cpp
```diff
@@ -254,7 +263,12 @@ std::string presence::build_json(bool with_id) const {
 				{ "type", i.type }
 			});
 			if (!i.url.empty()) j2["url"] = i.url;
-			if (!i.state.empty()) j2["details"] = i.state; // bot activity is details, not state
+
+            if(i.type == activity_type::at_custom) {
```
braindigitalis, line 267: these are space indented, this is why it is not working right.
Please change your editor to only use tab indents with tab size 8. You can then re-indent it

src/dpp/presence.cpp
```diff
@@ -72,7 +72,16 @@ presence::presence() : user_id(0), guild_id(0), flags(0)
 
 presence::presence(presence_status status, activity_type type, const std::string& activity_description) {
 	dpp::activity a;
+
+    /* Even if type is custom, a name is still required.
```
braindigitalis, line 76: also space indented, see below

braindigitalis (changes requested):

src/dpp/presence.cpp
Jaskowicz1: Ah is that what that is. Thank you!

Jaskowicz1: Indentation has been corrected.

braindigitalis (approved):

PR #854, opened 2023-09-14
docs: add auto testing of examples

braindigitalis:
This PR splits out the examples into `docpages/example_code`. These are individual .cpp files which should be able to be compiled as standalone programs into an executable.

There is also a CMakeLists.txt that builds them, and an action that runs this test build on push. This way we can identify bad examples that dont even compile. It also allows us to properly use syntax highlighting and linting on them and ensure they meet our code style guide.


braindigitalis:
An example of a successful run of the action:
https://github.com/brainboxdotcc/DPP/actions/runs/6179898895/job/16775550064?pr=854

Jaskowicz1:
Not reviewed it yet, but this is an extremely good feature to go through with for the docs.

My biggest question is, will the GitHub actions test this on different compilers (for example, Clang, G++, and MSVC)? If not, we should look into this as we've had an issue in the past with this.

braindigitalis:
> Not reviewed it yet, but this is an extremely good feature to go through with for the docs.
> 
> My biggest question is, will the GitHub actions test this on different compilers (for example, Clang, G++, and MSVC)? If not, we should look into this as we've had an issue in the past with this.

no I think that is overkill. we would have to build 3 copies of dpp with coro, and compile about 40 example programs on each. these are impossible to test without spamming pushes to GitHub and waiting for hours each time, as there is no way to locally simulate a windows GitHub action. smells like arm6 cross compile all over again

Jaskowicz1:
> > Not reviewed it yet, but this is an extremely good feature to go through with for the docs.
> > My biggest question is, will the GitHub actions test this on different compilers (for example, Clang, G++, and MSVC)? If not, we should look into this as we've had an issue in the past with this.
> 
> no I think that is overkill. we would have to build 3 copies of dpp with coro, and compile about 40 example programs on each. these are impossible to test without spamming pushes to GitHub and waiting for hours each time, as there is no way to locally simulate a windows GitHub action. smells like arm6 cross compile all over again

That's understandable. Still great either way!

RealTimeChris (approved):

PR #856, opened 2023-09-15
added [[maybe_unused]] in some places where it was needed

1mi0:
- [*] My pull request is made against the `dev` branch.
- [*] I have ensured that the changed library can be built on your target system. I did not introduce any platform-specific code.
- [*] I have ensured that all methods and functions are **fully documented** using doxygen style comments.
- [*] I tested my commits, by adding a test case to the unit tests if needed
- [*] I have ensured that I did not break any existing API calls.
- [?] My code follows the [coding style guide](https://dpp.dev/coding-standards.html) (if you are not sure, match the code style of existing files including indent style etc).
- [*] I have not built my pull request using AI, a static analysis tool or similar without any human oversight. Where I have generated this pull request using a tool, I have justified why this is needed.

### DIFF
--- a/include/dpp/json_interface.h
+++ b/include/dpp/json_interface.h
@@ -44,7 +44,7 @@ namespace dpp {
 		 * @param j nlohmann::json object
 		 * @return T& Reference to self for fluent calling
 		 */
-		T& fill_from_json(nlohmann::json* j) {
+		T& fill_from_json([[maybe_unused]] nlohmann::json* j) {
 			throw dpp::logic_exception("JSON interface doesn't implement parse_from_json");
 		}
 
@@ -54,7 +54,7 @@ namespace dpp {
 		 * @param with_id Include the ID in the JSON
 		 * @return std::string JSON string version of object
 		 */
-		virtual std::string build_json(bool with_id = false) const {
+		virtual std::string build_json([[maybe_unused]] bool with_id = false) const {
 			throw dpp::logic_exception("JSON interface doesn't implement build_json");
 		}
 	};
--- a/include/dpp/timed_listener.h
+++ b/include/dpp/timed_listener.h
@@ -73,7 +73,7 @@ public:
 		/* Attach event */
 		listener_handle = ev(listener);
 		/* Create timer */
-		th = cl->start_timer([this](dpp::timer timer_handle) {
+		th = cl->start_timer([this]([[maybe_unused]] dpp::timer timer_handle) {
 			/* Timer has finished, detach it from event.
 			 * Only allowed to tick once.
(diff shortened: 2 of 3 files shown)


braindigitalis:
thanks! please retarget this pr against the dev branch

include/dpp/json_interface.h
```diff
@@ -44,7 +44,7 @@ namespace dpp {
 		 * @param j nlohmann::json object
 		 * @return T& Reference to self for fluent calling
 		 */
-		T& fill_from_json(nlohmann::json* j) {
+		T& fill_from_json([[maybe_unused]] nlohmann::json* j) {
```
braindigitalis, line 47: how is this ever unused? we always pass json in.

include/dpp/json_interface.h
```diff
@@ -54,7 +54,7 @@ namespace dpp {
 		 * @param with_id Include the ID in the JSON
 		 * @return std::string JSON string version of object
 		 */
-		virtual std::string build_json(bool with_id = false) const {
+		virtual std::string build_json([[maybe_unused]] bool with_id = false) const {
```
braindigitalis, line 57: how can this be maybe unused when it's an optional parameter?

braindigitalis (changes requested):

include/dpp/timed_listener.h
```diff
@@ -73,7 +73,7 @@ template <typename attached_event, class listening_function> class timed_listene
 		/* Attach event */
 		listener_handle = ev(listener);
 		/* Create timer */
-		th = cl->start_timer([this](dpp::timer timer_handle) {
+		th = cl->start_timer([this]([[maybe_unused]] dpp::timer timer_handle) {
```
Jaskowicz1, line 76: Why is this "maybe unused"? When would you use a timer without a lambda or storing the timer 🤔

1mi0:
To answer all questions:
[[maybe_unused]] in our situation tells the compiler that we don't use the parameters inside of their functions. Not that we won't use them when we call the function or in other overloads | overrides of the function. In these specific 4 cases it just so happens that those arguments are not used inside of their lambda/function causing unused variable errors in my instance and warnings in other instances.

include/dpp/json_interface.h
Mishura4: It's unused in this definition, causing a warning

include/dpp/json_interface.h
Mishura4: Same as my other comment

Mishura4 (approved):

braindigitalis (approved):

PR #871, opened 2023-09-20
docs: fixed a ref issue and added more info in discordvoiceclient.h

Jaskowicz1:
discordvoiceclient.h was not referencing `dpp::cluster::on_voice_buffer_send` correctly, this has been altered. I've also changed the notice of `the audio data needs to be...` to be a warning, to which it also states that the audio will be affected if it isn't at that hz (just to inform people what will happen otherwise).

### DIFF
--- a/include/dpp/discordvoiceclient.h
+++ b/include/dpp/discordvoiceclient.h
@@ -685,13 +685,13 @@ public:
 	 * ready to send and know its length it is advisable to call this
 	 * method multiple times to enqueue the entire stream audio so that
 	 * it is all encoded at once (unless you have set use_opus to false).
-	 * Constantly calling this from the dpp::on_voice_buffer_send callback
-	 * can and will eat a TON of cpu!
+	 * **Constantly calling this from dpp::cluster::on_voice_buffer_send
+	 * can, and will, eat a TON of cpu!**
 	 * 
 	 * @param audio_data Raw PCM audio data. Channels are interleaved,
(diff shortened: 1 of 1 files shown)


braindigitalis (approved):

include/dpp/discordvoiceclient.h
```diff
@@ -685,13 +685,13 @@ class DPP_EXPORT discord_voice_client : public websocket_client
...
-	 * Constantly calling this from the dpp::on_voice_buffer_send callback
-	 * can and will eat a TON of cpu!
+	 * **Constantly calling this from dpp::cluster::on_voice_buffer_send
+	 * can, and will, eat a TON of cpu!**
 	 * 
 	 * @param audio_data Raw PCM audio data. Channels are interleaved,
 	 * with each channel's amplitude being a 16 bit value.
 	 * 
-	 * The audio data should be 48000Hz signed 16 bit audio.
+	 * @warning **The audio data needs to be 48000Hz signed 16 bit audio, otherwise, the audio will come through incorrectly!**
```
Neko-Life, line 694: You might want to replace "incorrectly" with a more explanatory term. Having a mismatched sample rate can alter the pitch and tempo of the audio playback. Specifically, a lower sample rate can result in a higher pitch and faster tempo, while a higher sample rate can produce the opposite effect

Neko-Life (commented):
oh well just noticed it's merged already

PR #874, opened 2023-09-21
Enhancement: Reducing memory usage by removing the vtable in managed.

RealTimeChris:
Have applied the CRTP to keep the interfaces identical, as well as added snowflakes to the necessary classes, and also added a managed_base class to allow for storing them all in the same deletion_queue.

## Code change checklist

- [x] I have ensured that all methods and functions are **fully documented** using doxygen style comments


include/dpp/integration.h
```diff
@@ -80,8 +80,10 @@ struct DPP_EXPORT integration_app {
 /**
  * @brief Represents an integration on a guild, e.g. a connection to twitch.
  */
-class DPP_EXPORT integration : public managed, public json_interface<integration> {
+class DPP_EXPORT integration : public managed<integration>, public json_interface<integration> {
 public:
+	//** Unique identifier */
+	snowflake id;
```
braindigitalis, line 86: I don't really like this, it's basically added tons of duplicate lines to every object to save 8 bytes of ram per instance, I think readability is more important in this case

Mishura4:
The id could be part of managed<T>, that would help with the code duplication

RealTimeChris:
> The id could be part of managed, that would help with the code duplication

It was deliberately removed from managed in order to eliminate the vtable.

Mishura4:
> > The id could be part of managed, that would help with the code duplication
> 
> It was deliberately removed from managed in order to eliminate the vtable.

Having just the snowflake in there isn't going to produce a vtable though.

RealTimeChris:
> > > The id could be part of managed, that would help with the code duplication
> > 
> > 
> > It was deliberately removed from managed in order to eliminate the vtable.
> 
> Having just the snowflake in there isn't going to produce a vtable though.

No but having the virtual destructor that is then required to properly destroy the id will.

Mishura4:
Virtual destructor is needed if you want to destroy a variable through a pointer to its base class, so the derived object is properly destroyed. Whether the snowflake is in the base or not doesn't change that

PR #820, opened 2023-08-29
📝 Clean up the documentation

Eremiell:
* Resolves bunch of unrelated documentation issues
* Wrong capitalisation of headers and brands
* Wrong indentation of code examples
* Various typos, missing Oxford commas, bad style, or typography
* Examples didn't follow Style Guidelines in many places
* Tries to slim down on unneeded newlines
* Bunch of files have been wrongly named or wrongly placed
* Attempt to update some outdated details
* Overall there are other things the would deserve some work but that felt too ambitious for this already huge changeset

- [ ] My pull request is made against the `dev` branch.
- [ ] I have ensured that the changed library can be built on your target system. I did not introduce any platform-specific code.
- [ ] I have ensured that all methods and functions are **fully documented** using doxygen style comments.
- [ ] I tested my commits, by adding a test case to the unit tests if needed
- [ ] I have ensured that I did not break any existing API calls.
- [ ] My code follows the [coding style guide](https://dpp.dev/coding-standards.html) (if you are not sure, match the code style of existing files including indent style etc).
- [ ] I have not built my pull request using AI, a static analysis tool or similar without any human oversight. Where I have generated this pull request using a tool, I have justified why this is needed.

PR #875, opened 2023-09-21
docs: remade the linux.md page.

Jaskowicz1:
This PR remakes the `linux.md` page, basically taking the old buildcmake.md page and adding it here (but, ofc, improved), meaning the page now covers g++ and CMake. This PR also merges the changes on `linux.md` from #820.

- [x] I tested that my change works before raising the PR (via running `oxygen`, and testing examples).

### DIFF
--- a/docpages/building/linux.md
+++ b/docpages/building/linux.md
@@ -1,40 +1,179 @@
 \page buildlinux Building on Linux
 
-\note You might not need to build a copy of the library for Linux - precompiled deb files for 64 bit and 32 bit Debian and Ubuntu are provided in the GitHub version releases. Unless you are on a different Linux distribution which does not support the installation of deb files, or wish to submit fixes and enhancements to the library itself you should have an easier time installing the precompiled version instead.
+\warning **This page will cover building D++ from source.** We advise that you don't explore this route unless you wish to submit fixes/enhancements, or if you know this is right for you. If you wish to use precompiled binaries (the release version), head on over to \ref installing "this page" and find follow the tutorial that matches your operating system.
+
+This tutorial will cover two ways of using D++ from source. These two ways are:
+
+- Using D++ Source via `g++`
+- Using D++ Source via `CMake`
+
+These pages will be split by lines, you can scroll down until you find the next separating line.
+
+Let's start off with `g++`
+
+_ _ _
+
+# g++
 
 ## 1. Build Source Code
 
-    cmake -B ./build
-    cmake --build ./build -j8
+```bash
+cmake -B ./build
+cmake --build ./build -j8
+```
     
 Replace the number after `-j` with a number suitable for your setup, usually the same as the number of cores on your machine. `cmake` will fetch any dependencies that are required for you and ensure they are compiled alongside the library.
 
 ## 2. Install to /usr/local/include and /usr/local/lib
 
-    cd build; sudo make install
+```bash
+cd build
+sudo make install
+```
 
-## 3. Installation to a different directory
+## 3. Installation to a Different Directory
 
-If you want to install the library, its dependencies and header files to a different directory, specify this directory when running `cmake`:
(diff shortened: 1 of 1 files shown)


braindigitalis (changes requested):
not sure we should be merging the pages together... it means duplicated content on Google which is bad and means two places to maintain.
simply link to the cmake guide at the bottom as we did before?

Jaskowicz1:
> not sure we should be merging the pages together... it means duplicated content on Google which is bad and means two places to maintain. simply link to the cmake guide at the bottom as we did before?

They're not the same. The CMake guide no longer uses source, the new addition here shows CMake with source. Ideally, I'd like two separate pages, unless you can come up with a better idea!

braindigitalis:
> > not sure we should be merging the pages together... it means duplicated content on Google which is bad and means two places to maintain. simply link to the cmake guide at the bottom as we did before?
> 
> They're not the same. The CMake guide no longer uses source, the new addition here shows CMake with source. Ideally, I'd like two separate pages, unless you can come up with a better idea!

problem is this a guide on how to build the lib and now it's also a guide on how to build a bot... the lines are blurred and it isn't so clear any more. perhaps a separate new page for it that this one links to?

Jaskowicz1:
> > > not sure we should be merging the pages together... it means duplicated content on Google which is bad and means two places to maintain. simply link to the cmake guide at the bottom as we did before?
> > 
> > 
> > They're not the same. The CMake guide no longer uses source, the new addition here shows CMake with source. Ideally, I'd like two separate pages, unless you can come up with a better idea!
> 
> problem is this a guide on how to build the lib and now it's also a guide on how to build a bot... the lines are blurred and it isn't so clear any more. perhaps a separate new page for it that this one links to?

Yeah that's actually a good point, I didn't really think of that. So, if we're to make a separate page for building a bot using source with CMake, where would we place that? I don't think placing it next to the `buildcmake.md` page is a good idea.

braindigitalis:
> > > > not sure we should be merging the pages together... it means duplicated content on Google which is bad and means two places to maintain. simply link to the cmake guide at the bottom as we did before?
> > > 
> > > 
> > > They're not the same. The CMake guide no longer uses source, the new addition here shows CMake with source. Ideally, I'd like two separate pages, unless you can come up with a better idea!
> > 
> > 
> > problem is this a guide on how to build the lib and now it's also a guide on how to build a bot... the lines are blurred and it isn't so clear any more. perhaps a separate new page for it that this one links to?
> 
> Yeah that's actually a good point, I didn't really think of that. So, if we're to make a separate page for building a bot using source with CMake, where would we place that? I don't think placing it next to the `buildcmake.md` page is a good idea.

bury it at the bottom of this one with a hyperlink to it so that you have to be looking for it to find it, this encourages people by default to install how we can support them easier, same way we funnel them towards the windows template.

Jaskowicz1:
> > > > > not sure we should be merging the pages together... it means duplicated content on Google which is bad and means two places to maintain. simply link to the cmake guide at the bottom as we did before?
> > > > 
> > > > 
> > > > They're not the same. The CMake guide no longer uses source, the new addition here shows CMake with source. Ideally, I'd like two separate pages, unless you can come up with a better idea!
> > > 
> > > 
> > > problem is this a guide on how to build the lib and now it's also a guide on how to build a bot... the lines are blurred and it isn't so clear any more. perhaps a separate new page for it that this one links to?
> > 
> > 
> > Yeah that's actually a good point, I didn't really think of that. So, if we're to make a separate page for building a bot using source with CMake, where would we place that? I don't think placing it next to the `buildcmake.md` page is a good idea.
> 
> bury it at the bottom of this one with a hyperlink to it so that you have to be looking for it to find it, this encourages people by default to install how we can support them easier, same way we funnel them towards the windows template.

I like that idea haha. However, if it's not listed as a "subpage" anywhere, it'll become a top page (like FAQ) so it needs to be under some sub-section. I think we should also have a discussion about the linux template, if you're available for that :)

PR #876, opened 2023-09-22
Enhancement: Improving memory efficiency by eliminating the vtable in the managed class.

RealTimeChris:
By making the destructor protected - we can remove the possibility that someone tries to delete one of the derived classes by deleting an instance of the managed class, which should allow for not having a virtual destructor.

## Code change checklist
- [x] I have ensured that all methods and functions are **fully documented** using doxygen style comments

### DIFF
--- a/include/dpp/cache.h
+++ b/include/dpp/cache.h
@@ -21,20 +21,116 @@
  ************************************************************************************/
 
 #pragma once
-#include <dpp/export.h>
 #include <dpp/snowflake.h>
+#include <dpp/exception.h>
+#include <dpp/channel.h>
 #include <dpp/managed.h>
 #include <unordered_map>
+#include <dpp/export.h>
+#include <dpp/emoji.h>
+#include <dpp/guild.h>
+#include <dpp/user.h>
+#include <dpp/role.h>
+#include <typeindex>
 #include <mutex>
-#include <shared_mutex>
 
 namespace dpp {
 
+class cache_registry;
+
+extern DPP_EXPORT std::unordered_map<std::type_index, cache_registry*> regs;
 extern DPP_EXPORT std::unordered_map<managed*, time_t> deletion_queue;
 extern DPP_EXPORT std::mutex deletion_mutex;
 
+template<typename T> class cache;
+
+/**
+ * @brief The `cache_registry` class defines an abstract base class for cache objects.
+ *
+ * It provides a mechanism for registering, managing, and performing garbage collection
+ * on cache objects. Cache objects must inherit from this class and implement the
+ * required functions.
+ */
+class cache_registry {
+public:
+	/**
+	 * @brief Pure virtual function for performing garbage collection.
+	 */
+	virtual void garbage_collection() = 0;
+
+	/**
+	 * @brief Pure virtual function to get the type index for a cache.
+	 * @return Type index representing the cache's type.
+	 */
+	virtual std::type_index get_type_index() const = 0;
+
+	/**
+	 * @brief Static method to register a cache object.
+	 * @param registry A pointer to the cache_registry object to register.
+	 */
+	static void register_cache(cache_registry* registry) {
+		auto type_index = registry->get_type_index();
+		if (registries().find(type_index) != registries().end()) {
+			throw cache_exception{ "Sorry, but there was already a cache by that type that has been registered." };
+		}
+		registries()[registry->get_type_index()] = registry;
+	}
+
+	/**
+	 * @brief Static method to call garbageCollection for all registered caches.
+	 * This method invokes garbage collection on all registered cache objects.
+	 */
+	static void call_garbage_collection() {
+		for (auto& [key, value] : registries()) {
+			value->garbage_collection();
+		}
+	}
+
+	/**
+	 * @brief Static method to retrieve a cache object by type.
+	 *
+	 * @tparam T The type of the cache to retrieve.
+	 * @return A pointer to the cache object of type T if found, or nullptr if not found.
+	 */
+	template <typename T>
+	static cache<T>* get_cache() {
+		std::type_index typeIndex(typeid(T));
+		for (auto& [key,value]: registries()) {
+			if (value->get_type_index() == typeIndex) {
+				return static_cast<cache<T>*>(value);
+			}
+		}
+		// If cache of type T not found, create and register a new one, then return it.
+		register_cache(new cache<T>{});
+		return get_cache<T>();
+	}
+
+private:
+	/**
+	 * @brief Static vector to store registered cache objects.
+	 * This vector holds pointers to all registered cache objects.
+	 */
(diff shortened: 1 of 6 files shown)


braindigitalis:
why are you so obsessed with getting rid of the vtable here?
I already said it's over engineered and hard to read... theres no need to go this far to save 8 bytes.

RealTimeChris:
> why are you so obsessed with getting rid of the vtable here? I already said it's over engineered and hard to read... theres no need to go this far to save 8 bytes.

Because I tend to fixate on things periodically - and I disagree that it's over engineered and hard to read - I've barely changed anything. Is it slightly more complex? Yes. Is it worth the memory savings? I would say it is - though I suppose that's what is up for debate here. The final say is ultimately yours though.

include/dpp/cache.h
```diff
@@ -269,5 +264,53 @@ cache_decl(role, find_role, get_role_cache, get_role_count);
...
 cache_decl(emoji, find_emoji, get_emoji_cache, get_emoji_count);
 
+/** Run garbage collection across all caches removing deleted items
+ * that have been deleted over 60 seconds ago.
+ */
+ /* Because other threads and systems may run for a short while after an event is received, we don't immediately
+  * delete pointers when objects are replaced. We put them into a queue, and periodically delete pointers in the
+  * queue. This also rehashes unordered_maps to ensure they free their memory.
+  */
+  /* Because other threads and systems may run for a short while after an event is received, we don't immediately
```
Jaskowicz1, line 274: You've wrote this comment twice.

src/unittest/test.cpp
```diff
@@ -1349,7 +1350,7 @@ Markdown lol \\|\\|spoiler\\|\\| \\~\\~strikethrough\\~\\~ \\`small \\*code\\* b
 
 		bool message_edit_tested = false;
 		bot.on_message_update([&](const dpp::message_update_t &event) {
-			if (event.msg.author == bot.me.id) {
+			if (event.msg.author.id == bot.me.id) {
```
Jaskowicz1, line 1353: What's the difference here? The parent class of author handles this right?

include/dpp/cache.h
```diff
@@ -269,5 +264,53 @@ cache_decl(role, find_role, get_role_cache, get_role_count);
...
+   * delete pointers when objects are replaced. We put them into a queue, and periodically delete pointers in the
+   * queue. This also rehashes unordered_maps to ensure they free their memory.
+   */
+template<typename value_type> inline void garbage_collection() {
+	time_t now = time(nullptr);
```
Jaskowicz1, line 279: If we're really crunching for memory, would something like:
```c++
time_t now;
time(&now);
```
not be better? [See godbolt here](https://godbolt.org/z/E9KEhETx9).

I'm still pretty new to all of this so I could be wrong!

include/dpp/managed.h
```diff
@@ -72,6 +68,12 @@ namespace dpp {
 		 * @return false objects are the same id
 		 */
 		bool operator!=(const managed& other) const noexcept;
+
+	protected:
+		/**
+		 * @brief Destroy the managed object
+		 */
+		~managed() = default;
```
Jaskowicz1, line 76: This now breaks the way that derived classes deconstruct now, no? Like, isn't this meant to be polymorphic?

src/unittest/test.cpp
RealTimeChris: It stops an implicit conversion to managed, which is now basically an abstract base class.

include/dpp/managed.h
RealTimeChris: Well, by having it as virtual - it creates a vtable which adds 8bytes of ram to every instance of it and its derived classes. But, by not setting it to virtual, if we attempt to delete a derived class via this base classes' destructor - it won't destroy the derived class. By setting this to non-virtual but also protected, we eliminate the risk that someone will attempt to destroy any of these classes using this base classes' destructor.

include/dpp/cache.h
RealTimeChris: Thanks!

Jaskowicz1: No problem :)

src/unittest/test.cpp
Jaskowicz1: Ah right, I get you!

include/dpp/cache.h
Jaskowicz1: > I'm not sure - why?

Less instructions, less of the `QWORD PTR [rbp-8]`, I'm not well versed in assembly (actually, never done assembly lol) but from a bit of googling, the suggestion seems to be better but only if you're really optimising (so modern architecture will probably not see this change performance whatsoever).

include/dpp/cache.h
```diff
@@ -269,5 +264,49 @@ cache_decl(role, find_role, get_role_cache, get_role_count);
...
+					repeat = true;
+					break;
+				}
+			}
+		} while (repeat);
+		if (deletion_queue.size() == 0) {
+			deletion_queue = {};
+		}
+	}
+	if constexpr (std::is_same_v<emoji, value_type>) {
```
Mishura4, line 294: Why this btw, and making this function a template? Can't this function rehash every cache like how it was before?

RealTimeChris: So we can static cast the managed pointers to their derived types and avoid falling into not deleting them properly, given that we've made the managed destructor protected.

Mishura4: But for example dpp::get_emoji_cache() is a `dpp::cache<emoji>`, right? So when doing `dpp::get_emoji_cache()->rehash();` like how it was before, that would already call the function with the correct type, as `dpp::cache<emoji>::rehash` uses `emoji` anyways (see https://github.com/brainboxdotcc/DPP/blob/master/include/dpp/cache.h#L233)

RealTimeChris: Right but when we call garbage_collection repeatedly - one for each of the types, we don't want to rehash it for each of the calls.

Mishura4: Well you wouldn't need to call garbage_collection multiple times, just one like how it was before

RealTimeChris: No because we need access to the derived type to static_cast the managed pointer into, as seen here:
https://github.com/brainboxdotcc/DPP/blob/03d126bbe5e431c249639018e603c0c1ea147783/include/dpp/cache.h#L283

Mishura4: I see. Another thing however, how are we certain `deletion_queue` only contains objects of that type? The way i see it in https://github.com/brainboxdotcc/DPP/blob/03d126bbe5e431c249639018e603c0c1ea147783/include/dpp/cache.h#L137, any cache can add to it

RealTimeChris: There I've solved the "cache of any types" issue. It might be too over engineered, let me know what you think.

Issue #714, opened 2023-06-18
Bot is unable to properly connect to certain Discord voice servers 


KontraCity:
**Git commit reference**
cd5b9cab3fb6a49ece575e0f86475268f886766b

**Describe the bug**
The problem is that `on_voice_ready` callback is not called when bot connects to certain voice channels.
Example of successful voice server connection:
1. DPP logs show this:
`DEBUG: Sending op 4 to join VC, guild xxx channel xxx`
`DEBUG: Connecting voice for guild xxx channel xxx`
`DEBUG: Connecting new voice session...`
`DEBUG: Voice websocket established; UDP endpoint: xxx.xxx.xxx.xxx:xxx [ssrc=xxx] with 7 modes`
`DEBUG: External IP address: xxx.xxx.xxx.xxx`
2. Bot connects with following request:
`GET /?v=4 HTTP/1.1`
`Host: russia8508.discord.media`
`pragma: no-cache`
`User-Agent: DPP/0.1`
`Upgrade: WebSocket`
`Connection: Upgrade`
`Sec-WebSocket-Key: xxx`
`Sec-WebSocket-Version: 13`
3. Server responds with HTTP 101 response:
`HTTP/1.1 101 Switching Protocols`
`Connection: upgrade`
`sec-websocket-accept: xxx`
`upgrade: websocket`
`CF-Cache-Status: DYNAMIC`
`Report-To: {"endpoints":[{"url":"https:\/\/a.nel.cloudflare.com\/report\/v3?s=xxx"}],"group":"cf-nel","max_age":604800}`
`NEL: {"success_fraction":0.01,"report_to":"cf-nel","max_age":604800}`
`Strict-Transport-Security: max-age=15552000; includeSubDomains; preload`
`X-Content-Type-Options: nosniff`
`Server: cloudflare`
`CF-RAY: xxx-DME`
4. Server sends JSON payload:
`1{"op":8,"d":{"v":4,"heartbeat_interval":13750.0}}`
`Connection: upgrade`
`sec-websocket-accept: xxx`
`upgrade: websocket`
`CF-Cache-Status: DYNAMIC`
`Report-To: {"endpoints":[{"url":"https:\/\/a.nel.cloudflare.com\/report\/v3?s=xxx"}],"group":"cf-nel","max_age":604800}`
`NEL: {"success_fraction":0.01,"report_to":"cf-nel","max_age":604800}`
[...]

PR #883, opened 2023-09-23
refactor: changed user-agent, more logging, removed infinite loop

Jaskowicz1:
This PR changes the user-agent in wsclient (to match versions.h, sorta), adds more logging in sslclient.cpp, and removes an infinite loop in `discord_voice_client::thread_run`.

The logging in sslclient.cpp will now send a warning when we encounter a HTTP code that isn't 204, 101, or 200. This means people can see what's going on and report it, without modifying D++ which is very helpful the more we move towards getting people to use pre-compiled binaries (view the image above the checklist for a preview).

The infinite loop in `discord_voice_client::thread_run` was caused by the error code 400 (you can view #714 for more info). Whilst we haven't fix the issue that #714 is about, we now tell users what's going on if things start to break (as mentioned above). The infinite loop simply checks how many times the `do while` loop has ran, if it's ran 5 times (without a 3 second break inbetween a loop) then we break the loop. This will mean we also don't spam the console too much with the above addition.

image

## Code change checklist


src/dpp/discordvoiceclient.cpp
```diff
@@ -397,11 +397,34 @@ bool discord_voice_client::is_playing() {
...
 		bool error = false;
 		ssl_client::read_loop();
 		ssl_client::close();
+
+		time_t current_time = time(nullptr);
+		/* Here, we check if it's been longer than 3 seconds since the previous loop,
+		 * this gives us time to see if it's an actual disconnect, or an error.
+		 * This will prevent us from looping too much, meaning error codes do not cause an infinite loop.
+		 */
+		if(current_time - time_since_loop >= 3) {
```
Mishura4, line 414: `time_since_loop` is uninitialized here the first time this loop is executed, this is UB

src/dpp/wsclient.cpp
```diff
@@ -52,7 +52,7 @@ void websocket_client::connect()
 		"GET " + this->path + " HTTP/1.1\r\n"
 		"Host: " + this->hostname + "\r\n"
 		"pragma: no-cache\r\n"
-		"User-Agent: DPP/0.1\r\n"
+		"User-Agent: " + "DiscordBot (https://github.com/brainboxdotcc/DPP, 16.0.38)" + "\r\n"
```
Mishura4, line 55: Why this concatenation here?

src/dpp/sslclient.cpp
```diff
@@ -504,6 +505,14 @@ void ssl_client::read_loop()
 							case SSL_ERROR_NONE:
 								/* Data received, add it to the buffer */
 								if (r > 0) {
+									std::string data(server_to_client_buffer, r);
+									std::smatch matches;
+
+									if(std::regex_search(data, matches, std::regex(R"(^(HTTP\/1\.1 .+))"))) {
```
Mishura4, line 511: won't this regex match the whole document? what if it's a discord message with "200" in the message content?

Mishura4 (dismissed):

src/dpp/wsclient.cpp
Jaskowicz1: Placeholder, so it can be replaced later with the actual version detection stuff.

src/dpp/discordvoiceclient.cpp
Jaskowicz1: Whoops, old habits 🙃

src/dpp/sslclient.cpp
Jaskowicz1: No, it shouldn't. Granted, I used it from #714 but I'll double check this tomorrow.

src/dpp/discordvoiceclient.cpp
```diff
@@ -397,11 +397,34 @@ bool discord_voice_client::is_playing() {
...
+		time_t current_time = time(nullptr);
+		/* Here, we check if it's been longer than 3 seconds since the previous loop,
+		 * this gives us time to see if it's an actual disconnect, or an error.
+		 * This will prevent us from looping too much, meaning error codes do not cause an infinite loop.
+		 */
+		if(current_time - last_loop_time >= 3) {
+			times_looped = 0;
+		}
+		/* This does mean we'll always have times_looped at a minimum of 1, this is intended. */
+		times_looped += 1;
```
braindigitalis, line 418: `times_looped++` is perhaps cleaner

src/dpp/sslclient.cpp
braindigitalis: dont use std::regex. It is really really slow, like, unbearably so. You should just use string splitting instead using basic std::string methods.

src/dpp/wsclient.cpp
braindigitalis: please use the actual version number, you should fix the broken version string in queues.cpp, the static one, and expose it as non-static const or constexpr in the dpp namespace, you can then use it here.

The reason the version string there is broken is that the define values contain the version content as hex. You'll need to convert them as hex into text, so 0x10 is version 10:

```cpp
static std::string http_version = "DiscordBot (https://github.com/brainboxdotcc/DPP, " + std::to_string(DPP_VERSION_MAJOR) + "." + std::to_string(DPP_VERSION_MINOR) + "." + std::to_string(DPP_VERSION_PATCH) + ")";
```

braindigitalis (changes requested):

src/dpp/utility.cpp
```diff
@@ -29,7 +29,6 @@
 #include <thread>
 #include <functional>
 #include <chrono>
-#include <ctime>
```
braindigitalis, line 32: why was ctime removed? dont we use this?

src/unittest/test.cpp
```diff
@@ -169,10 +169,16 @@ Markdown lol \\|\\|spoiler\\|\\| \\~\\~strikethrough\\~\\~ \\`small \\*code\\* b
 	set_test(READFILE, false);
 	std::string rf_test = dpp::utility::read_file(SHARED_OBJECT);
 	FILE* fp = fopen(SHARED_OBJECT, "rb");
-	fseek(fp, 0, SEEK_END);
-	size_t off = (size_t)ftell(fp);
-	fclose(fp);
-	set_test(READFILE, off == rf_test.length());
+	if (!fp) {
```
braindigitalis, line 172: this does not need to be changed - it expects that the file is in place and that you run the unit test from the correct location. e.g. `./unittest` not `./build/unittest` or similar

src/dpp/utility.cpp
Jaskowicz1: Not sure but it compiles

Neko-Life: it was double include

src/unittest/test.cpp
Neko-Life: @Jaskowicz1 can revert commit 7ef50c0bdf3d0f38798b220e00c8255f7be4166d

src/dpp/sslclient.cpp
```diff
@@ -504,6 +504,15 @@ void ssl_client::read_loop()
 							case SSL_ERROR_NONE:
 								/* Data received, add it to the buffer */
 								if (r > 0) {
+									const std::string data(server_to_client_buffer, r);
+									const std::vector<std::string>& data_lines = utility::tokenize(data);
+									const std::string& http_reponse(data_lines[0]);
```
braindigitalis, line 509: don't use a reference, it is not valid after the function returns

src/dpp/sslclient.cpp
```diff
@@ -504,6 +504,15 @@ void ssl_client::read_loop()
...
+
+									/* Does the first line begin with a http code and does it not contain either 204, 101 or 200? */
+									if(http_reponse.rfind("HTTP/1.1", 0) != std::string::npos && http_reponse.find("204") == std::string::npos && http_reponse.find("101") == std::string::npos && http_reponse.find("200") == std::string::npos) {
+										log(ll_warning, "Received unhandled code: " + http_reponse);
```
braindigitalis, line 513: you could just use tokenize again to split by space. this code will fail if anywhere else in the first line has these numbers e.g.

`HTTP/1.1 403 The maximum number of users (200) is exceeded`

convoluted example but you get what I mean yes?

braindigitalis (changes requested):

src/dpp/sslclient.cpp
Jaskowicz1: I get what you mean yeah :)

src/dpp/sslclient.cpp
Jaskowicz1: For both?

braindigitalis: yes just 
`const std::vector<std::string> data_lines =`
is fine

src/dpp/sslclient.cpp
```diff
@@ -505,12 +505,18 @@ void ssl_client::read_loop()
...
 									const std::string& http_reponse(data_lines[0]);
 
-									/* Does the first line begin with a http code and does it not contain either 204, 101 or 200? */
-									if(http_reponse.rfind("HTTP/1.1", 0) != std::string::npos && http_reponse.find("204") == std::string::npos && http_reponse.find("101") == std::string::npos && http_reponse.find("200") == std::string::npos) {
-										log(ll_warning, "Received unhandled code: " + http_reponse);
+									/* Does the first line begin with a http code? */
+									if(http_reponse.rfind("HTTP/1.1", 0) != std::string::npos) {
+										/* Now let's split the first line by every space, meaning we can check the actual HTTP code. */
+										const std::vector<std::string> line_split_by_space = utility::tokenize(data_lines[0], " ");
+										const std::string& http_code(line_split_by_space[1]);
```
braindigitalis, line 515: you need to check there are at least 3 elements

src/dpp/sslclient.cpp
```diff
@@ -505,12 +505,18 @@ void ssl_client::read_loop()
...
+
+										/* Does the http code not contain either 204, 101 or 200? If it doesn't, log it. */
+										if(http_code.find("204") == std::string::npos && http_code.find("101") == std::string::npos && http_code.find("200") == std::string::npos)
```
braindigitalis, line 518: you can just convert http code to int and then use == to check the code no need for .find()

braindigitalis (changes requested):

src/dpp/sslclient.cpp
Jaskowicz1: Ah yeah hahaha, whoops

src/dpp/sslclient.cpp
braindigitalis: you got a & ref here again too that isnt needed

braindigitalis (approved):

PR #884, opened 2023-09-24
Enhancement: Making AudioMixer a member instead of invoking it statically to cut down CPU usage by roughly 50%

RealTimeChris:
Storing it along with the array of floats that are used as an intermediary for converting to/from avx registers seems to cut down CPU usage while streaming audio in my implementation by roughly 50%. As opposed to reallocating the array for every function call.

## Code change checklist

### DIFF
--- a/include/dpp/discordvoiceclient.h
+++ b/include/dpp/discordvoiceclient.h
@@ -55,6 +55,8 @@ struct OpusRepacketizer;
 
 namespace dpp {
 
+class audio_mixer;
+
 // !TODO: change these to constexpr and rename every occurrence across the codebase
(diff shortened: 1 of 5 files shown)


include/dpp/discordvoiceclient.h
```diff
@@ -38,6 +38,7 @@
 #include <dpp/dispatcher.h>
 #include <dpp/cluster.h>
 #include <dpp/discordevents.h>
+#include <dpp/isa_detection.h>
```
braindigitalis, line 41: we cannot include this here.
including this here means that the end user code must set AVX_TYPE. it cannot be included in a header.

braindigitalis (changes requested):

braindigitalis (approved):

PR #896, opened 2023-09-27
docs: changed the notes on read_file.

Jaskowicz1:
This PR alters the notes on `dpp::utility::read_file`. After experimenting in DPP-UE, I realised that there was no note as to what happens if the file is big and causes the event to timeout, causing a crash in DPP-UE and an error log in DPP. I made the note to give more information to users about this, making sure they know to use `thinking` if the file is big.

The PR also renames the previous note to be a warning. I feel that the warning is very much required so people know that the function can, and will, block the thread, if the file is big.

### DIFF
--- a/include/dpp/utility.h
+++ b/include/dpp/utility.h
@@ -448,8 +448,8 @@ namespace dpp {
 
 		/**
 		 * @brief Read a whole file into a std::string.
-		 * Be sure you have enough memory to read the file, if you are reading a large file.
-		 * @note Be aware this function can block! If you are regularly reading large files, consider caching them.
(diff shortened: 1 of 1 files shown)


Henonicks:
You should also add the note to docs themselves. You are adding documentation anyways.

Jaskowicz1:
> You should also add the note to docs themselves. You are adding documentation anyways.

? This is the docs themselves?

Henonicks:
> ? This is the docs themselves?

You've only edited the .h file but the dpp.dev docs will stay the same so the code documentation won't match.

Henonicks:
Mbad, didn't know that

braindigitalis (approved):

PR #902, opened 2023-09-30
move the error checks from sslclient.cpp to wsclient.cpp where they belong, so they dont cause excessive data copies or false alarms

braindigitalis:
The code added in a previous PR would cause unceccessary checks of the data every time we did SSL_read, this really belongs in wsclient when looking for the 101 reply, so i moved it.

## Code change checklist

### DIFF
--- a/src/dpp/wsclient.cpp
+++ b/src/dpp/wsclient.cpp
@@ -159,7 +159,7 @@ bool websocket_client::handle_buffer(std::string &buffer)
 					} else if (status.size() < 3) {
 						log(ll_warning, "Malformed HTTP response on websocket");
 						return false;
-					} else {
+					} else if (status[1] != "200" && status[1] != "204") {
 						log(ll_warning, "Received unhandled code: " + status[1]);
 						return false;
 					}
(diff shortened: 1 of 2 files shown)


src/dpp/wsclient.cpp
```diff
@@ -156,7 +156,11 @@ bool websocket_client::handle_buffer(std::string &buffer)
 						}
 		
 						state = CONNECTED;
-					} else {
+					} else if (status.size() < 3) {
```
Jaskowicz1, line 159: Could this not go at the top? Meaning the if statement above could just be `if(status[1 == "101")`? with a check above that being 
```c++
if (status.size() < 3) {
    log(ll_warning, "Malformed HTTP response on websocket");
    return false;
}
````

braindigitalis: to be honest we should never see this. it is the least likely branch to occur not most likely, unless discord websocket endpoints start speaking complete nonsense...

RealTimeChris (approved):

Mishura4 (approved):

PR #909, opened 2023-10-02
breaking, fix: guild_member state tracking

braindigitalis:
- [x] breaking, fix: guild_member state tracking.
    You must now use add_role, remove_role or set_roles, also set_nickname on a guild_member for edit_guild_mebmer to have any effect on that value. This fixes longstanding issue where more permissions are required than should be, to edit guild members
- [x] test: move mutltiheader cookie test to extended, impossible to test in uk and eu due to GDPR

## Code change checklist

### DIFF
--- a/include/dpp/guild.h
+++ b/include/dpp/guild.h
@@ -412,7 +412,7 @@ public:
 	/**
 	 * @brief Remove a role from this member
 	 * @note This call sets the role change bit, which causes the new role
-	 * list to be sent if this is passed to dpp::clusterguild_edit_member
+	 * list to be sent if this is passed to dpp::cluster::guild_edit_member
 	 * or dpp::cluster::guild_add_member
 	 * 
 	 * @param role_id Role ID to remove
@@ -423,7 +423,7 @@ public:
 	/**
 	 * @brief Set a new role list for this member
 	 * @note This call sets the role change bit, which causes the new role
-	 * list to be sent if this is passed to dpp::clusterguild_edit_member
+	 * list to be sent if this is passed to dpp::cluster::guild_edit_member
 	 * or dpp::cluster::guild_add_member
 	 * 
 	 * @param role_ids Roles to set
--- a/include/dpp/guild.h
+++ b/include/dpp/guild.h
@@ -196,6 +196,8 @@ protected:
 	std::vector<snowflake> roles;
 	/** A set of flags built from the bitmask defined by dpp::guild_member_flags */
 	uint16_t flags;
+
+	friend void from_json(const nlohmann::json& j, guild_member& gm);
 public:
 	/** Guild id */
 	snowflake guild_id;
@@ -337,7 +339,7 @@ public:
 	 * 
 	 * @return std::vector<dpp::snowflake> roles
 	 */
-	std::vector<dpp::snowflake> get_roles();
+	std::vector<dpp::snowflake> get_roles() const;
 
 	/**
 	 * @brief Find the dpp::user object for this member. This is an alias for dpp::find_user
--- a/include/dpp/guild.h
+++ b/include/dpp/guild.h
@@ -401,7 +401,7 @@ public:
 	/**
 	 * @brief Add a role to this member
 	 * @note This call sets the role change bit, which causes the new role
-	 * list to be sent if this is passed to dpp::clusterguild_edit_member
+	 * list to be sent if this is passed to dpp::cluster::guild_edit_member
 	 * or dpp::cluster::guild_add_member
 	 * 
 	 * @param role_id Role ID to add
--- a/include/dpp/guild.h
+++ b/include/dpp/guild.h
@@ -332,7 +332,7 @@ public:
 	 * 
 	 * @return std::string nickname
 	 */
-	std::string get_nickname();
+	std::string get_nickname() const;
 
 	/**
 	 * @brief Get the roles
--- a/include/dpp/guild.h
+++ b/include/dpp/guild.h
@@ -189,11 +189,14 @@ enum guild_member_flags : uint16_t {
  * This contains the user's nickname, guild roles, and any other guild-specific flags.
  */
 class DPP_EXPORT guild_member {
-public:
+protected:
 	/** Nickname, or empty string if they don't have a nickname on this guild */
 	std::string nickname;
 	/** List of roles this user has on this guild */
 	std::vector<snowflake> roles;
+	/** A set of flags built from the bitmask defined by dpp::guild_member_flags */
+	uint16_t flags;
+public:
 	/** Guild id */
 	snowflake guild_id;
 	/** User id */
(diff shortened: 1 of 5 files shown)


src/dpp/guild.cpp
```diff
@@ -112,6 +112,27 @@ std::string guild_member::get_mention() const {
...
+}
+
+guild_member& guild_member::add_role(dpp::snowflake role_id) {
+	roles.emplace_back(role_id);
+	flags |= gm_roles_action;
+	return *this;
+}
+
+guild_member& guild_member::remove_role(dpp::snowflake role_id) {
+	roles.erase(std::remove_if(roles.begin(), roles.end(), [&role_id](dpp::snowflake role) {
```
Mishura4, line 126: Can just use `std::remove` here instead of using `std::remove_if` with a lambda

src/dpp/guild.cpp
```diff
@@ -211,19 +232,23 @@ std::string guild_member::build_json(bool with_id) const {
...
 	}
 
-	if (!this->roles.empty()) {
-		j["roles"] = {};
-		for (auto & role : this->roles) {
-			j["roles"].push_back(std::to_string(role));
+	if (this->flags & gm_roles_action) {
+		if (!this->roles.empty()) {
+			j["roles"] = {};
+			for (auto & role : this->roles) {
```
Mishura4, line 246: `const auto&`!

src/dpp/guild.cpp
```diff
@@ -211,19 +232,23 @@ std::string guild_member::build_json(bool with_id) const {
...
+			j["nick"] = json::value_t::null;
+		}
 	}
 
-	if (!this->roles.empty()) {
-		j["roles"] = {};
-		for (auto & role : this->roles) {
-			j["roles"].push_back(std::to_string(role));
+	if (this->flags & gm_roles_action) {
+		if (!this->roles.empty()) {
```
Mishura4, line 244: This whole branch is redundant, the `for` loop will be a no-op anyways if it's empty

include/dpp/guild.h
```diff
@@ -378,6 +380,40 @@ class DPP_EXPORT guild_member {
...
+	 * 
+	 * @param role_id Role ID to remove
+	 * @return guild_member& Reference to self
+	 */
+	guild_member& remove_role(dpp::snowflake role_id);
+
+	/**
+	 * @brief Set a new role list for this member
+	 * @note This call sets the role change bit, which causes the new role
+	 * list to be sent if this is passed to dpp::clusterguild_edit_member
```
Mishura4, line 409: Typo in the documentation: `dpp::clusterguild_edit_member`

include/dpp/guild.h
```diff
@@ -378,6 +380,40 @@ class DPP_EXPORT guild_member {
...
+	 * 
+	 * @param role_id Role ID to add
+	 * @return guild_member& Reference to self
+	 */
+	guild_member& add_role(dpp::snowflake role_id);
+
+	/**
+	 * @brief Remove a role from this member
+	 * @note This call sets the role change bit, which causes the new role
+	 * list to be sent if this is passed to dpp::clusterguild_edit_member
```
Mishura4, line 398: Typo in the documentation: `dpp::clusterguild_edit_member`

Mishura4 (changes requested):
Request changes + making the fields private as we discussed on Discord, and good for me

include/dpp/guild.h
```diff
@@ -180,6 +180,8 @@ enum guild_member_flags : uint16_t {
 	gm_bypasses_verification = 0b0000000010000000,
 	/** Member has started onboarding */
 	gm_started_onboarding = 0b0000000100000000,
+	gm_roles_action    = 0b0000001000000000,
```
Mishura4, line 183: Will this be sent to Discord? Will that cause issues?

braindigitalis: no, we serialise these into guild member data ourselves, our bitfields don't match discord's definitions

src/dpp/guild.cpp
braindigitalis: did you mean std::erase, seems to be C++20 only?

https://en.cppreference.com/w/cpp/container/vector/erase2

src/dpp/guild.cpp
Mishura4: Huh
I am talking about this->roles yes
You check if it's empty then you loop over it to push_back roles in it, the loop is also a check that it's empty

src/dpp/role.cpp
```diff
@@ -413,8 +413,9 @@ members_container role::get_members() const {
 		}
 		for (auto & m : g->members) {
 			/* Iterate all members and use std::find on their role list to see who has this role */
-			auto i = std::find(m.second.roles.begin(), m.second.roles.end(), this->id);
-			if (i != m.second.roles.end()) {
+			auto r = m.second.get_roles();
```
Mishura4, line 416: Possibly an unintended copy here : you probably want `const auto&`

src/dpp/guild.cpp
```diff
@@ -211,19 +239,19 @@ std::string guild_member::build_json(bool with_id) const {
...
+		} else {
+			j["nick"] = json::value_t::null;
+		}
 	}
 
-	if (!this->roles.empty()) {
-		j["roles"] = {};
-		for (auto & role : this->roles) {
+	if (gm_roles_action) {
+    		j["roles"] = {};
```
Mishura4, line 251: Space indent :omg:

src/dpp/guild.cpp
```diff
@@ -123,9 +123,7 @@ guild_member& guild_member::add_role(dpp::snowflake role_id) {
 }
 
 guild_member& guild_member::remove_role(dpp::snowflake role_id) {
-	roles.erase(std::remove_if(roles.begin(), roles.end(), [&role_id](dpp::snowflake role) {
-		return role == role_id;
-	}), roles.end());
+	std::remove(roles.begin(), roles.end(), role_id);
```
Mishura4, line 126: Should be `roles.erase(std::remove(roles.begin(), roles.end(), role_id));`

src/dpp/guild.cpp
```diff
@@ -112,6 +112,34 @@ std::string guild_member::get_mention() const {
...
+	return *this;
+}
+
+guild_member& guild_member::remove_role(dpp::snowflake role_id) {
+	roles.erase(std::remove(roles.begin(), roles.end(), role_id), roles.end());
+	flags |= gm_roles_action;
+	return *this;
+}
+
+std::string guild_member::get_nickname() {
```
Mishura4, line 131: Shouldn't this return a const reference? also, shouldn't the method be const-qualified?

src/dpp/guild.cpp
```diff
@@ -112,6 +112,34 @@ std::string guild_member::get_mention() const {
...
+	return nickname;
+}
+
+std::vector<dpp::snowflake> guild_member::get_roles() const {
```
Mishura4, line 135: Shouldn't this return a const reference?

Mishura4 (changes requested):

Jaskowicz1 (approved):
Bellissimo,

Mishura4 (approved):
Toad doing a thumbs up motion

PR #913, opened 2023-10-03
ci: changes to labeler, stopped labeler from changing every push.

Jaskowicz1:
This PR adds the code tag to labeler. I have manually created the "code" label, labeler now detects any changes for "src/" and "include/" and adds the "code" tag to them".

This PR also limits the labeler bot to only run on an opened and reopened PR, meaning you can remove/add labels and the bot won't re-run.


Jaskowicz1:
Changed the title of this PR as it now does more than what was originally stated.

PR #924, opened 2023-10-05
feat: added support for hsl colors

ruslan-ilesik:
feat: added support for hsl colors to be used in embed

## Code change checklist

### DIFF
--- a/include/dpp/utility.h
+++ b/include/dpp/utility.h
@@ -424,8 +424,8 @@ namespace dpp {
 		 * @brief Convert doubles to HSL for sending in embeds
 		 *
 		 * @param h hue value, between 0 and 1 inclusive
-		 * @param s saturation value in percentage , between 0 and 1 inclusive
-		 * @param l lightness value in percentage , between 0 and 1 inclusive
+		 * @param s saturation value in percentage, between 0 and 1 inclusive
+		 * @param l lightness value in percentage, between 0 and 1 inclusive
 		 * @return uint32_t returned integer colour value
 		 */
 		uint32_t DPP_EXPORT hsl(double h, double s, double l);
@@ -434,8 +434,8 @@ namespace dpp {
 		 * @brief Convert ints to HSL for sending in embeds
 		 *
 		 * @param h hue value, between 0 and 360 inclusive
-		 * @param s saturation value in percentage , between 0 and 100 inclusive
-		 * @param l lightness value in percentage , between 0 and 100 inclusive
+		 * @param s saturation value in percentage, between 0 and 100 inclusive
+		 * @param l lightness value in percentage, between 0 and 100 inclusive
 		 * @return uint32_t returned integer colour value
 		 */
 		uint32_t DPP_EXPORT hsl(int h, int s, int l);
--- a/include/dpp/utility.h
+++ b/include/dpp/utility.h
@@ -425,7 +425,7 @@ namespace dpp {
 		 *
 		 * @param h hue value, between 0 and 1 inclusive
 		 * @param s saturation value in procents, between 0 and 1 inclusive
-		 * @param l ligthness value in procents, between 0 and 1 inclusive
+		 * @param l lightness value in procents, between 0 and 1 inclusive
 		 * @return uint32_t returned integer colour value
 		 */
 		uint32_t DPP_EXPORT hsl(double h, double s, double l);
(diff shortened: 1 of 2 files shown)


include/dpp/utility.h
```diff
@@ -420,6 +420,26 @@ namespace dpp {
 		 */
 		uint32_t DPP_EXPORT cmyk(int c, int m, int y, int k);
 
+		/**
+		 * @brief Convert doubles to HSL for sending in embeds
+		 *
+		 * @param h hue value, between 0 and 1 inclusive
+		 * @param s saturation value in procents, between 0 and 1 inclusive
```
Jaskowicz1, line 427: "procents"? Do you mean percent? Punctuation should also be here, with a capital letter and a full stop.

include/dpp/utility.h
```diff
@@ -420,6 +420,26 @@ namespace dpp {
...
+		 * @param l lightness value in procents, between 0 and 1 inclusive
+		 * @return uint32_t returned integer colour value
+		 */
+		uint32_t DPP_EXPORT hsl(double h, double s, double l);
+
+		/**
+		 * @brief Convert ints to HSL for sending in embeds
+		 *
+		 * @param h hue value, between 0 and 356 inclusive
```
Jaskowicz1, line 436: Shouldn't this be "between 0 and 360"?

braindigitalis (approved):
looks nicer now. as a future change id like to make the 1/3 etc constexpr values (in a new pr)
it is likely that all our RGB, HSL etc functions could be constexpr in the future as they're mainly used to define constant colours.

src/dpp/utility.cpp
```diff
@@ -33,6 +33,7 @@
 #include <fstream>
 #include <streambuf>
 #include <array>
+#include <cmath>
```
Jaskowicz1, line 36: Why is cmath included?

src/dpp/utility.cpp
```diff
@@ -299,7 +300,48 @@ namespace dpp {
 		uint32_t cmyk(int c, int m, int y, int k) {
 			return cmyk(c / 255.0, m / 255.0, y / 255.0, k / 255.0);
 		}
+
+		uint32_t hsl(int h, int s, int l) {
+			double hue = static_cast<double>(h) / 360.0;
+			double saturation = static_cast<double>(s) / 100.0;
+			double lightness = static_cast<double>(l) / 100.0;
+			return hsl(hue,saturation,lightness);
```
Jaskowicz1, line 308: [This does not follow the code convention](https://dpp.dev/coding-standards.html#:~:text=braces%20with%20statements!-,lists,-Lists%20should%20have). It needs to be `hsl(hue, saturation, lightness)`

src/dpp/utility.cpp
```diff
@@ -299,7 +300,48 @@ namespace dpp {
...
+			if (s == 0) {
+				r = g = b = l; // Gray scale
+			} else {
+				double q = l < 0.5 ? l * (1 + s) : l + s - l * s;
+				double p = 2 * l - q;
+				r = hue_to_rgb(p, q, h + 1.0 / 3.0);
+				g = hue_to_rgb(p, q, h);
+				b = hue_to_rgb(p, q, h - 1.0 / 3.0);
+			}
+			return rgb(r,g,b);
```
Jaskowicz1, line 342: ditto the comment on line 308 :)

src/dpp/utility.cpp
```diff
@@ -299,7 +299,48 @@ namespace dpp {
...
+			double lightness = static_cast<double>(l) / 100.0;
+			return hsl(hue, saturation, lightness);
+		}
 		
+		uint32_t hsl(double h, double s, double l) {
+			const auto hue_to_rgb = [](double p, double q, double t){
+				if (t < 0) {
+					t += 1;
+				}
+				if (t > 1) {
```
Jaskowicz1, line 315: I feel this should be an else if, so that we're not double checking if the top value was already done (watch someone snipe this saying that doing `if() if()` is better than `if() else if()`. It could also just be a ternary operator.

src/dpp/utility.cpp
```diff
@@ -299,7 +299,48 @@ namespace dpp {
...
+					t -= 1;
+				} 
+				if (t < 1.0 / 6.0) {
+					return p + (q - p) * 6.0 * t;
+				}
+				if (t < 0.5) {
```
Jaskowicz1, line 321: I think these could be `if else` too.

Jaskowicz1 (approved):

PR #931, opened 2023-10-09
fix: threads now send correctly to discord.

Jaskowicz1:
This PR fixes an issue with threads, as they weren't sending correctly to discord [according to this part of the documentation](https://discord.com/developers/docs/resources/channel#:~:text=json%20params%20(thread)). This means threads can now be locked and archived correctly.

## Code change checklist

### DIFF
--- a/src/dpp/channel.cpp
+++ b/src/dpp/channel.cpp
@@ -484,7 +484,7 @@ std::string thread::build_json(bool with_id) const {
 	j["auto_archive_duration"] = this->metadata.auto_archive_duration;
 	j["locked"] = this->metadata.locked;
 	
-	if(this->get_type() == dpp::channel_type::CHANNEL_PRIVATE_THREAD) {
+	if(is_private_thread()) {
 		j["invitable"] = this->metadata.invitable;
 	}
 
--- a/src/dpp/channel.cpp
+++ b/src/dpp/channel.cpp
@@ -478,14 +478,34 @@ channel& channel::fill_from_json(json* j) {
 }
 
 std::string thread::build_json(bool with_id) const {
-	json j = json::parse(channel::build_json(with_id));
-	j["type"] = (flags & CHANNEL_TYPE_MASK);
-	j["thread_metadata"] = this->metadata;
-	if (!this->applied_tags.empty()) {
-		j["applied_tags"] = json::array();
-		for (auto &tag_id: this->applied_tags) {
-			if (tag_id) {
-				j["applied_tags"].push_back(tag_id);
+	json j;
+	j["name"] = this->name;
+	j["archived"] = this->metadata.archived;
+	j["auto_archive_duration"] = this->metadata.auto_archive_duration;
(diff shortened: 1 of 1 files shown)


src/dpp/channel.cpp
```diff
@@ -478,14 +478,34 @@ channel& channel::fill_from_json(json* j) {
 }
 
 std::string thread::build_json(bool with_id) const {
-	json j = json::parse(channel::build_json(with_id));
```
braindigitalis, line 481: removing this line removes all attributes from channel we might want to keep like type, name, parent id, etc?

braindigitalis (changes requested):

src/dpp/channel.cpp
```diff
@@ -478,14 +478,34 @@ channel& channel::fill_from_json(json* j) {
...
-	j["type"] = (flags & CHANNEL_TYPE_MASK);
-	j["thread_metadata"] = this->metadata;
-	if (!this->applied_tags.empty()) {
-		j["applied_tags"] = json::array();
-		for (auto &tag_id: this->applied_tags) {
-			if (tag_id) {
-				j["applied_tags"].push_back(tag_id);
+	json j;
+	j["name"] = this->name;
```
braindigitalis, line 482: name comes from the parent dpp::channel json

src/dpp/channel.cpp
```diff
@@ -478,14 +478,34 @@ channel& channel::fill_from_json(json* j) {
...
+	j["archived"] = this->metadata.archived;
+	j["auto_archive_duration"] = this->metadata.auto_archive_duration;
+	j["locked"] = this->metadata.locked;
+	
+	if(is_private_thread()) {
+		j["invitable"] = this->metadata.invitable;
+	}
+
+	j["rate_limit_per_user"] = this->rate_limit_per_user;
```
braindigitalis, line 491: rate_limit_per_user comes from the parent json

src/dpp/channel.cpp
```diff
@@ -478,14 +478,34 @@ channel& channel::fill_from_json(json* j) {
...
+
+	if (is_forum() || is_media_channel()) {
```
braindigitalis, line 493: is_forum or is_media_channel are set on the parent, are they not? not the thread within it?

src/dpp/channel.cpp
Jaskowicz1: Yeah, I completely forgot about the use case of creating threads. I'll see if we could maybe keep the old info whilst updating stuff.

src/dpp/channel.cpp
Jaskowicz1: 👍

Jaskowicz1: 👍

src/dpp/channel.cpp
Jaskowicz1: 👍

braindigitalis (approved):

PR #945, opened 2023-10-14
fix: dont require guild cache to fill member roles in message

braindigitalis:
In the event all caches are disabled, roles list in the member for a message was empty. This was because it was internally using find_guild and skipping this part of the deserialisation. It does not need to do this. This fix ensures that if all caching is off, and we have the guild members intent, the guild member list in the messages member field will be popupated. I'm relying on this in the yeet bot.

## Code change checklist

### DIFF
--- a/src/dpp/message.cpp
+++ b/src/dpp/message.cpp
@@ -1053,7 +1053,7 @@ message& message::fill_from_json(json* d, cache_policy_t cp) {
 	/* Fill in member record, cache uncached ones */
 	guild* g = find_guild(this->guild_id);
 	this->member = {};
-	if (g && d->find("member") != d->end()) {
+	if (guild_id && d->find("member") != d->end()) {
 		json& mi = (*d)["member"];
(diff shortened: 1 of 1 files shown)


src/dpp/message.cpp
```diff
@@ -1053,30 +1053,30 @@ message& message::fill_from_json(json* d, cache_policy_t cp) {
 	/* Fill in member record, cache uncached ones */
 	guild* g = find_guild(this->guild_id);
```
Jaskowicz1, line 1054: This should be moved to be defined as part of the else if, on line `1065` as, if you have the cache off, you're still calling the cache method and then never using it.

Jaskowicz1 (changes requested):

src/dpp/message.cpp
Jaskowicz1: After conversations in discord, this will be resolved in a separate PR.

Jaskowicz1 (approved):

Mishura4 (approved):

PR #955, opened 2023-10-19
docs: added a note to setting_status for guild cache

Jaskowicz1:
This PR adds a note to `setting_status.md` about the guild cache, telling people how they can still display a guild count without the guild cache on (if they wish to do so). I'm not entirely sure if you can safely call the two methods I provided every 120 seconds without getting rate-limited, so if there's a better method then please do say!


braindigitalis (changes requested):
I wouldn't tell people to call get guilds. the result is huge, and doesn't work if the bot is on over 1000 guilds. it's also heavily rate limited.
application_get_current is the only practical option without cache.

getting application also works without any privileged intents

braindigitalis (approved):

PR #959, opened 2023-10-20
feat: added roles to emojis

Jaskowicz1:
This PR adds roles to emojis, allowing bots to restrict emojis to certain roles. You can view discord's documentation about this at [this page](https://discord.com/developers/docs/resources/emoji#emoji-object).

This also changes how we store the user, along with how we send information to discord. We now store the actual user object rather than a user_id (as discord provides the user object, this helps people who may not have cache on so they don't have to make another API request to get the user).

The information above is a breaking change.

## Code change checklist

### DIFF
--- a/include/dpp/emoji.h
+++ b/include/dpp/emoji.h
@@ -72,7 +72,7 @@ protected:
 	json to_json_impl(bool with_id = false) const;
 
 public:
-	std::string 			name{};		//<! emoji name
+	std::string 			name{};		//!< emoji name
 	std::vector<snowflake>		roles;		//!< roles allowed to use this emoji
 	snowflake 			user_id;	//!< user id that created this emoji
 	std::string 			image_data{};	//!< Image data for the emoji if uploading
--- a/include/dpp/emoji.h
+++ b/include/dpp/emoji.h
@@ -73,7 +73,7 @@ protected:
 
 public:
 	std::string 			name{};		//<! emoji name
-	std::vector<dpp::snowflake>	roles;		//!< roles allowed to use this emoji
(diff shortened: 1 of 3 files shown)


Jaskowicz1:
image

Here's an image of it working! In this image, I restricted an emoji to my role only, meaning Brain couldn't use the emoji "memelol" until I set the roles variable to a blank array.

braindigitalis (approved):

include/dpp/emoji.h
```diff
@@ -71,22 +71,11 @@ class DPP_EXPORT emoji : public managed, public json_interface<emoji> {
...
-	snowflake user_id{0};
-	/**
-	 * @brief Flags for the emoji from dpp::emoji_flags
-	 */
-	uint8_t flags{0};
-	/**
-	 * @brief Image data for the emoji if uploading
-	 */
-	std::string image_data{};
+	std::string 			name{};		//!< emoji name
```
Mishura4, line 74: Why change the documentation formatting here?

Mishura4 (changes requested):

include/dpp/emoji.h
braindigitalis: yeah probably best not to, because we're getting rid of all the `//!<` formatting junk

Jaskowicz1: this was already agreed upon in discord that we would ignore this and sort it in a later PR.

Mishura4: But why does your PR change the style?
This is *your* PR introducing the style we want to stop using, this sounds like the perfect time to revert this change

Mishura4 (approved):

PR #961, opened 2023-10-20
fix: make embed color optional

erics118:
made `embed::color` a `std::optional`
Discord does not use `0x000000` as the default embed color, so this actually does result in a visual change to the embeds created that dont have a color, (also has a more noticable change in light mode)

Before:
CleanShot 2023-10-20 at 17 05 12CleanShot 2023-10-20 at 17 05 04
After:
CleanShot 2023-10-20 at 17 05 21CleanShot 2023-10-20 at 17 05 06

## Code change checklist

### DIFF
--- a/include/dpp/message.h
+++ b/include/dpp/message.h
@@ -658,7 +658,7 @@ struct DPP_EXPORT embed {
 	/** Optional: timestamp of embed content */
 	time_t				timestamp;
 	/** Optional: color code of the embed */
-	uint32_t			color;
+	std::optional<uint32_t>			color;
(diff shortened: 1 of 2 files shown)


Mishura4 (changes requested):
Might as well change https://github.com/brainboxdotcc/DPP/blob/b6854d3d26a2f25452994507747cdbfaeb8efd1d/src/dpp/message.cpp#L621 as well so it doesn't fill it to 0 even if the field is present

Mishura4 (approved):

Issue #957, opened 2023-10-19
D++ does not support setting images (e.g) banner through guild_edit


Mishura4:
It seems the [Modify Guild](https://discord.com/developers/docs/resources/guild#modify-guild) endpoint supports setting images, for example the guild banner, through the guild object. This requires sending the image data encoded as Base64 in the `banner` field.

[cluster::guild_edit](https://dpp.dev/classdpp_1_1cluster.html#a54e2174822bb1796a541f849d568664c) takes a dpp::guild as a parameter. However its image fields, for example [banner](https://dpp.dev/classdpp_1_1guild.html#ad5736b31cf35641cb2c78f9353d3a8a5) are a [utility::iconhash](https://dpp.dev/structdpp_1_1utility_1_1iconhash.html), which do not support base64 strings as they are a pair of integers.

Supporting this will require refactoring those fields as well as how retrieving the image URLs work.

Jaskowicz1:
I'll fix this :)

braindigitalis:
please don't change how we store the banner. iconhash is the most efficient storage for it. instead please provide a method for uploading a guild banner without adding variables to the class which will make it bigger and be empty 99.999% of the time

Jaskowicz1:
> please don't change how we store the banner. iconhash is the most efficient storage for it. instead please provide a method for uploading a guild banner without adding variables to the class which will make it bigger and be empty 99.999% of the time

The idea isn't to change how it's stored, but I do need to discuss how to tackle some of the issues on discord.

Neko-Life:
add a new method to set banner with base64 string. `set_banner_base64` that accept string param, possibly with some util to read image file as base64 `readfile_base64` for convenience

Jaskowicz1:
> > please don't change how we store the banner. iconhash is the most efficient storage for it. instead please provide a method for uploading a guild banner without adding variables to the class which will make it bigger and be empty 99.999% of the time
> 
> My idea isn't to change how it's stored, but I do need to discuss how to tackle some of the issues on discord.

i lied

Jaskowicz1:
I've decided i'm not gonna take the issue :)

PR #965, opened 2023-10-21
fix(#957): support iconhash/image data variant fields

Mishura4:
WIP fix for #957 
Do not squash

## Code change checklist

### DIFF
--- a/include/dpp/emoji.h
+++ b/include/dpp/emoji.h
@@ -89,7 +89,7 @@ public:
 	/**
 	 * @brief Image data for the emoji, if uploading.
 	 */
-	std::string image_data;
+	utility::image_data image_data;
 
 	/**
 	 * @brief Flags for the emoji from dpp::emoji_flags.
@@ -185,7 +185,7 @@ public:
 	bool is_available() const;
 
 	/**
-	 * @brief Load an image into the object as base64
+	 * @brief Load an image into the object
 	 *
 	 * @param image_blob Image binary data
 	 * @param type Type of image. It can be one of `i_gif`, `i_jpg` or `i_png`.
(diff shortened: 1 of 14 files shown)


braindigitalis:
my only concern with this change is that it changes ALL iconhash, giving them ALL this new upload method, when only ONE icon in the entire of discord api supports this approach (guild banner). This seems to then portray to the user they can upload any icon this way (bot pfp, guild icon, etc) when they cant. There are also other api endpoints that support different methods of uploads (message creation, webhook creation - both these are massively different in the way they work with webhook being closest to what guild banner does here).
Do we have a way in place to prevent the upload functionality working in places where it should not?
Also is the variant smaller, or larger than the iconhash? I dont know the internals of variant, isnt it as big as its largest possible type in the template?

braindigitalis:
Looking into the way theoretically a variant might work, the size of the icon type will be no larger than iconhash, because sizeof(iconhash) would be larger than sizeof(image_data). Therefore the only question remains is: how do we make it so that it doesnt seem to users like they could use the upload methods on a user pfp or something?

Mishura4:
(please do not squash merge - many changes! 📈)

include/dpp/emoji.h
```diff
@@ -194,6 +194,16 @@ class DPP_EXPORT emoji : public managed, public json_interface<emoji> {
...
 
+	/**
+	 * @brief Load an image into the object
+	 *
+	 * @param image_blob Image binary data
+	 * @param type Type of image. It can be one of `i_gif`, `i_jpg` or `i_png`.
+	 * @return emoji& Reference to self
+	 * @throw dpp::length_exception Image content exceeds discord maximum of 256 kilobytes
+	 */
+	emoji& load_image(const std::byte* data, uint32_t size, const image_type type);
```
braindigitalis, line 205: why are we using std::byte instead of e.g. uint8_t?
std::byte has a very limited number of operators associated with its wrapper class which make it very difficult to do basic things with at times.

braindigitalis: Also dpp::utility::read_file does not return an array of std::byte, it returns string.

include/dpp/role.h
```diff
@@ -74,45 +74,71 @@ class DPP_EXPORT role : public managed, public json_interface<role> {
 	 * @brief Role name
 	 * Between 1 and 100 characters.
 	 */
-	std::string name;
+	std::string name{};
 	/**
 	 * @brief Guild ID
 	 */
-	snowflake guild_id;
+	snowflake guild_id{0};
```
braindigitalis, line 81: no need for this change to the class, makes the PR bigger.
We usually set defaults in the constructor, not via {} in the header.

braindigitalis:
When this is merged to dev, which examples in the dpp docs will need to be changed?

braindigitalis (approved):

Jaskowicz1:
> When this is merged to dev, which examples in the dpp docs will need to be changed?

From how this PR looks, no docs need to be changed.

PR #980, opened 2023-10-31
fix: exorcised the scary head. halloween is over

Mishura4:
Please go the the `Preview` tab and select the appropriate pull request template for your changes:

* [Code Changes](?expand=1&template=code_change_template.md)
* [Documentation Changes](?expand=1&template=docs_change_template.md)

### DIFF
--- a/src/dpp/cluster.cpp
+++ b/src/dpp/cluster.cpp
@@ -35,16 +35,6 @@
 
 namespace dpp {
 
-#ifdef _WIN32
-	#ifdef _DEBUG
-		extern "C" void you_are_using_a_debug_build_of_dpp_on_a_release_project() {
-		}
-	#else
-		extern "C" void you_are_using_a_release_build_of_dpp_on_a_debug_project() {
-		}
-	#endif
-#endif
-
 /**
  * @brief An audit reason for each thread. These are per-thread to make the cluster
  * methods like cluster::get_audit_reason and cluster::set_audit_reason thread safe across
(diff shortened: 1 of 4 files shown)


Mishura4:
The links to select a template didn't work.

Mishura4:
Here's an example of the new error message:
image

src/dpp/cluster.cpp
```diff
@@ -70,6 +60,38 @@ template<typename T> std::function<void(const T&)> make_intent_warning(cluster*
...
+			"Mismatched Debug/Release configurations between project and dpp.dll.\n"
+			"Please ensure both your program and the D++ DLL file are both using the same configuration.\n"
+			"The program will now terminate.",
+			"D++ Debug/Release mismatch",
+			MB_OK | MB_ICONERROR
+		);
+		/* Use std::runtime_rror here because dpp exceptions use std::string and that would crash when catching, because of ABI */
+		throw std::runtime_error("Mismatched Debug/Release configurations between project and dpp.dll");
+	}
+	return true;
```
braindigitalis, line 83: does it need to return bool at all? either it's valid and returns or it's not and it throws. there is no false 😁

braindigitalis (changes requested):

src/dpp/cluster.cpp
Mishura4: It needs to return *something*. It's in conjunction with the global variables at https://github.com/brainboxdotcc/DPP/pull/980/files#diff-904a828bab8b0e5e1fa5ab9a9f230d8675fbedc077dac48d80bacb56b2e3cc70R106, they need to be initialized with *something* so that the function is called on program startup.

braindigitalis: what stops is_valid_config being optimised away as unused? I guess it's never been an issue, I'm just curious

Mishura4: When linking D++ dynamically, the compiler cannot determine that a call to `validate_configuration` has no side effects, therefore it will not optimize it away, as that would potentially violate the as-if rule (a compiler must maintain observable behavior when optimizing)

Specifically: https://eel.is/c++draft/basic.stc.static#2

PR #999, opened 2023-11-06
refactor: changed the way messages holds files

Jaskowicz1:
This PR changes the way that messages hold files. Rather than there being 3 vectors of strings for: name, content and mimetypes, there is now a vector of `message_file_data` structs.

This doesn't break the way you add files, but rather breaks the way you read files you've added to a message.

## Code change checklist

### DIFF
--- a/src/dpp/cluster/appcommand.cpp
+++ b/src/dpp/cluster/appcommand.cpp
@@ -136,39 +136,19 @@ void cluster::guild_commands_get(snowflake guild_id, command_completion_event_t 
 }
 
 void cluster::interaction_response_create(snowflake interaction_id, const std::string &token, const interaction_response &r, command_completion_event_t callback) {
-	std::vector<std::string> file_names{};
-	std::vector<std::string> file_contents{};
-	std::vector<std::string> file_mimetypes{};
-
-	for(message_file_data data : r.msg.file_data) {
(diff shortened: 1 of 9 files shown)


Jaskowicz1:
image
image
This is proof of it still working!

src/dpp/cluster/appcommand.cpp
```diff
@@ -136,51 +136,101 @@ void cluster::guild_commands_get(snowflake guild_id, command_completion_event_t
...
-	}, r.msg.filename, r.msg.filecontent, r.msg.filemimetype);
+	}, file_names, file_contents, file_mimetypes);
 }
 
 void cluster::interaction_response_edit(const std::string &token, const message &m, command_completion_event_t callback) {
+	std::vector<std::string> file_names{};
+	std::vector<std::string> file_contents{};
+	std::vector<std::string> file_mimetypes{};
+
+	for(message_file_data data : m.file_data) {
```
braindigitalis, line 161: can't we just send the vector over to the function instead of splitting it into 3 like this? seems hacky to split it like this.

braindigitalis (changes requested):

Jaskowicz1:
Reverted to draft as more changes are required.

Jaskowicz1:
No longer in draft, ready for review!

PR #1001, opened 2023-11-13
fix: fixed unittest for mac, ctest now works

Jaskowicz1:
This PR fixes unit tests on Mac, along with also fixing `ctest -VV` so it now runs the unit test.

This PR also adds .DS_STORE to the .gitignore.

## Code change checklist

- [x] I have ensured that all methods and functions are **fully documented** using doxygen style comments.
- [x] My code follows the [coding style guide](https://dpp.dev/coding-standards.html).
- [x] I tested that my change works before raising the PR.
- [x] I have ensured that I did not break any existing API calls.
- [x] I have not built my pull request using AI, a static analysis tool or similar without any human oversight.

PR #1003, opened 2023-11-14
ci: unit tests now use ctest, macos now unit tests

Jaskowicz1:
This PR modifies the ci.yml file, making our unit tests now use `ctest -VV`. Since #1001 fixed the unit tests for Mac OS, we now unit test on the Mac OS runner.

The checklists do not apply to this PR.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -157,15 +157,8 @@ jobs:
         env:
           DONT_RUN_VCPKG: true
 
-      - name: Run unit tests
+      - name: Run offline unit tests
         run: cd build && ctest -VV
-        env:
-          DPP_UNIT_TEST_TOKEN: ${{secrets.DPP_UNIT_TEST_TOKEN}}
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -157,6 +157,16 @@ jobs:
         env:
           DONT_RUN_VCPKG: true
 
+      - name: Run unit tests
+        run: cd build && ctest -VV
+        env:
+          DPP_UNIT_TEST_TOKEN: ${{secrets.DPP_UNIT_TEST_TOKEN}}
```
craigpipeandpiper, line 163: what if the mac unit test runs at the same time as the linux one? their messages will conflict and both will fail, because this is how discord api works, both will see each others output. One of them (this one?) should be offline only unit tests.

Jaskowicz1: Oh right, yeah Mac can do the offline test then :)

braindigitalis (changes requested):
see comment

braindigitalis (approved):

PR #1010, opened 2023-11-19
ci: Mac changes, self-hosted runner now available

Jaskowicz1:
This PR changes the ci.yml workflow to now use a matrix for MacOS. This also means we now have a self-hosted Mac runner (using Apple silicone) available!

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -133,8 +133,15 @@ jobs:
   macos:
     permissions:
       contents: write
-    name: macOS x64
-    runs-on: macos-latest
+    name: macOS ${{matrix.cfg.arch}} (${{matrix.cfg.cpp-version}})
+    runs-on: ${{matrix.cfg.os}}
+    strategy:
+      fail-fast: false # Don't fail everything if one fails. We want to test each OS/Compiler individually
+      matrix:
+        # arm64 is a self-hosted runner on a Mac M2 Mini, ran inside a virtual machine by Archie Jaskowicz.
+        cfg:
+          - { arch: 'x64', concurrency: 3, os: ubuntu-20.04, cpp-version: clang++-10, cmake-flags: '', cpack: 'no' }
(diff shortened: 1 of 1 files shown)


Jaskowicz1:
Timezone will be changed on the Mac VM to be UTC in a few hours. That should fix the unit test.

I also want to note, I’ve purposefully made it so both Mac architectures run the unit test, as a precaution. Though, I would like to look into the possibility of running two online unit tests (one on Linux, one on Mac arm64)

braindigitalis:
the only way to have two concurrent runners is to have two tokens and two test servers with different channel, bot and event IDs. it would likely be overkill, plus each would need to be limited to one at a time like the Linux unit tests run is.

Jaskowicz1:
> the only way to have two concurrent runners is to have two tokens and two test servers with different channel, bot and event IDs. it would likely be overkill, plus each would need to be limited to one at a time like the Linux unit tests run is.

Ah right, yeah that's fair enough.

Jaskowicz1:
Timezone has been fixed on the Mac (it appeared to be under PST, it is now under UTC) and the CI has passed. This PR is ready!

braindigitalis:
> Timezone has been fixed on the Mac (it appeared to be under PST, it is now under UTC) and the CI has passed. This PR is ready!

what, don't you know? EVERYONE lives in the USA 😆

braindigitalis (approved):

Jaskowicz1:
> > Timezone has been fixed on the Mac (it appeared to be under PST, it is now under UTC) and the CI has passed. This PR is ready!
> 
> what, don't you know? EVERYONE lives in the USA 😆

🤣

Jaskowicz1:
Downscaled the VM due to it requesting a lot of resources. Will re-run the action to make sure it hasn't slowed down too much.

Jaskowicz1:
Job finished in 3 minutes 48 seconds, even when downscaled!

PR #1036, opened 2023-12-04
feat: add error codes to exceptions

braindigitalis:
This lets users check what type of exception by `exception.code()` instead of having to rely on string comparison of `exception.what()`. This is a non-breaking change as the previous functionality still exists alongside the old style constructors for anyone who is directly instantiating their own dpp exceptions in their bot (not sure why they would be, though).

Note the huge enum `exception_error_code` in exception.h. This will need to be maintained periodically by us as Discord add new errors. It can also be used with `get_error()` of a callback and matched against `get_error().code`.

## Code change checklist

### DIFF
--- a/include/dpp/exception.h
+++ b/include/dpp/exception.h
@@ -106,7 +106,7 @@ enum exception_error_code {
 	err_unknown = 37,
 	err_bad_request = 400,
 	err_unauthorized = 401,
-	err_payment_requred = 402,
+	err_payment_required = 402,
 	err_forbidden = 403,
 	err_not_found = 404,
 	err_method_not_allowed = 405,
--- a/include/dpp/exception.h
+++ b/include/dpp/exception.h
@@ -43,7 +43,7 @@ namespace dpp {
  * here. In this case, or where the code is not specific, refer to `what()`.
  */
 enum exception_error_code {
-	err_this_is_not_an_error = 0,
+	err_no_code_specified = 0,
 	err_zlib_see_errno = -1,
 	err_zlib_init_stream = -2,
 	err_zlib_init_data = -3,
@@ -411,7 +411,7 @@ public:
 	 * 
 	 * @param what reason message
 	 */
-	explicit exception(const char* what) : msg(what), error_code(err_this_is_not_an_error) { }
+	explicit exception(const char* what) : msg(what), error_code(err_no_code_specified) { }
 
 	/**
 	 * @brief Construct a new exception object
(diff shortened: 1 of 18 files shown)


include/dpp/exception.h
```diff
@@ -53,22 +411,38 @@ class exception : public std::exception
...
+	explicit exception(exception_error_code code, const char* what) : msg(what), error_code(code) { }
 
 	/**
 	 * @brief Construct a new exception object
 	 * 
 	 * @param what reason message
 	 * @param len length of reason message
 	 */
-	exception(const char* what, size_t len) : msg(what, len) { }
+	exception(const char* what, size_t len) : msg(what, len), error_code(err_this_is_not_an_error) { }
```
Mishura4, line 430: Should we make the default "unknown error" or something like that instead? Otherwise this will make the usual way of constructing an exception "not an error", that doesn't feel right

include/dpp/exception.h
```diff
@@ -53,22 +411,38 @@ class exception : public std::exception
...
 
 	/**
 	 * @brief Construct a new exception object
 	 * 
 	 * @param what reason message
 	 */
-	explicit exception(const std::string& what) : msg(what) { }
+	explicit exception(const std::string& what) : msg(what), error_code(err_this_is_not_an_error) { }
```
Mishura4, line 437: Same here

include/dpp/sync.h
```diff
@@ -27,54 +27,55 @@
...
+	 * @tparam F Type of class method in dpp::cluster to call.
+	 * @tparam Ts Function parameters in method call
+	 * @param c A pointer to dpp::cluster object
+	 * @param func pointer to class method in dpp::cluster to call. This can call any
+	 * dpp::cluster member function who's last parameter is a dpp::command_completion_event_t callback type.
+	 * @param args Zero or more arguments for the method call
+	 * @return An instantiated object of type T
+	 * @throw dpp::rest_exception On failure of the method call, an exception is thrown
+	 */
+	template<typename T, class F, class... Ts> T sync(class cluster* c, F func, Ts&&... args) {
```
Jaskowicz1, line 54: This shouldn't be tab-indented as it's a namespace, https://discord.com/channels/825407338755653642/825411104208977952/1159604805979881472

include/dpp/exception.h
braindigitalis: we already have unknown error :) its value is 37

braindigitalis: perhaps `this_exception_has_no_code`

Jaskowicz1 (changes requested):

PR #1043, opened 2023-12-19
fix: avoid memory leaks in event_map

rept1d:
This PR fixes memory leaks detected by Valgrind which occur because events allocated in initialization of `eventmap` are never deallocated.

These leaks are harmless because these objects are currently supposed to live for the whole duration of the program, but it's still good to not have these leaks at all. I fix them by making the pointers stored in `eventmap` (renamed to `event_map` for better readability) point to events with static storage duration. This way we also avoid unnecessary memory allocations.

I should note that it seems a bit weird to me that these event objects live that long in the first place. They have no reason for that other than making the implementation simpler. It might be worth looking for a cleaner a solution in a separate PR.

## Code change checklist


Jaskowicz1:
Congrats on your first PR! I know you said you've tested this, but have you ran a unit test with this and made sure there's no errors there?

braindigitalis:
pretty sure there's already a type called event_map (a container of guild events) but apart from that looks good

rept1d:
> pretty sure there's already a type called event_map (a container of guild events) but apart from that looks good

I don't think there is, there are only `scheduled_event_map` and `event_member_map`.

braindigitalis (approved):

PR #1044, opened 2023-12-19
fix: use of uninitialized value in command_option copy ctor

rept1d:
Basically all examples of using `dpp::command_option` invoke undefined behavior Let's look at its typical usage example from the docs:
```cpp
dpp::slashcommand newcommand("blep", "Send a random adorable animal photo", bot.me.id);
newcommand.add_option(
    dpp::command_option(dpp::co_string, "animal", "The type of animal", true)
);
```

The problem here is that `dpp::command_option::command_option` does not initialize a boolean field called `focused`. And then inside `add_option` the implicitly-defined copy constructor is called. This constructor, among other things, copies that `focused` field, but since it was never initialized, we get UB.

This PR ensures that `focused` is initialized in `dpp::command_option::command_option` with `false`.

## Code change checklist

### DIFF
--- a/src/dpp/slashcommand.cpp
+++ b/src/dpp/slashcommand.cpp
@@ -331,7 +331,7 @@ command_option_choice &command_option_choice::fill_from_json_impl(nlohmann::json
 }
 
 command_option::command_option(command_option_type t, const std::string &n, const std::string &d, bool r) :
-	type(t), name(n), description(d), required(r), autocomplete(false)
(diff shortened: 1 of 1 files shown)


Jaskowicz1 (approved):
I'm honestly surprised this didn't already exist lol

braindigitalis:
the reason this never caused an issue is because focused is only set in an auto complete event at which point it is explicitly set to true or false on each event. this just squashes warnings/static analysis errors and makes things clean and explicit. looks good.

PR #1046, opened 2023-12-19
refactor: fix and improve utf8 utilities

rept1d:
This PR solves the following problems:
- `dpp::utf8substr` is just broken: `dpp::utf8substr("abcdefg", 2, 2)` returns `"cdef"`. Probably unnoticed due to a very limited variety of usage scenarios.
- `dpp::utf8substr` and `dpp::utf8len` claim to have a specific result (empty string or zero, respectively) if the supplied string is not a valid UTF-8 sequence. In reality though, not only do they not verify if the string is valid UTF-8, they actually invoke UB for some inputs (e.g. `dpp::utf8len("\xC0\xC0")`) due to reading past the end of the string.
- The implementations of these functions are convoluted (use of `goto`, defining lots of variables in one line etc.).
- They only work with `std::string`s and not `std::string_view`s. In particular, `dp::utf8substr` returns a new `std::string`, which causes allocations and copies that could be avoided in some cases.

What I did:
- Dropped the "guarantees" on what exactly these functions return in case of non-UTF-8 strings. We shouldn't pretend that they are provided. Now we claim that the result is "unspecified" (so no UB).
- Introduced a separate function called `utf8subview` which returns `std::string_view` instead of `std::string`, and made use of it in several places across the code base, where applicable.
- Replaced `const std::string&` with `std::string_view` in parameter declarations of these functions.
- Replaced `std::string::size_type` with `size_t` in parameter declarations of these functions because IMO it only makes everything unnecessarily verbose, especially if we assume that `std::string::size_type` might be different from `std::string_view::size_type`. It's almost never used in other places anyway, so why bother in this one in particular?
- Completely rewrote the implementations of both functions. One important difference in their behavior is that now they don't treat null terminators in the middle of strings as their end markers. These are now no different from any other ASCII character. It's also important that we don't rely on their presence at the end because `std::string_view`s are not guaranteed to be null-terminated.

## Code change checklist

### DIFF
--- a/include/dpp/utility.h
+++ b/include/dpp/utility.h
@@ -738,12 +738,13 @@ uint32_t DPP_EXPORT hsl(int h, int s, int l);
 std::string DPP_EXPORT debug_dump(uint8_t* data, size_t length);
 
 /**
- * @brief Returns the length of a UTF-8 string in codepoints
- * 
+ * @brief Returns the length of a UTF-8 string in codepoints.
+ * @note Result is unspecified for strings that are not valid UTF-8.
+ *
  * @param str string to count length of
- * @return size_t length of string (0 for invalid utf8)
+ * @return size_t Length of string
  */
-size_t DPP_EXPORT utf8len(const std::string &str);
+size_t DPP_EXPORT utf8len(std::string_view str);
 
 /**
  * @brief Return substring of a UTF-8 encoded string in codepoints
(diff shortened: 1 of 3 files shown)


braindigitalis (approved):

Mishura4 (approved):
Looks good, we should add unit tests for all of this as well, though this can be a separate commit

Jaskowicz1:
Super detailed PR description, I love it!

PR #1053, opened 2023-12-29
fix: cmyk calculations - blue value calculations

harrrson:
During conversion from cmyk to rgb, blue value was calculated with m (magenta) instead of k (black)

## Code change checklist


braindigitalis (approved):

braindigitalis:
thanks for the PR! we will merge this when the CI completes

PR #1065, opened 2024-01-18
feat: Added expire and issued timestamps in attachment

harrrson:
Added methods to extract expiration and issued timestamps from url

```c++
dpp::embed test{};
test.set_title("Test for file")
    .add_field("filename", fileAttachment.filename)
    .add_field("description", fileAttachment.description)
    .add_field("url", fileAttachment.url)
    .add_field("proxy_url", fileAttachment.proxy_url)
    .add_field("Expiration timestamp", std::to_string(fileAttachment.get_expire_time()))
    .add_field("issuing timestamp", std::to_string(fileAttachment.get_issued_time()));
```
obraz

## Code change checklist

### DIFF
--- a/src/dpp/message.cpp
+++ b/src/dpp/message.cpp
@@ -883,10 +883,14 @@ bool attachment::is_remix() const {
 }
 
 time_t attachment::get_expire_time() const {
-	std::string attributes = url.substr(url.find('?')+1);
+	size_t attr_position = url.find('?');
+	/*If no attributes were sent in url, we do not need to parse more*/
+	if(url.npos == attr_position){
+		return 0;
+	}
+	std::string attributes = url.substr(attr_position+1);
 	std::vector<std::string> attr_list = utility::tokenize(attributes,"&");
 	auto ex_attr = std::find_if(attr_list.begin(), attr_list.end(),[](const std::string& s){return s.substr(0,3) == "ex=";});
-	/*Check if discord sent us the 'ex' attribute*/
 	if(attr_list.end() == ex_attr){
 		return 0;
 	}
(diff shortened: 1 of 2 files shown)


src/dpp/message.cpp
```diff
@@ -882,6 +882,38 @@ bool attachment::is_remix() const {
 	return flags & a_is_remix;
 }
 
+time_t attachment::get_expire_time() const {
+	size_t attr_position = url.find('?');
```
Jaskowicz1, line 886: This can be `const`

src/dpp/message.cpp
```diff
@@ -882,6 +882,38 @@ bool attachment::is_remix() const {
...
+	/*If no attributes were sent in url, we do not need to parse more*/
+	if(url.npos == attr_position){
+		return 0;
+	}
+	std::string attributes = url.substr(attr_position+1);
+	std::vector<std::string> attr_list = utility::tokenize(attributes,"&");
```
Jaskowicz1, line 892: imo, you should put `url.substr(attr_position+1)` inside `utility::tokenize` as you don't use `attributes` anywhere else but this. However, I'd like other thoughts on this.

src/dpp/message.cpp
```diff
@@ -882,6 +882,38 @@ bool attachment::is_remix() const {
 	return flags & a_is_remix;
 }
 
+time_t attachment::get_expire_time() const {
+	size_t attr_position = url.find('?');
+	/*If no attributes were sent in url, we do not need to parse more*/
```
Jaskowicz1, line 887: Space the comments correctly!
`/* If ......... */`

Jaskowicz1 (changes requested):

src/dpp/message.cpp
Mishura4: which we don't enforce

harrrson: Yeah, generally i do not use const in local variables (no particular reason, just forgetting to use it). I can add it if needed

src/dpp/message.cpp
braindigitalis: please space stuff out, (attr_position + 1) not (attr_position+1)

src/dpp/message.cpp
Jaskowicz1: > Yeah, generally i do not use const in local variables (no particular reason, just forgetting to use it). I can add it if needed

No need :)

src/dpp/message.cpp
harrrson: I generally like to add intermediate variables to make the code easier to follow, iirc compiler optimizes them away if able to

src/dpp/message.cpp
harrrson: Done

braindigitalis (approved):
approved and good to merge, would be nice if there were unit tests for this though

Jaskowicz1:
> approved and good to merge, would be nice if there were unit tests for this though

A todo list for stuff would be lovely

Jaskowicz1 (approved):

PR #1067, opened 2024-01-21
Added dpp::cluster::guild_member_timeout_remove and its sync/coro versions

AA1999:
I've added dpp::cluster::guild_member_timeout_remove, dpp::cluster::guild_member_timeout_remove_sync and dpp::cluster::co_guild_member_timeout_remove and added documentation for them.

### DIFF
--- a/include/dpp/cluster.h
+++ b/include/dpp/cluster.h
@@ -2390,6 +2390,19 @@ public:
 	void guild_member_timeout(snowflake guild_id, snowflake user_id, time_t communication_disabled_until, command_completion_event_t callback = utility::log_error());
 
 	/**
+	 * @brief Remove the timeout of a guild member.
+	 * A shortcut for guild_member_timeout(guild_id, user_id, 0, callback)
+	 * Fires a `Guild Member Update` Gateway event.
+	 * @see https://discord.com/developers/docs/resources/guild#modify-guild-member
+	 * @note This method supports audit log reasons set by the cluster::set_audit_reason() method.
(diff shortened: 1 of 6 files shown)


braindigitalis:
how did you add the coro and sync version? did you do it by rerunning the builder scripts for it?

AA1999:
I just called guild_timeout_remove_sync and co_guild_timeout with communication_disabled_until set to 0
Returned the result in the coro version

braindigitalis:
the coro and sync are auto generated, you shouldn't manually put stuff in those files. when someone runs cmake it will remove your changes.

you need to make sure the auto generation picks up your changes. because they don't match the correct pattern this won't work as is. you should make them not just call another function and instead call the REST stuff

AA1999:
Where's the script for these?
Is it in one of the CMakeLists?

Jaskowicz1:
> Where's the script for these? Is it in one of the CMakeLists? Also why not call the function with a 0 parameter if I'm doing the exact same thing with one different variable?

No, it happens when you call `cmake -B build` or `cmake ..`

AA1999:
I did use cmake -B and cmake --build before pushing
Dunno if that'd help or not

braindigitalis:
it's a set of php scripts in the buildtools folder that do this, there's a base class, and a derived version of it for coro and sync.

you should not need to modify these scripts.

Jaskowicz1 (approved):
Better this time around!

PR #1068, opened 2024-01-25
ci: labeler no longer fails

Jaskowicz1:
This PR fixes labeler from failing CIs due to the update that was made which broke configuration files.

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -10,6 +10,7 @@ documentation:
     - '**SECURITY.md'
     - '**CONTRIBUTING.md'
     - '**CODE_OF_CONDUCT.md'
+
 build:
 - changed-files:
   - any-glob-to-any-file:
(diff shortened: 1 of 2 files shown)


.github/labeler.yml
```diff
@@ -1,35 +1,52 @@
...
+    - '**vcpkg/**'
+    - '**makerelease.sh'
+    - '**sign.sh'
+    - '**Dockerfile'
+
 submodules:
-- '**.gitmodules'
-- '**doxygen-awesome-css/**' # Ideally, nobody should be touching this, but it's here just in-case.
+- changed-files:
+  - any-glob-to-any-file:
```
braindigitalis, line 37: this new syntax is horrible, I died a bit reading it.
how do the devs of this action sleep at night, those-kebab-case-psychopaths lol

braindigitalis (approved):

.github/labeler.yml
Jaskowicz1: lol ikr

PR #1069, opened 2024-01-26
feat: added newly_created bool to threads

Jaskowicz1:
This PR adds a `newly_created` bool to threads. As `on_thread_create` is also fired when a thread is added to an existing private thread (for example, adding a thread to a forum channel), there was no way to distinguish the actual creation event, Discord provides a `newly_created` bool so we now allow users to see that.

I avoided using the flags here in-case Discord add more flags (as `dpp::channel_flags` is close to its limit) to channels/threads.

## Code change checklist

### DIFF
--- a/include/dpp/thread.h
+++ b/include/dpp/thread.h
@@ -166,6 +166,12 @@ public:
 	uint8_t member_count = 0;
 
 	/**
+	 * @brief Was this thread newly created?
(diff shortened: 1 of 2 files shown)


braindigitalis (approved):

braindigitalis:
threads aren't cached, so bool is fine here

Jaskowicz1:
> threads aren't cached, so bool is fine here

Ah good good!

PR #1083, opened 2024-02-17
Added change_voice_channel to discordclient

SelfishPig:
A disconnect does not need to be issued before switching voice channels. This new function allows you to change voice channels without disconnecting beforehand.

### DIFF
--- a/include/dpp/discordclient.h
+++ b/include/dpp/discordclient.h
@@ -505,6 +505,20 @@ public:
 	discord_client& connect_voice(snowflake guild_id, snowflake channel_id, bool self_mute = false, bool self_deaf = false);
 
 	/**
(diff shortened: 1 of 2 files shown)


Jaskowicz1 (changes requested):
Congrats on your first PR! It's looking great, however, the code is space indented! If you just wanna correct it to be tab indented then it should be good to go!

SelfishPig:
> Congrats on your first PR! It's looking great, however, the code is space indented! If you just wanna correct it to be tab indented then it should be good to go!

Thanks, all done!

Jaskowicz1 (approved):

braindigitalis (approved):

PR #1085, opened 2024-02-18
Merge change_voice_channel into connect_voice

SelfishPig:
There's actually no reason for change_voice_channel to be it's own function. When would a user ever call connect_voice and not expect it to connect to the channel they specified if it's not already connected? This change doesn't break any existing functionality.


Jaskowicz1:
So, will `change_voice_channel` exist alongside this? Your current changes show you're keeping `change_voice_channel` and changing `connect_voice`.

SelfishPig:
> So, will `change_voice_channel` exist alongside this? Your current changes show you're keeping `change_voice_channel` and changing `connect_voice`.

There we go. Cheers.

braindigitalis (approved):

PR #1100, opened 2024-03-01
feat: add embed-only message ctor

osirais:
This PR adds an embed-only constructor to `dpp::message`. Since `message(snowflake channel_id, const embed& _embed)` exists, it seemed natural and useful to have a version without the channel id.

```c++
// this is probably the most common reply on my bot
event.reply(dpp::message{}.add_embed(embed));

// now you could just do this
event.reply(dpp::message{embed});

// or this with the implicit conversion
event.reply(embed);
```

## Code change checklist

### DIFF
--- a/include/dpp/message.h
+++ b/include/dpp/message.h
@@ -2024,12 +2024,19 @@ public:
 	message(snowflake channel_id, const std::string &content, message_type type = mt_default);
 
 	/**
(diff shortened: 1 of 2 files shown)


Mishura4 (approved):
Thank you for contributing!

braindigitalis (approved):

PR #1103, opened 2024-03-03
fix: get size of object instead of pointer

VoltrexKeyva:
The `this` keyword is evaluated to a pointer to an object of a given type so `sizeof(this)` returns the size of the pointer instead of the object so use `sizeof(*this)` instead.

## Code change checklist


braindigitalis (approved):

Mishura4 (approved):
👍

PR #1109, opened 2024-03-22
feat: Added is_guild_owner for dpp::guild_member

AA1999:
…member is the guild owner.

Added a helper method so that checking if the member is the guild owner is easier.

## Code change checklist

- [✅] I have ensured that all methods and functions are **fully documented** using doxygen style comments.
- [✅] My code follows the [coding style guide](https://dpp.dev/coding-standards.html).
- [✅] I tested that my change works before raising the PR.
- [✅] I have ensured that I did not break any existing API calls.
- [✅] I have not built my pull request using AI, a static analysis tool or similar without any human oversight.

### DIFF
--- a/include/dpp/guild.h
+++ b/include/dpp/guild.h
@@ -524,6 +524,7 @@ public:
 	 * @brief Returns true if the user is the guild owner.
 	 * @return true if the user is the guild owner.
 	 * @return false if the user is not the guild owner or the guild is not in the cache.
+	 * @note If dpp cache is disabled (which should be enabled by default) this function will always return false.
 	 */
 	bool is_guild_owner() const;
 
--- a/include/dpp/guild.h
+++ b/include/dpp/guild.h
@@ -521,10 +521,10 @@ public:
 	bool has_rejoined() const;
 
 	/**
-	 * @brief Returns true if the user is the guild owner.
+	 * @brief Is this user also the guild member?
 	 * @return true if the user is the guild owner.
 	 * @return false if the user is not the guild owner or the guild is not in the cache.
-	 * @note If dpp cache is disabled (which should be enabled by default) this function will always return false.
+	 * @note If dpp cache is disabled, this function will always return false.
 	 */
 	bool is_guild_owner() const;
 
--- a/include/dpp/guild.h
+++ b/include/dpp/guild.h
@@ -521,6 +521,13 @@ public:
 	bool has_rejoined() const;
 
 	/**
+	 * @brief Returns true if the user is the guild owner.
+	 * @return true if the user is the guild owner.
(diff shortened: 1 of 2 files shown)


src/dpp/guild.cpp
```diff
@@ -347,6 +347,11 @@ bool guild_member::has_rejoined() const {
 	return flags & dpp::gm_did_rejoin;
 }
 
+bool guild_member::is_guild_owner() const {
+	auto* _guild = find_guild(user_id);
+	return _guild != nullptr && _guild->owner_id == this->user_id;
```
Mishura4, line 352: So if the guild is not found in the cache this returns false, this is a bit weird, do we have any other functions in D++ that do this or is this potentially going to surprise people? @braindigitalis

braindigitalis: it might surprise people who have cache off yes

braindigitalis: it should probably be a method of dpp::guild that takes a user. not really needed though you can just check user id == guild owner id where you have both.

most people do have guild cache on. best to clearly document this assumption in the docs of the function.

src/dpp/guild.cpp
```diff
@@ -347,6 +347,11 @@ bool guild_member::has_rejoined() const {
 	return flags & dpp::gm_did_rejoin;
 }
 
+bool guild_member::is_guild_owner() const {
+	auto* _guild = find_guild(user_id);
```
braindigitalis, line 351: this is wrong. you're using user id to search for a guild

braindigitalis (changes requested):

src/dpp/guild.cpp
AA1999: Yeah I might have messed it up, will fix

include/dpp/guild.h
```diff
@@ -520,6 +520,14 @@ class DPP_EXPORT guild_member : public json_interface<guild_member> {
 	 */
 	bool has_rejoined() const;
 
+	/**
+	 * @brief Returns true if the user is the guild owner.
+	 * @return true if the user is the guild owner.
+	 * @return false if the user is not the guild owner or the guild is not in the cache.
+	 * @note If dpp cache is disabled (which should be enabled by default) this function will always return false.
```
braindigitalis, line 527: this should be reworded:
```cpp
/*
 * @note If the guild cache is disabled in the cluster constructor, this function will always return false.
 */
```
we dont need to tell them that its on by default, other docs indicate this

Jaskowicz1 (approved):

Jaskowicz1:
Renamed this PR to correctly align with our conventions.

braindigitalis (approved):

PR #1123, opened 2024-04-06
fix: Signal Handlers on Non-Windows Platforms in sslclient

Nidhoegger:
Only set signals to Ignore on non-Linux of no signal-handler is in place. Also set the SA_RESTART flag to avoid operations returning EINTR.

## Code change checklist

### DIFF
--- a/src/dpp/sslclient.cpp
+++ b/src/dpp/sslclient.cpp
@@ -250,9 +250,9 @@ ssl_client::ssl_client(const std::string &_hostname, const std::string &_port, b
 {
 #ifndef WIN32
 	set_signal_handler(SIGALRM);
-	set_signal_handler(SIGHUP);
 	set_signal_handler(SIGXFSZ);
 	set_signal_handler(SIGCHLD);
+	signal(SIGHUP, SIG_IGN);
 	signal(SIGPIPE, SIG_IGN);
 #else
 	// Set up winsock.
--- a/src/dpp/sslclient.cpp
+++ b/src/dpp/sslclient.cpp
@@ -229,7 +229,7 @@ void set_signal_handler(int signal)
 	struct sigaction sa;
 	sigaction(signal, nullptr, &sa);
 	if (sa.sa_flags == 0 && sa.sa_handler == nullptr) {
-		memset(&sa, 0, sizeof(sa));
+		sa = {};
 		sigaction(signal, &sa, nullptr);
 	}
 }
--- a/src/dpp/sslclient.cpp
+++ b/src/dpp/sslclient.cpp
@@ -228,8 +228,7 @@ void set_signal_handler(int signal)
 {
 	struct sigaction sa;
 	sigaction(signal, nullptr, &sa);
-	if (sa.sa_flags == 0 && sa.sa_handler == nullptr)
-	{
+	if (sa.sa_flags == 0 && sa.sa_handler == nullptr) {
 		memset(&sa, 0, sizeof(sa));
 		sigaction(signal, &sa, nullptr);
 	}
--- a/src/dpp/sslclient.cpp
+++ b/src/dpp/sslclient.cpp
@@ -238,49 +238,34 @@ ssl_client::ssl_client(const std::string &_hostname, const std::string &_port, b
 {
 #ifndef WIN32
 	struct sigaction sa;
+	struct sigaction action;
+	memset(&action, 0, sizeof(action));
+	action.sa_handler = SIG_IGN;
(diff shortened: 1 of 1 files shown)


Jaskowicz1:
Thanks for your contribution! This PR doesn't seem like it's carried anything from master but next time please make your changes on a branch based from dev, rather than using the master branch!

Jaskowicz1:
Also, next time, please read the description of the PR and choose the template that accurately reflects your PR. I've added the template in for you this time, please fill it out (and correct your PR title to follow our conventions please!)

src/dpp/sslclient.cpp
```diff
@@ -237,11 +237,36 @@ ssl_client::ssl_client(const std::string &_hostname, const std::string &_port, b
...
-	signal(SIGCHLD, SIG_IGN);
-	signal(SIGXFSZ, SIG_IGN);
+	struct sigaction sa;
+	struct sigaction action;
+	memset(&action, 0, sizeof(action));
+	action.sa_handler = SIG_IGN;
+	action.sa_flags = SA_RESTART;
+
+	sigaction(SIGALRM, nullptr, &sa);
+	if (sa.sa_flags == 0 && sa.sa_handler == nullptr) {
```
braindigitalis, line 247: DRY, can you please make this a function and call the function with each signal name instead?

braindigitalis (changes requested):
changes requested

src/dpp/sslclient.cpp
Nidhoegger: Done

src/dpp/sslclient.cpp
```diff
@@ -237,11 +249,11 @@ ssl_client::ssl_client(const std::string &_hostname, const std::string &_port, b
 	keepalive(reuse)
 {
 #ifndef WIN32
-	signal(SIGALRM, SIG_IGN);
-	signal(SIGHUP, SIG_IGN);
+	set_signal_handler(SIGALRM);
+	set_signal_handler(SIGHUP);
```
braindigitalis, line 253: I think SIGHUP should be ignored regardless? if it is captured we don't know if this will interrupt the library and break it? the signals will execute inside the libraries internal threads.

Nidhoegger:
Since not seeing this on GitHub directly, I will reply here and hope it
works. IIRC SIGHUP is generated when the controlling TTY is lost without
prior detaching (earlier it was called daemonization iirc). Today,
frameworks like runit or launchd are used for daemonization of services, so
if SIGHUP is generated, the controlling terminal is closed (most likely)
and thus the process will die anyway. So IMHO it does not change anything
if SIGHUP is ignored or not.

Am Sa., 6. Apr. 2024 um 18:12 Uhr schrieb Craig Edwards (Brain) <
***@***.***>:

Nidhoegger:
As said: If they have a controlling TTY and its closed, they will
terminate, no matter if they ignore SIGHUP or not. If they fork and use
setsid and detach from a terminal, then they wont get SIGHUP on closure.
Ignoring SIGHUP will only make them not get the signal and terminate
anyways.

Am So., 7. Apr. 2024 um 11:05 Uhr schrieb Craig Edwards (Brain) <
***@***.***>:

> There are many bots that don't use launchd etc for launching and will
> directly fork. these still need to be supported
>
> ⁣Kind regards
>
> Craig Edwards
>
> Brainbox.cc
> Creating fun stuff since 1993​
>

Nidhoegger:
Anyway, changed it. There is no harm done in ignoring it IMHO

src/dpp/sslclient.cpp
```diff
@@ -223,6 +223,18 @@ int connect_with_timeout(dpp::socket sockfd, const struct sockaddr *addr, sockle
 #endif
 }
 
+#ifndef _WIN32
+void set_signal_handler(int signal)
+{
+	struct sigaction sa;
+	sigaction(signal, nullptr, &sa);
+	if (sa.sa_flags == 0 && sa.sa_handler == nullptr) {
+		memset(&sa, 0, sizeof(sa));
```
braindigitalis, line 232: why do we memset it to empty here after retrieving it with sigaction?

Nidhoegger:
To set everything to a known state. Not everything is checked and other
fields might be set.Message ID:
***@***.***>

>

Nidhoegger:
Feel free to change it. I dont see a reason to.

Jaskowicz1:
Is this latest change tested?

braindigitalis (approved):

PR #1127, opened 2024-04-13
Fix/uncleared track meta on stop audio

### LINKED ISSUE #1124: dpp::discord_voice_client: track_meta is not cleared when stop_audio() is called
**Git commit reference**
89b43ba1445f99114f1c583529d01ad210918c35

**Describe the bug**
When voice client's stop_audio() is called, all 2-byte markers `0xFFFF` in `outbuf` are deleted, however `track_meta` vector is not cleared.

**To Reproduce**
1. Send a voice marker with metadata "ABC"
2. Call `stop_audio()`
3. Send another voice marker with metadata "DEF"
4. When `on_voice_track_marker()` callback is called by `DEF` marker, `event.track_meta` is `ABC`. One marker was inserted after `stop_audio()`, but `track_meta` size is 2, so `event.track_meta` contains metadata from "deleted" marker.

**Expected behavior**
All markers metadata should be deteled when `stop_audio()` is called.
The second marker in the example should have `event.track_meta` = "DEF" instead of "ABC"

**System Details:**
 - Linux rpi4 6.1.0-rpi6-rpi-v8 SMP PREEMPT Debian 1:6.1.58-1+rpt2 (2023-10-27) aarch64 GNU/Linux
 - Desktop Discord Client

**Suggested solution**
Adding `this->track_meta.clear()` at `/src/dpp/discordvoiceclient.cpp` at line 695.

Neko-Life:
## Changes in this PR
- Close #1124 
- Avoid hardcoded size check
- Faster erase of outbuf when skipping marker
- Tested on Musicat -alpha, no problem so fat

## Code change checklist

### DIFF
--- a/src/dpp/discordvoiceclient.cpp
+++ b/src/dpp/discordvoiceclient.cpp
@@ -1174,14 +1174,20 @@ discord_voice_client& discord_voice_client::skip_to_next_marker() {
 				});
 
 		if (i != outbuf.end()) {
-			if (i != outbuf.begin()) {
-				/* Skip queued packets up to found marker */
-				outbuf.erase(outbuf.begin(), i-1);
-			}
+			/* Skip queued packets until including found marker */
+			outbuf.erase(outbuf.begin(), i+1);
 		} else {
 			/* No market found, skip the whole queue */
 			outbuf.clear();
 		}
+	}
+
+	if (tracks > 0) {
+		tracks--;
+	}
+
+	if (!track_meta.empty()) {
+		track_meta.erase(track_meta.begin());
 	}
 
 	return *this;
--- a/src/dpp/discordvoiceclient.cpp
+++ b/src/dpp/discordvoiceclient.cpp
@@ -1171,7 +1171,7 @@ discord_voice_client& discord_voice_client::skip_to_next_marker() {
 		/* Find the first marker to skip to */
 		auto i = std::find_if(outbuf.begin(), outbuf.end(), [](const voice_out_packet &v){
 				return v.packet.size() == sizeof(uint16_t) && (*((uint16_t*)(v.packet.data()))) == AUDIO_TRACK_MARKER;
-				});
+			 });
 
 		if (i != outbuf.end()) {
 			/* Skip queued packets until including found marker */
--- a/src/dpp/discordvoiceclient.cpp
+++ b/src/dpp/discordvoiceclient.cpp
@@ -849,7 +849,7 @@ void discord_voice_client::write_ready()
 		std::lock_guard<std::mutex> lock(this->stream_mutex);
 		if (!this->paused && outbuf.size()) {
 			type = send_audio_type;
-			if (outbuf[0].packet.size() == 2 && (*((uint16_t*)(outbuf[0].packet.data()))) == AUDIO_TRACK_MARKER) {
+			if (outbuf[0].packet.size() == sizeof(uint16_t) && (*((uint16_t*)(outbuf[0].packet.data()))) == AUDIO_TRACK_MARKER) {
 				outbuf.erase(outbuf.begin());
 				track_marker_found = true;
 				if (tracks > 0) {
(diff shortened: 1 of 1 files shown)


src/dpp/discordvoiceclient.cpp
```diff
@@ -1165,20 +1167,29 @@ uint32_t discord_voice_client::get_tracks_remaining() {
...
-		outbuf.erase(outbuf.begin());
-	}
-	if (outbuf.size()) {
-		/* Remove the actual track marker out of the buffer */
-		outbuf.erase(outbuf.begin());
+	if (!outbuf.empty()) {
+		/* Find the first marker to skip to */
+		auto i = std::find_if(outbuf.begin(), outbuf.end(), [](const voice_out_packet &v){
+				return v.packet.size() == sizeof(uint16_t) && (*((uint16_t*)(v.packet.data()))) == AUDIO_TRACK_MARKER;
+				});
```
Jaskowicz1, line 1174: I'd just un-indent this `});` by one tab (so it's inline with `std::find_if`) so it's better looking :)

Jaskowicz1 (changes requested):

src/dpp/discordvoiceclient.cpp
Neko-Life: done

Mishura4 (approved):
👍

src/dpp/discordvoiceclient.cpp
```diff
@@ -1165,20 +1167,29 @@ uint32_t discord_voice_client::get_tracks_remaining() {
...
-	while (!outbuf.empty() && outbuf[0].packet.size() != sizeof(uint16_t) && (*((uint16_t*)(outbuf[0].packet.data()))) != AUDIO_TRACK_MARKER) {
-		outbuf.erase(outbuf.begin());
-	}
-	if (outbuf.size()) {
-		/* Remove the actual track marker out of the buffer */
-		outbuf.erase(outbuf.begin());
+	if (!outbuf.empty()) {
+		/* Find the first marker to skip to */
+		auto i = std::find_if(outbuf.begin(), outbuf.end(), [](const voice_out_packet &v){
+				return v.packet.size() == sizeof(uint16_t) && (*((uint16_t*)(v.packet.data()))) == AUDIO_TRACK_MARKER;
```
Mishura4, line 1173: Actually it looks like we still have indentation issues with this lambda here, there is an extra tab

src/dpp/discordvoiceclient.cpp
```diff
@@ -1165,20 +1167,29 @@ uint32_t discord_voice_client::get_tracks_remaining() {
...
+			 });
```
Mishura4, line 1174: And here an extra tab + space

Mishura4 (changes requested):

src/dpp/discordvoiceclient.cpp
Neko-Life: it's default vim indent behavior with correct shiftwidth and tabstop, maybe we need to create .clang-format file to make formatting easier across different editor

src/dpp/discordvoiceclient.cpp
Neko-Life: yes i added the space, fixed

Mishura4 (approved):
👍 LGTM ship it

Mishura4:
@Jaskowicz1

src/dpp/discordvoiceclient.cpp
Mishura4: Use a real IDE like Visual Studio catsuit

Jaskowicz1 (approved):

src/dpp/discordvoiceclient.cpp
Neko-Life: i mean u can just export the built-in formatting config of VS into an actual .clang-format file r8? surely such "full featured real IDE" like VS can do a simple export image image image

PR #1149, opened 2024-05-13
fix: remove redundancy with constexpr variables

0xBLAHAJ:
`constexpr` variables are evaluated at compile time, rendering the `inline` and `const` keywords useless since they don't persist in memory
this is only beneficial for file size as `constexpr` takes priority over other keywords (so probably the lowest priority pull request yet)

## Code change checklist


Mishura4:
Constexpr variables are not implicitly inline, `inline` is not redundant on variables. Functions are, but not variables.

Try including the header in two translation units without the header, you'll get multiple definition errors.

braindigitalis:
These changes wouldnt be made here anyway; the unicode_emojis.h is auto generated so any changes made directly to the file will be dropped.
Perhaps we should have an action that auto identifies and flags changes to auto-generated files where there hasn't also been a change to the generator script that creates it?

Jaskowicz1:
> These changes wouldnt be made here anyway; the unicode_emojis.h is auto generated so any changes made directly to the file will be dropped. Perhaps we should have an action that auto identifies and flags changes to auto-generated files where there hasn't also been a change to the generator script that creates it?

Sounds like a good idea!

Mishura4:
No, constexpr variables *are* variables, you can take their address. What ASM the compuler generates has nothing to do with this.

As i said, try including the header without the inline in two translation units.

braindigitalis:
but have you tried it, in dpp?

also, did you read the bit about this entire file being **auto generated**? Running cmake with all the correct dev depenencies will overwrite your changes.

0xBLAHAJ:
> No, constexpr variables _are_ variables, you can take their address. What ASM the compiler generates has nothing to do with this and **especially** with optimizations enabled.

Not sure what you mean by 'what ASM the compiler generates has nothing to do with this', this is about how compilers handle `inline`, here it's redundant because of the `constexpr` keyword
`constexpr` variables *can* have a place in memory but this isn't always the case, it actually is only the case if the variable in question can not *always* be evaluated at compile time (e.g. if you try to reference it's address, it obviously needs to place it somewhere in runtime memory thanks to [ASLR](https://en.wikipedia.org/wiki/Address_space_layout_randomization))
image
devenv_J05JOpoIWl
(compiled with -Od on MSVC)

> As i said, try including the header without the inline in two translation units. Header guards also have nothing to do with this.

You mentioned multiple definition errors which are non-existent if you have `#pragma once` at the beginning of the file (`unicode_emoji.h` has it) because it indicates the compiler should only include that header file once (hence the name)
image
`constexpr` variables are also just not source of multiple definitions
image

> but have you tried it, in dpp? 

compiled w/ x64-clang-debug on a win32 machine, did not break anything because it simply does not change how clang (or any compiler for that matter) reads the file

Mishura4:
> Not sure what you mean by 'what ASM the compiler generates has nothing to do with this', this is about how compilers handle inline, here it's redundant because of the constexpr keyword

**No, it's not**. `inline` in c++ is about allowing the linker to merge multiple identical definitions into one, it's not about the compiler replacing a variable indirection with its value.

Again. Try including the file in 2 translation units, without the inline, in one project. You'll see the multiple definition errors.

PR #1166, opened 2024-06-10
QoL fix: Enhance the entitlement stuff with data from the actual api docs now there are some

braindigitalis:
Adds missing fields from the docs and no longer assumes guild and user are both `owner_id`
 
## Code change checklist


braindigitalis:
**NOTE** 

This is a breaking change, as it changes the fields directly within dpp::entitlement. The old fields were insufficient for the reality of the API.

Jaskowicz1 (approved):

PR #1169, opened 2024-06-13
feat: added warnings for certain events that require an intent

Jaskowicz1:
This PR adds warnings for some events (guild member events and presence update) if the user does not have the specific intent required.

As these events do not fire from Discord if the intent wasn't specified, I could not use the `set_warning_callback` feature.

This is to try and inform users about what went wrong and how they can fix it, to try be more user friendly as D++ is currently silent about it (see [link a](https://discord.com/channels/825407338755653642/825411707521728512/1235314587868594309) and [link b](https://discord.com/channels/825407338755653642/825411707521728512/1250187664632840293)).

## Code change checklist

### DIFF
--- a/src/dpp/cluster.cpp
+++ b/src/dpp/cluster.cpp
@@ -171,6 +171,22 @@ void cluster::start(bool return_after) {
 		this->terminating.wait(thread_lock);
 	};
 
+	if (on_guild_member_add && !(intents & dpp::i_guild_members)) {
+		log(ll_warning, "You have attached an event to cluster::on_guild_member_add() but have not specified the privileged intent dpp::i_guild_members. This event will not fire.");
+	}
+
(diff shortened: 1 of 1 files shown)


braindigitalis (changes requested):
what about if the user hooks the event after starting the cluster?
also, while looking at this do we make the one for messages less spammy too? basically make them all emit once?

Jaskowicz1:
> what about if the user hooks the event after starting the cluster?

Couldn't come up with a solution to that, that didn't involve constantly checking every second if the events have been bound to, when I was writing this (and I still don't have a solution).

> also, while looking at this do we make the one for messages less spammy too? basically make them all emit once?

Either that or add something like `bot.suppress_warnings(true);` to prevent warning logs for this (for people who do know what they're doing)? I think being spammy drills in the idea that they're doing something wrong to people who might not understand.

braindigitalis (approved):

PR #1177, opened 2024-06-21
Use string views

BowDown097:
This will make the library easier to use for std::string_view users as a LOT of explicit std::string creations are needed atm. There should be some performance improvements as well (less allocations, blablabla). Tested with my own bot which is pretty complex and works 100%.

This is completely backwards compatible in my testing, with the exception of implicit conversions. For example, within the library itself, there were some assignments of a utility::icon to a string that would implicitly create a utility::iconhash - now it does not, and does not compile without explicitly creating the iconhash. With my own bot this happened with std::smatch.

There's also some new stuff / changes outside of just changing std::strings to std::string_views:
- A new utility function for maps, dpp::utility::emplace_or_assign, which essentially is insert_or_assign but with the in-place construction that emplace offers.
- All non-floating point instances of std::stoX that would take in a std::string_view have been changed to use std::from_chars.
- 2 new types, dpp::membuf and dpp::imemstream, in stringops.h, which allow input streams over a raw character array, necessary for std::string_view. These are not compiled if std::ispanstream is available, which was added in C++23, as it is far superior. All instances of std::istringstream that would take in a std::string_view have been replaced to use one or the other, depending on what's available. (P.S. should these be exported with DPP_EXPORT?)

## Code change checklist

### DIFF
--- a/include/dpp/stringops.h
+++ b/include/dpp/stringops.h
@@ -31,7 +31,37 @@
 #include <iostream>
 #include <charconv>
 
+#ifdef __cpp_lib_spanstream
+#include <spanstream>
+#endif
+
 namespace dpp {
+
+#ifndef __cpp_lib_spanstream
+/**
+ * @brief streambuf for imemstream
+ * @note From https://stackoverflow.com/a/13059195
+ */
+struct membuf : std::streambuf
+{
+    membuf(const char* base, size_t size)
+    {
+        char* p(const_cast<char*>(base));
+        this->setg(p, p, p + size);
+    }
+};
+
+/**
+ * @brief An input stream over a raw character array. Superseded by std::ispanstream in C++23
+ * @note From https://stackoverflow.com/a/13059195
+ */
+struct imemstream : virtual membuf, std::istream
+{
+    imemstream(const char* base, size_t size)
+        : membuf(base, size), std::istream(static_cast<std::streambuf*>(this)) {}
+};
+#endif
+
 /**
  * @brief Convert a string to lowercase using tolower()
  * 
@@ -41,9 +71,23 @@ namespace dpp {
  */
 template <typename T> std::basic_string<T> lowercase(const std::basic_string<T>& s)
 {
-	std::basic_string<T> s2 = s;
+    std::basic_string<T> s2(s);
 	std::transform(s2.begin(), s2.end(), s2.begin(), tolower);
 	return s2;
+}
+
+/**
+ * @brief Convert a string to lowercase using tolower()
+ *
+ * @tparam T type of string
+ * @param s String_view to lowercase
+ * @return std::basic_string<T> lowercased string
+ */
+template <typename T> std::basic_string<T> lowercase(std::basic_string_view<T> s)
+{
+    std::basic_string<T> s2(s);
+    std::transform(s2.begin(), s2.end(), s2.begin(), tolower);
+    return s2;
 }
 
 /**
@@ -55,9 +99,23 @@ template <typename T> std::basic_string<T> lowercase(const std::basic_string<T>&
  */
 template <typename T> std::basic_string<T> uppercase(const std::basic_string<T>& s)
 {
-	std::basic_string<T> s2 = s;
+    std::basic_string<T> s2(s);
 	std::transform(s2.begin(), s2.end(), s2.begin(), toupper);
 	return s2;
+}
+
+/**
+ * @brief Convert a string to uppercase using toupper()
+ *
+ * @tparam T type of string
+ * @param s String_view to uppercase
+ * @return std::basic_string<T> uppercased string
+ */
+template <typename T> std::basic_string<T> uppercase(std::basic_string_view<T> s)
+{
+    std::basic_string<T> s2(s);
+    std::transform(s2.begin(), s2.end(), s2.begin(), toupper);
+    return s2;
 }
 
 /**
@@ -66,10 +124,11 @@ template <typename T> std::basic_string<T> uppercase(const std::basic_string<T>&
  * @param s String to trim
  * @return std::string trimmed string
  */
-inline std::string rtrim(std::string s)
+inline std::string rtrim(std::string_view s)
 {
-	s.erase(s.find_last_not_of(" \t\n\r\f\v") + 1);
-	return s;
+    std::string s_cpy(s);
+    s_cpy.erase(s_cpy.find_last_not_of(" \t\n\r\f\v") + 1);
+    return s_cpy;
 }
 
 /**
@@ -78,10 +137,11 @@ inline std::string rtrim(std::string s)
  * @param s string to trim
  * @return std::string trimmed string
  */
-inline std::string ltrim(std::string s)
+inline std::string ltrim(std::string_view s)
 {
-	s.erase(0, s.find_first_not_of(" \t\n\r\f\v"));
-	return s;
+    std::string s_cpy(s);
+    s_cpy.erase(0, s_cpy.find_first_not_of(" \t\n\r\f\v"));
+    return s_cpy;
 }
 
 /**
@@ -90,7 +150,7 @@ inline std::string ltrim(std::string s)
  * @param s string to trim 
  * @return std::string trimmed string
  */
-inline std::string trim(std::string s)
+inline std::string trim(std::string_view s)
 {
 	return ltrim(rtrim(s));
 }
@@ -120,11 +180,19 @@ template<class T> std::string comma(T value)
  * @param f Numeric base, e.g. `std::dec` or `std::hex`
  * @return T Returned numeric value
  */
-template <typename T> T from_string(const std::string &s, std::ios_base & (*f)(std::ios_base&))
+template <typename T> T from_string(std::string_view s, std::ios_base & (*f)(std::ios_base&))
 {
 	T t;
-	std::istringstream iss(s);
-	iss >> f, iss >> t;
+
+#ifdef __cpp_lib_spanstream
+    std::span<const char> span(s.begin(), s.end());
+    std::ispanstream iss(span);
+    iss >> f, iss >> t;
+#else
+    imemstream ims(s.data(), s.size());
+    ims >> f, ims >> t;
+#endif
+
 	return t;
 }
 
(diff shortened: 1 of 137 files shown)


Jaskowicz1 (changes requested):
Thanks for the contribution!

Please correct your indentation to comply with the [Coding Style Standards](https://dpp.dev/coding-standards.html). You're using spaces, not tabs.

include/dpp/stringops.h
```diff
@@ -78,10 +137,11 @@ inline std::string rtrim(std::string s)
  * @param s string to trim
  * @return std::string trimmed string
  */
-inline std::string ltrim(std::string s)
+inline std::string ltrim(std::string_view s)
```
Jaskowicz1, line 140: Ditto the comment from rtrim.

include/dpp/stringops.h
```diff
@@ -55,21 +99,36 @@ template <typename T> std::basic_string<T> lowercase(const std::basic_string<T>&
...
+}
+
 /**
  * @brief trim from end of string (right)
  * 
  * @param s String to trim
  * @return std::string trimmed string
  */
-inline std::string rtrim(std::string s)
+inline std::string rtrim(std::string_view s)
```
Jaskowicz1, line 127: I don't agree with this change. You're now having to create a new string which adds more bytes (as you've now got the bytes from `s` + the bytes from `s_cpy`), which now adds 16 more bytes every time you call this function. This should be left as std::string.

Jaskowicz1 (changes requested):

braindigitalis:
I'm not sure why every instance of a string parameter, plus various strings in objects must all be changed to string views. this is a huge change which is impossible to review and impossible to say what side effects it has as it changes literally every function. 
how can we say this doesn't break things, the unit tests or one test bot don't scratch the surface for such a far reaching change.

it's better to concentrate on one subsystem at a time doing a gradual refactor than what looks like global search and replace.

string view is not a magic bullet.

braindigitalis (changes requested):
far too big to review sorry

src/dpp/bignum.cpp
```diff
@@ -47,11 +47,11 @@ struct openssl_bignum {
 	}
 };
 
-bignumber::bignumber(const std::string& number_string) : ssl_bn(std::make_shared<openssl_bignum>()) {
+bignumber::bignumber(std::string_view number_string) : ssl_bn(std::make_shared<openssl_bignum>()) {
 	if (dpp::lowercase(number_string.substr(0, 2)) == "0x") {
-		BN_hex2bn(&ssl_bn->bn, number_string.substr(2, number_string.length() - 2).c_str());
+        BN_hex2bn(&ssl_bn->bn, std::string(number_string.substr(2)).c_str());
```
Mishura4, line 52: If you construct a std::string anyways then what's the point? You now have an extra allocation when someone passes an std::string

src/dpp/cluster.cpp
```diff
@@ -321,8 +321,8 @@ json error_response(const std::string& message, http_request_completion_t& rv)
 	return j;
 }
 
-void cluster::post_rest(const std::string &endpoint, const std::string &major_parameters, const std::string &parameters, http_method method, const std::string &postdata, json_encode_t callback, const std::string &filename, const std::string &filecontent, const std::string &filemimetype, const std::string &protocol) {
-	rest->post_request(std::make_unique<http_request>(endpoint + (!major_parameters.empty() ? "/" : "") + major_parameters, parameters, [endpoint, callback](http_request_completion_t rv) {
+void cluster::post_rest(std::string_view endpoint, std::string_view major_parameters, std::string_view parameters, http_method method, std::string_view postdata, json_encode_t callback, std::string_view filename, std::string_view filecontent, std::string_view filemimetype, std::string_view protocol) {
+    rest->post_request(std::make_unique<http_request>(std::string(endpoint) + (!major_parameters.empty() ? "/" : "") + std::string(major_parameters), parameters, [callback](http_request_completion_t rv) {
```
Mishura4, line 325: Same thing, what's the point here of changing the parameter

src/dpp/cluster.cpp
```diff
@@ -349,7 +349,7 @@ void cluster::post_rest_multipart(const std::string &endpoint, const std::string
 		file_mimetypes.push_back(data.mimetype);
 	}
 
-	rest->post_request(std::make_unique<http_request>(endpoint + (!major_parameters.empty() ? "/" : "") + major_parameters, parameters, [endpoint, callback](http_request_completion_t rv) {
+    rest->post_request(std::make_unique<http_request>(std::string(endpoint) + (!major_parameters.empty() ? "/" : "") + std::string(major_parameters), parameters, [callback](http_request_completion_t rv) {
```
Mishura4, line 352: Again, what's the point

src/dpp/cluster/entitlement.cpp
```diff
@@ -62,15 +62,15 @@ void cluster::entitlements_get(snowflake user_id, const std::vector<snowflake>&
 
 	j["exclude_ended"] = exclude_ended;
 
-	rest_request_list<entitlement>(this, API_PATH "/applications", me.id.str(), "entitlements", m_get, j, callback);
+    rest_request_list<entitlement>(this, API_PATH "/applications", me.id.str(), "entitlements", m_get, std::string(j), callback);
```
Mishura4, line 65: Again what's the point™

I'll stop commenting that because there are tens of these

src/dpp/commandhandler.cpp
```diff
@@ -81,13 +81,15 @@ commandhandler& commandhandler::add_prefix(const std::string &prefix)
...
-commandhandler& commandhandler::add_command(const std::string &command, const parameter_registration_t &parameters, command_handler handler, const std::string &description, snowflake guild_id)
+commandhandler& commandhandler::add_command(std::string_view command, const parameter_registration_t &parameters, command_handler handler, std::string_view description, snowflake guild_id)
 {
 	command_info_t i;
 	i.func = std::move(handler);
 	i.guild_id = guild_id;
 	i.parameters = parameters;
-	commands[lowercase(command)] = i;
+
+    std::string command_lower = lowercase(command);
```
Mishura4, line 91: This is good, could have been a separate PR

src/dpp/discordclient.cpp
```diff
@@ -289,7 +289,12 @@ bool discord_client::handle_frame(const std::string &buffer)
 				j = json::parse(data);
 			}
 			catch (const std::exception &e) {
-				log(dpp::ll_error, "discord_client::handle_frame(JSON): " + std::string(e.what()) + " [" + data + "]");
+                std::string log_msg = "discord_client::handle_frame(JSON): ";
```
Mishura4, line 292: This is fine, also should have been a separate PR

src/dpp/dispatcher.cpp
```diff
@@ -267,11 +265,7 @@ command_value interaction_create_t::get_parameter(const std::string& name) const
 	return {};
 }
 
-voice_receive_t::voice_receive_t(discord_client* client, const std::string& raw, discord_voice_client* vc, snowflake _user_id, const uint8_t* pcm, size_t length) : event_dispatch_t(client, raw), voice_client(vc), user_id(_user_id) {
-	reassign(vc, _user_id, pcm, length);
-}
-
-voice_receive_t::voice_receive_t(discord_client* client, std::string&& raw, discord_voice_client* vc, snowflake _user_id, const uint8_t* pcm, size_t length) : event_dispatch_t(client, std::move(raw)), voice_client(vc), user_id(_user_id) {
```
Mishura4, line 274: Why? This was good for performance, what it needed was add std::move to the usage of `raw`. With your change you have an extra allocation

src/dpp/dns.cpp
```diff
@@ -39,19 +39,20 @@ namespace dpp
...
 
-	const dns_cache_entry* resolve_hostname(const std::string& hostname, const std::string& port)
+    const dns_cache_entry* resolve_hostname(std::string_view hostname, std::string_view port)
 	{
 		addrinfo hints, *addrs;
 		dns_cache_t::const_iterator iter;
 		time_t now = time(nullptr);
 		int error;
 		bool exists = false;
+        std::string hostname_str(hostname);
```
Mishura4, line 49: 😭

src/dpp/httpsclient.cpp
```diff
@@ -113,29 +119,32 @@ multipart_content https_client::build_multipart(const std::string &json, const s
 	}
 }
 
-const std::string https_client::get_header(std::string header_name) const {
-	std::transform(header_name.begin(), header_name.end(), header_name.begin(), [](unsigned char c){
+const std::string https_client::get_header(std::string_view header_name) const {
+    std::string header_name_str(header_name);
```
Mishura4, line 123: Extra alloc here again

src/dpp/httpsclient.cpp
```diff
@@ -342,11 +351,13 @@ http_connect_info https_client::get_host_info(std::string url) {
...
 	if (colon_pos != std::string::npos) {
-		hci.hostname = url.substr(0, colon_pos);
-		hci.port = atoi(url.substr(colon_pos + 1, url.length()).c_str());
-		if (hci.port == 0) {
-			hci.port = 80;
-		}
+        hci.hostname = url.substr(0, colon_pos);
+
+        std::string_view port_str = url.substr(colon_pos + 1, url.length());
+        auto [_, err] = std::from_chars(port_str.data(), port_str.data() + port_str.size(), hci.port);
```
Mishura4, line 357: This is good

include/dpp/stringops.h
```diff
@@ -120,11 +180,19 @@ template<class T> std::string comma(T value)
...
  * @return T Returned numeric value
  */
-template <typename T> T from_string(const std::string &s, std::ios_base & (*f)(std::ios_base&))
+template <typename T> T from_string(std::string_view s, std::ios_base & (*f)(std::ios_base&))
 {
 	T t;
-	std::istringstream iss(s);
-	iss >> f, iss >> t;
+
+#ifdef __cpp_lib_spanstream
```
Mishura4, line 187: This will be harder to maintain for very little benefit

Mishura4 (changes requested):
So overall yes there are places that could use string_view, but blanket replacing *all of them* has no benefit and is even a downgrade. If someone passes a string which you convert to sv and then construct again a string, what's the benefit exactly?

We also don't need the c++23-specific features, keep in mind this is code we have to use and improve and, and making a change to something that used c++23 if it's enabled and doesn't if not means we have to change and test both, which isn't tenable.

There are some positive changes in there but those could be in separate PRs that would be easier to review and test.

Also, be mindful of intendation changes.

PR #1178, opened 2024-06-21
Use string_view for message object ctors + setters

BowDown097:
This makes the methods more flexible with types of input, most notably with string_views themselves so the user doesn't have to create a temporary std::string.

## Code change checklist


Mishura4 (approved):
LGTM 👍

braindigitalis (approved):

PR #1193, opened 2024-07-08
fix: unit tests are now built statically when BUILD_SHARED_LIBS is off

Jaskowicz1:
This PR fixes the unit tests not being statically built when `BUILD_SHARED_LIBS` is off (which caused linker errors).


library/CMakeLists.txt
```diff
@@ -111,8 +111,6 @@ if(NOT BUILD_SHARED_LIBS)
 	endif()
 endif()
 
-include("${CMAKE_CURRENT_SOURCE_DIR}/../cmake/colour.cmake")
```
braindigitalis, line 114: how come this got taken out?

braindigitalis (approved):

library/CMakeLists.txt
Jaskowicz1: It already exists in another file, so don't need to include it here anymore (as the main CMakeLists.txt doesn't use it)

PR #1198, opened 2024-07-13
ci: allow runners to use mold, increased concurrency to 4

Jaskowicz1:
This PR attempts to speed up the action runners by allowing them to use mold and also changing the concurrency to 3 instead of 2 (github action runners were updated to 4 cores a while ago).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,17 +41,17 @@ jobs:
         # GitHub hosted runners on Azure
         # arm7hf is a self-hosted docker-based runner at Brainbox.cc. Raspberry Pi 4, 8gb 4-core with NEON
         cfg:
-          - { arch: 'amd64', concurrency: 2, os: ubuntu-20.04, package: clang-10, cpp-version: clang++-10, cmake-flags: '', cpack: 'no', ctest: 'no' }
-          - { arch: 'amd64', concurrency: 2, os: ubuntu-22.04, package: clang-11, cpp-version: clang++-11, cmake-flags: '', cpack: 'no', ctest: 'no' }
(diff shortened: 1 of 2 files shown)


braindigitalis:
can we bump up the concurrency on gh runners to 4 then instead of 3? do we see any improvement?

Jaskowicz1:
> can we bump up the concurrency on gh runners to 4 then instead of 3? do we see any improvement?

That was done in the recent push :)

Seems that we barely get an improvement across the board. A couple of runners are faster and a couple of them aren't any different.

Mishura4:
I mean mold will only affect linking which is a fraction of the time the runners spend on D++
mold really shines when LTO is enabled

Jaskowicz1:
@braindigitalis Ignore last comment about "little results".

CodeQL is now ~3 minutes faster (comparing https://github.com/brainboxdotcc/DPP/actions/runs/9921615282/job/27409774387?pr=1197 with https://github.com/brainboxdotcc/DPP/actions/runs/9921876515/job/27410330135?pr=1198).

The Action Runners now builds on average 30 seconds to 2 minutes faster (seems like Clang15, g++13, g++14, and g++15, were nearly 2 minutes faster now).

Self hosted runner arm7hf didn't benefit at all, same with self hosted runner arm64.

OSX didn't have any change to concurrency, nor linker. Neither did Windows.

Jaskowicz1:
> I mean mold will only affect linking which is a fraction of the time the runners spend on D++ mold really shines when LTO is enabled

true true, the reason it's dropped a good bit is the concurrency being upped to 4.

Jaskowicz1:
CodeQL is now down another minute! We're seeing 17 mins now!

braindigitalis (approved):

PR #1203, opened 2024-07-18
feat: std::format support for dpp::snowflake

ruslan-ilesik:
This pr adds std::format support for dpp::snowflake, if we do not have c++20 or have no include for <format> it will be automatically disabled for compilation.

Test case was implemented to check formatting. 

## Code change checklist


src/unittest/test.cpp
```diff
@@ -379,6 +379,30 @@ Markdown lol \\|\\|spoiler\\|\\| \\~\\~strikethrough\\~\\~ \\`small \\*code\\* b
 		set_test(SNOWFLAKE, success);
 	}
 
+        { // test snowflake: std::format support
+
+                #if __cplusplus >= 202002L
```
braindigitalis, line 384: DRY,  I'm concerned someone may edit one of these checks and not the other, can we have it purely in one place please, maybe used to set our own define called DPP_HAS_FORMAT

braindigitalis (changes requested):

include/dpp/snowflake.h
```diff
@@ -281,3 +281,32 @@ struct std::hash<dpp::snowflake>
 		return std::hash<uint64_t>{}(s.value);
 	}
 };
+
+#if __cplusplus >= 202002L
+//fix for https://github.com/llvm/llvm-project/issues/77773 and
+// apple support on clang 14 https://developer.apple.com/xcode/cpp/
+#if (defined(__cpp_lib_format) || \
```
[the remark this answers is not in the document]
ruslan-ilesik: Do we have runners with them to check?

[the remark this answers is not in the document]
braindigitalis: of course but they don't run unit tests

src/unittest/test.cpp
ruslan-ilesik: Done

include/dpp/snowflake.h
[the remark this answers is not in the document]
ruslan-ilesik: done

CMakeLists.txt
```diff
@@ -115,3 +115,48 @@ else()
...
+
+	)
+	set(CMAKE_REQUIRED_FLAGS ${temp_flags})
+	if(HAS_FORMAT)
+		message("-- INFO: Found std::format support")
+		target_compile_definitions(dpp PUBLIC DPP_HAS_FORMAT)
+	else ()
+		message("-- INFO: std::format support not found")
+	endif()
+endif ()
```
Jaskowicz1, line 162: Add an empty line after this please!

include/dpp/snowflake.h
```diff
@@ -26,6 +26,11 @@
 #include <cstdint>
 #include <type_traits>
 
+#ifdef DPP_HAS_FORMAT
+#include <format>
+#endif
+
+
```
Jaskowicz1, line 33: Double empty line not needed

include/dpp/snowflake.h
```diff
@@ -281,3 +286,23 @@ struct std::hash<dpp::snowflake>
...
+        constexpr typename TP::iterator parse(TP& ctx) {
+                return ctx.begin();
+        }
+
+        template<class TF>
+        typename TF::iterator format(const dpp::snowflake& snowflake, TF& ctx) const {
+                return std::format_to(ctx.out(), "{}", snowflake.str());
+        }
+};
+#endif //DPP_HAS_FORMAT
```
Jaskowicz1, line 308: Empty line after this too.

Jaskowicz1 (changes requested):

Henonicks:
tested this and got
`snowflake: std::format support                               PASS`
at the end

ruslan-ilesik:
So with last commit i have introduced usage of Cmake feature called [configure_file](https://cmake.org/cmake/help/v3.0/command/configure_file.html) which allows us autogenerate headers with some definitions when build dpp, this way we can remove need to define DPP_HAS_FORMAT as it will be defined in it when compiling it.

 In theory same way we can eliminate need to specify -DDPP_CORO when compiling bot with dpp enabled in lib.

CMakeLists.txt
```diff
@@ -115,3 +115,53 @@ else()
...
+
+configure_file(
+		${CMAKE_CURRENT_SOURCE_DIR}/include/dpp/autoconfig.h.in
+		${CMAKE_CURRENT_SOURCE_DIR}/include/dpp/autoconfig.h
+		@ONLY
+)
```
Jaskowicz1, line 167: Add a blank line after this.

Jaskowicz1 (changes requested):

braindigitalis (changes requested):
theres a massive problem with this!
If cmake detects that C++20 is available it will always enable std::format. This means the user is **forced to use -std=c++20** for all their bots, even where they dont want to use C++20. This fundamentally breaks compatibility.
Nothing tells the user they must use C++20 anywhere, and furthermore, if they have a need for C++17 they simply cannot ever use C++17 any more.

ruslan-ilesik:
If lib is build with c++20 then ye, bots will be required to use c++20, I can easily fix it by wrapping cmake define into check of c++ version.

Jaskowicz1 (approved):
I'm happy with it but need to see what @braindigitalis thinks about the `autoconfig.h` idea.

ruslan-ilesik:
I have tested changes locally on my wsl2 (ubuntu) and tests passed, also compiling bot with -DPP_FORMATTERS and without had required behavior

library/CMakeLists.txt
```diff
@@ -351,7 +355,7 @@ if (DPP_BUILD_TEST)
 			set (testsrc "")
 			file(GLOB testsrc "${modules_dir}/${testname}/*.cpp")
 			add_executable(${testname} ${testsrc})
-			if (DPP_CORO)
+			if (DPP_CORO OR DPP_FORMATTERS)
```
braindigitalis, line 358: these same changes need applying to library-vcpkg/CMakeLists.txt

braindigitalis (changes requested):

braindigitalis (approved):

PR #1204, opened 2024-07-18
ci: Added a clang++15 osx job, fixed clang++14 job.

Jaskowicz1:
This PR changes our OSX runner to have `clang++14` and `clang++15` jobs, instead of only `clang++14`. This PR also fixes the `clang++14` job, where it was actually `clang++15`.

The `clang++15` job is using Xcode 15.3 to ensure that `std::format` is tested (as versions below don't have `std::format`), however, it should be noted that we **NEED** to enforce checks for `std::format` features to prevent Xcode 15.2 and lower versions breaking, otherwise we will be heavily breaking OSX support.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -70,7 +70,7 @@ jobs:
 
       - name: Setup mold
         if: ${{ matrix.cfg.mold == 'yes' }}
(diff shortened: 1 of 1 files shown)


braindigitalis (approved):

braindigitalis:
lgtm!

PR #1208, opened 2024-07-22
feat: timeout is now 20 seconds. Users can now change the timeout length

Jaskowicz1:
This PR changes the default request timeout to be 20 seconds. This PR also adds a function to `dpp::cluster` to allow users to change the timeout length if they wish.

## Code change checklist

### DIFF
--- a/include/dpp/cluster.h
+++ b/include/dpp/cluster.h
@@ -203,6 +203,11 @@ public:
 	std::condition_variable terminating;
 
 	/**
+	 * @brief The time (in seconds) that a request is allowed to take.
+	 */
+	uint16_t request_timeout = 10;
+
+	/**
 	 * @brief Constructor for creating a cluster. All but the token are optional.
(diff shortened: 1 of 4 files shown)


include/dpp/cluster.h
```diff
@@ -420,6 +425,15 @@ class DPP_EXPORT cluster {
 	 */
 	const shard_list& get_shards();
 
+	/**
+	 * @brief Sets the request timeout.
+	 *
+	 * @param timeout The length of time (in seconds) that requests are allowed to take. Default: 20.
```
braindigitalis, line 431: are you sure? the value above is 10 :)

braindigitalis (changes requested):

include/dpp/cluster.h
Jaskowicz1: :facepalm: im silly lol

braindigitalis (approved):

PR #1229, opened 2024-08-25
fix: Improve compatiblity with socket

TracerDS:
This PR improves compatiblity with other projects that DPP is included into.

`#include <dpp/win32_safe_warnings.h>` is to prevent dll-interface warnings while compiling the main application.

`#ifndef SOCKET` is used to prevent possible incompatiblities with projects that already handle that case (such as network libraries). This way the socket will still be defined but only if it wasnt earlier.

### DIFF
--- a/include/dpp/export.h
+++ b/include/dpp/export.h
@@ -46,7 +46,6 @@
 	#else
 		/* Including the library */
 		#ifdef _WIN32
-			#include <dpp/win32_safe_warnings.h>
 			#define DPP_EXPORT __declspec(dllimport)
 		#else
 			#define DPP_EXPORT
@@ -116,11 +115,7 @@ extern bool DPP_EXPORT validate_configuration();
 
 }
 
-#ifndef _WIN32
-	#ifndef SOCKET
-		#define SOCKET int
-	#endif
-#else
+#ifdef _WIN32
 	#ifndef NOMINMAX
 		#define NOMINMAX
 	#endif
--- a/include/dpp/export.h
+++ b/include/dpp/export.h
@@ -46,6 +46,7 @@
 	#else
 		/* Including the library */
 		#ifdef _WIN32
+			#include <dpp/win32_safe_warnings.h>
 			#define DPP_EXPORT __declspec(dllimport)
 		#else
 			#define DPP_EXPORT
@@ -116,11 +117,13 @@ extern bool DPP_EXPORT validate_configuration();
 }
 
 #ifndef _WIN32
-	#define SOCKET int
+	#ifndef SOCKET
+		#define SOCKET int
+	#endif
 #else
-  #ifndef NOMINMAX
-	  #define NOMINMAX
-  #endif
+	#ifndef NOMINMAX
+		#define NOMINMAX
+	#endif
 
 	#include <WinSock2.h>
 #endif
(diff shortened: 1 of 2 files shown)


include/dpp/export.h
```diff
@@ -46,6 +46,7 @@
 	#else
 		/* Including the library */
 		#ifdef _WIN32
+			#include <dpp/win32_safe_warnings.h>
```
braindigitalis, line 49: this shouldnt be a public include, its up to the end user to decide what warnings to disable

include/dpp/export.h
```diff
@@ -116,11 +117,13 @@ extern bool DPP_EXPORT validate_configuration();
 }
 
 #ifndef _WIN32
-	#define SOCKET int
+	#ifndef SOCKET
```
braindigitalis, line 120: this is a bad idea, what youre saying is if the user defines SOCKET, then they break dpp and wsock2.h silently by changing the type in the DPP headers. (as opposed to it complaining about macro redefinition)

braindigitalis (changes requested):

include/dpp/export.h
Mishura4: Agreed. If you want to change the define, it could be a type alias instead something like `dpp::socket`

TracerDS: That would be the best way but didnt know where to put it, so I just edited the macro

include/dpp/export.h
TracerDS: Its either that or a ton of dll warnings flooding the output.
Maybe instead of that, DPP should have a preprocessor check for a macro:```cpp
#ifdef DPP_IGNORE_DLL_WARNINGS
    #include <dpp/win32_safe_warnings.h>
#endif
```

Mishura4: Or add `/wd4251` to the command line, as is the right way to do this :)

TracerDS: Designing code that doesnt throw warnings is the best way here

Mishura4: Which including a `pragma disable` is not.

And unfortunately C4251 is hell to solve. We would need to ditch the standard library entirely for function parameters. woohoo :)

TracerDS: But its the code's responsibility to deal with warnings, not user's one.
I mean, the warnings are easy but annoying to solve. There should be a macro to do it automatically

braindigitalis: > Which including a `pragma disable` is not.
> 
> And unfortunately C4251 is hell to solve. We would need to ditch the standard library entirely for function parameters. woohoo :)

not just function parameters but ANY USE of standard library in any class or struct.

braindigitalis: > But its the code's responsibility to deal with warnings, not user's one. I mean, the warnings are easy but annoying to solve. There should be a macro to do it automatically

absolutely an end user thing to silence. the standard template we support does do this.

Mishura4: It's the library's responsibility to fix the root of warnings yes. We won't blanket disable warnings for everyone using the library though, that could cause issues for other libraries or user code.

If you think C4251 is easy to solve the root cause of though, feel free to try. ;)

TracerDS: I understand

Mishura4 (approved):

braindigitalis (approved):

PR #1259, opened 2024-10-05
Feat/force disable dave on stage channel

Neko-Life:
## Changes in this PR

- Always force disable dave when joining stage channel

## Code change checklist

### DIFF
--- a/src/dpp/discordclient.cpp
+++ b/src/dpp/discordclient.cpp
@@ -622,6 +622,15 @@ discord_client& discord_client::connect_voice(snowflake guild_id, snowflake chan
 			return *this;
 		}
 	}
+
+	if (enable_dave) {
+		/* Disable dave if we're connecting to stage channel */
+		auto *c = find_channel(channel_id);
+		if (c != nullptr && c->is_stage_channel()) {
+			enable_dave = false;
+		}
+	}
+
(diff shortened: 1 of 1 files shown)


Jaskowicz1:
Can Dave not be used for stage channels? If it can, I don't think this PR is a good idea

braindigitalis:
> Can Dave not be used for stage channels? If it can, I don't think this PR is a good idea

stage channels are just voice channels.

if Dave isn't supported by a voice chat it automatically downgrades... besides we have no choice, supporting Dave will become mandatory some point next year

src/dpp/discordclient.cpp
```diff
@@ -622,6 +622,15 @@ discord_client& discord_client::connect_voice(snowflake guild_id, snowflake chan
 			return *this;
 		}
 	}
+
+	if (enable_dave) {
+		/* Disable dave if we're connecting to stage channel */
+		auto *c = find_channel(channel_id);
```
braindigitalis, line 628: I don't think this is a good idea, there's nothing special about stage channels that needs Dave disabled it's just the Dave pr isn't finished.

regardless this assumes channel cache is enabled which isn't going to work if you disable cache.

braindigitalis (changes requested):

PR #1271, opened 2024-10-12
Add namespace comment

fffelix-huang:
No functional change.

### DIFF
--- a/include/dpp/json_fwd.h
+++ b/include/dpp/json_fwd.h
@@ -29,4 +29,4 @@ namespace dpp {
 
 using json = nlohmann::json;
 
-}
+} // namespace dpp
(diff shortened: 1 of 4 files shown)


braindigitalis:
are there no other namespaces missing a similar comment? would be more likely to approve the PR if there was more substance to it. Please check for other missing namespace closing comments and add them into this PR :)

fffelix-huang:
There are several other missing namespace closing comment (Example: `include/dpp/unicode_emoji.h`). However since they are auto-generated I didn't add them.

braindigitalis:
> There are several other missing namespace closing comment (Example: `include/dpp/unicode_emoji.h`). However since they are auto-generated I didn't add them.

perhaps you can track down where the generators do this, and fix the generators? :)

Jaskowicz1:
why are we letting this PR be a thing btw, we discussed last year we did not care about this.

braindigitalis:
You are right, we did agree that, so it isnt needed.
On the opposite side of the coin there are a ton of these in src/dpp/dave that could likely be removed.

PR #1272, opened 2024-10-12
fix: vcpkg is always built with voice

braindigitalis:
fix vcpkg build to enable voice at all times. it brings in the deps it needs so does not need to check theyre there.

## Code change checklist


Jaskowicz1 (approved):
Whilst the PR itself looks great, will this not massively increase the length of compiling D++ for VCPKG users or will it still be less as we removed sodium?

PR #1280, opened 2024-10-14
refactor(breaking): Deprecate sync calls

braindigitalis:
Deprecate sync calls, from now on they will be auto generated with DPP_DEPRECATED on them advising users to use coro.

This bypasses sync depreciation warnings in the unit tests by using the template directly instead. I don't want to touch this with a ten foot pole right now as this is a whole world of fragile callback hell. This can be addressed in the future maybe, but for now, this isnt directly end user facing code and isnt going to cause anyone any hassle.

## Code change checklist


Mishura4 (approved):

ruslan-ilesik:
Imho, we should fully remove them when we will be c++20 lib by default. Because coro can directly replace sync calls without blocking and callback hell

Jaskowicz1 (approved):

PR #1308, opened 2024-10-23
feat: conan 2.x recipe

braindigitalis:
This cannot be used to create a pr on conan-center until after the next release (10.0.34) as it relies on this release's tarball to function, which is not yet in place.

## Code change checklist


braindigitalis:
@MikeRavenelle i would appreciate some feedback on this one if you have time please!

conanfile.py
```diff
@@ -0,0 +1,97 @@
...
+from conan.tools.scm import Git, Version
+from conan.tools.files import update_conandata, download, unzip
+
+required_conan_version = ">=2.0"
+
+class DPPConan(ConanFile):
+    name = "dpp"
+    version = "10.0.33"
+    default_user = "brainboxdotcc"
+    default_channel = "testing"
```
memsharded, line 15: Default user and channel do not have any effect in Conan 2.
The way to hardcode it in Conan 2 is using ``user="xxx"`` and ``channel="yyyy"``. But in many cases, specially opensource like ConanCenter, it is fine to go without user/channel.

conanfile.py
```diff
@@ -0,0 +1,97 @@
...
+    license = "https://github.com/brainboxdotcc/DPP/blob/master/LICENSE"
```
memsharded, line 16: If the license is not a official MIT/GPL/etc open-source license, maybe consider using the SPDX standard license specification, something like ``license = "DocumentRef-README:LicenseRef....``

conanfile.py
```diff
@@ -0,0 +1,97 @@
...
+    settings = "os", "compiler", "build_type", "arch"
+    options = {"shared": [True, False], "fPIC": [True, False]}
+    default_options = {"shared": True, "fPIC": True}
+
+    @property
+    def _min_cppstd(self):
+        return 17
+
+    @property
+    def _compilers_minimum_version(self):
```
memsharded, line 29: This is very often limited mostly by the cppstd minimum, not the compiler version itself, and it was a way to achieve some checks in Conan 1. 

Conan 2 mandates/strongly recommend usage of ``cppstd`` in profiles, so checking the min_cppstd might be enough in many cases.

conanfile.py
```diff
@@ -0,0 +1,97 @@
...
+        if self.options.shared:
+            self.options.rm_safe("fPIC")
+
+    def layout(self):
+        cmake_layout(self)
+
+    def export(self):
+        git = Git(self, self.recipe_folder)
+        scm_url, scm_commit = git.get_url_and_commit()
+        update_conandata(self, {"sources": {"commit": scm_commit, "url": scm_url}})
```
memsharded, line 67: we added a more standard way: https://docs.conan.io/2/examples/tools/scm/git/capture_scm/git_capture_scm.html, with ``git.coordinates_to_conandata()``. That will put the data in a "standard" location in the conandata.yml that can be leveraged by future tools (already drafter in workspace feature)

library/CMakeLists.txt
```diff
@@ -67,31 +67,40 @@ else()
...
 endif()
 
-
+if (CONAN_EXPORTED)
+	message("-- INFO: ${Green}Conan detected${ColourReset}... finding packages...")
+	find_package(OpenSSL REQUIRED COMPONENTS SSL Crypto)
+	find_package(ZLIB REQUIRED)
+	find_package(Opus)
+	find_package(Threads REQUIRED)
+	find_package(Git QUIET)
```
memsharded, line 76: Git and Threads are not managed by Conan, shouldn't that be "common" and independent of CONAN_EXPORTED?

memsharded (commented):
Looking good, just a few possible ideas for improvements or simplifications.

library/CMakeLists.txt
braindigitalis: they are only here because if we run the build through conan, we wanted to group the find_package together for easy reference, it makes no difference if they are here or elsewhere apart from ease of development for us

braindigitalis:
> Looking good, just a few possible ideas for improvements or simplifications.

Thanks, I've made these changes!

When it comes to preparing this for a PR on conan center, how do i go about this? Is there a page on the docs saying how to submit a new package to conan center as i must have missed it.

memsharded:
The main docs are in https://github.com/conan-io/conan-center-index/blob/master/docs/adding_packages/README.md

We are just in the process to moving to a Conan 2-only pipeline, a matter of a couple of weeks. Probably you might want to wait until this happens before submitting it.

I'd say that the recipe is quite good, it shouldn't be very difficult to add it, just some minor things ,like using a ``conandata.yml`` with the data, probably removing the ``export`` capture of the ``scm`` commit.

braindigitalis:
Also after completing a build of D++ through conan i get the following deprecation warnings about several of the dependencies, is this something to be concerned about, if the dependencies never updated to Conan 2.x and you're deprecating Conan 1.x soon won't this break all D++'s dependencies via Conan?

```
dpp/10.0.33: Package '70ff0fab11b405566a3aa83746595e4f42384a03' created
dpp/10.0.33: Full package reference: dpp/10.0.33#ceb89c69974c5d471818a4f592cfe2c5:70ff0fab11b405566a3aa83746595e4f42384a03#13a094550d605b980d4c81686fa3496e
dpp/10.0.33: Package folder /home/brain/.conan2/p/b/dpp1c4bd0de58a7b/p
WARN: deprecated: Usage of deprecated Conan 1.X features that will be removed in Conan 2.X:
WARN: deprecated:     'cpp_info.names' used in: opus/1.4, openssl/3.1.2, zlib/1.3
WARN: deprecated:     'cpp_info.build_modules' used in: openssl/3.1.2
WARN: deprecated:     'env_info' used in: openssl/3.1.2
```

memsharded:
> Also after completing a build of D++ through conan i get the following deprecation warnings about several of the dependencies, is this something to be concerned about, if the dependencies never updated to Conan 2.x and you're deprecating Conan 1.x soon won't this break all D++'s dependencies via Conan?

Don't worry too much. We will make sure to remove the legacy parts in ConanCenter recipes before completely deprecating them in the Conan client (that would raise an error). Hopefully this happens relatively fast after we move to the Conan 2-only pipeline, but it is true that there is also no intent to be aggressive about this, because the backlog there is very large, no need to overwhelm the system and reviewers with "stylistic/deprecation" only changes.

Jaskowicz1 (approved):

braindigitalis:
@memsharded having difficulty getting this to work on Conan center. everything I've done seems to conflict with what they are saying there e.g. must still support Conan 1.x? and ancient cmake from before 2019 seems to be a thing on their ci!

https://github.com/conan-io/conan-center-index/pull/25745

I would appreciate some input if you have time please 

also is Conan center different from Conan itself with different policies???

jcar87:
> @memsharded having difficulty getting this to work on Conan center. everything I've done seems to conflict with what they are saying there e.g. must still support Conan 1.x? and ancient cmake from before 2019 seems to be a thing on their ci!

Regarding the CMake version - when Conan 2 was designed, it was done in consultation with industry experts - and Conan 2 itself aims to be compatible (where possible, for the broadest compatibility) with CMake 3.15
https://github.com/conan-io/tribe/blob/main/design/004-tools-cmake.md


Please bear in mind that Conan Center has thousands of users, and what is "an ancient CMake from 2019" for some users, may actually be the version of CMake that users in enterprise environments are actually using - and we aim to support those as well. That's why the default in Conan Center CI is 3.15 in at least one of the platforms, so that we can also ensure broad compatibility with the recipes. It *does not mean* that all recipes need to be compatible with 3.15 - the recipes can express a requirement on a higher CMake version for this, as @memsharded  mentions above.

braindigitalis:
> > @memsharded having difficulty getting this to work on Conan center. everything I've done seems to conflict with what they are saying there e.g. must still support Conan 1.x? and ancient cmake from before 2019 seems to be a thing on their ci!
> 
> Regarding the CMake version - when Conan 2 was designed, it was done in consultation with industry experts - and Conan 2 itself aims to be compatible (where possible, for the broadest compatibility) with CMake 3.15 [conan-io/tribe@`main`/design/004-tools-cmake.md](https://github.com/conan-io/tribe/blob/main/design/004-tools-cmake.md?rgh-link-date=2024-10-29T16%3A38%3A57Z)
> 
> Please bear in mind that Conan Center has thousands of users, and what is "an ancient CMake from 2019" for some users, may actually be the version of CMake that users in enterprise environments are actually using - and we aim to support those as well. That's why the default in Conan Center CI is 3.15 in at least one of the platforms, so that we can also ensure broad compatibility with the recipes. It _does not mean_ that all recipes need to be compatible with 3.15 - the recipes can express a requirement on a higher CMake version for this, as @memsharded mentions above.

we simply cannot support 3.15 in dpp. we make extensive use of 3.16 features in our makefiles, and do not support any compilers/toolchains with a cmake this old. Our oldest supported compiler is gcc 8 and msvc 2019, msvc 2019 comes with cmake 3.36 (iirc).

If supporting 3.15 is a requirement, we would simply not list on conan until this version is incremented, this library is not intended to run on systems that old.

memsharded:
> we simply cannot support 3.15 in dpp. we make extensive use of 3.16 features in our makefiles, and do not support any compilers/toolchains with a cmake this old. Our oldest supported compiler is gcc 8 and msvc 2019, msvc 2019 comes with cmake 3.36 (iirc).

No, but you don't have to! That is a requirement by Conan client, and a default base version for ConanCenter, but not mandatory at all.

You can very easily request and use any CMake version in ConanCenter, and you can do it with a simple ``tool_requires = "cmake/[>=3.27 <4]"`` in your recipe (or the minimum version you need). You don't even need to stick to 3.16, you can request any higher version and use in your CMakeLists.txt the newer CMake features you want.

So please, just add to your ``conanfile.py`` that ``tool_requires = "cmake/...``. Thanks for the feedback!

PR #1326, opened 2024-11-08
fix: two segmentation faults in cluster::guild_edit_member and cluster::tick_timers/cluster::shutdown

KTuff:
`cluster::guild_edit_member` accesses invalid memory in its callback if the guild_member object is deallocated before the `post_rest` makes a call to the callback. Capturing the object as copy fixes this issue.

`cluster::shutdown` does not clear the `next_timer` map which causes segfaults in `cluster::tick_timers` if the timers are not removed before shutdown. Clearing the map after all timers are deleted resolves this.

## Code change checklist

### DIFF
--- a/src/dpp/cluster.cpp
+++ b/src/dpp/cluster.cpp
@@ -319,6 +319,7 @@ void cluster::shutdown() {
 		delete t.second;
 	}
 	timer_list.clear();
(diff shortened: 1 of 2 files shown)


braindigitalis (approved):
looks good to me, thanks for the contribution!

PR #1328, opened 2024-11-09
Change to valid C++ namespace identifier

egorpugin:


braindigitalis:
this is not needed, we do not include or use this file.

PR #1334, opened 2024-11-13
feat(breaking): 10.1 - socket engines, refactor, remove dpp::sync

braindigitalis:
Socket engines are a high performance abstraction layer for socket events. The end goal of this project is to increase the performance of DPP's internals at scale while using many less threads for the same task. Where currently each shard cluster has its own socket loop, and every request has its own internal socket loop, this will be a thing of the past with one core socket loop per cluster managed by `epoll`, `kqueue` or `poll`.

The correct socket loop type is selected automatically by cmake.

This project will take a long time to finish (it has already been in progress for a while now) and will be a significant internal change. As far as a library user is concerned there will be minimal change to how the library works if any.

# TODO

- [x] fix error handling/exceptions
- [x] fix auto reconnect 
- [x]  fix background startup if bot.start(true)
- [x]  move all events into thread pool
- [x] voice websockets
- [x] fix weird issue where voice only works if Discord clients join after bot
- [x] investigate if the threads left in queues.cpp can be refactored
- [x] investigate spurious https request timeouts (handshake failure?)
- [x] fix plaintext http requests
- [x] thoroughly test kqueue (needs a freebsd vm or shell or help from someone with macos)
- [x] thoroughly test poll (on windows)
- [x] fix `dpp::sync`
- [x] Fix auto reconnect of shard websockets (+ resume)
- [x] Fix auto reconnect of voice websockets (+ resume)
- [x] Fix runaway cpu after reconnect
- [x] Check that https requests are being freed when done, in queues.cpp
- [x] Track down sometimes segfaults on destruct of cluster
- [x] Documentation (incl new working webhooks example)

## Code change checklist

- [x] ~~I have ensured that I did not break any existing API calls.~~ This is a breaking new version change.


RealTimeChris (approved):
Looks good.

PR #1352, opened 2024-12-18
fix: large https request body could not be sent

braindigitalis:
Large https request bodies such as a huge bulk command create with tons of localisations can't be sent. Essentially anything that needs more than one buffer write would fail. This fixes that, tidies up some sslclient stuff, adds a ton of comments, and adds a facility to turn on raw logging for an sslclient on a case by case basis by calling a method.

## Code change checklist


RealTimeChris (approved):
Looks good. Nice job.

PR #1354, opened 2024-12-19
fix: make sure write events go out as soon as they can

braindigitalis:
We must set WANT_WRITE whenever we append to the write buffer of an ssl_client based object. This is the proper fix for the issue patched by @Neko-Life where voice wasn't sending via socketengine events. If the opcode 4 was not sent soon enough, then the connection would not succeed, this seems oddly temporal now. Setting WANT_WRITE in the ssl_client::write() function ensures all writes occur as soon as they possibly can.

## Code change checklist


Neko-Life (approved):
良さそう！！！

PR #1358, opened 2025-01-14
fix: Compile issue if task or awaitable is used by its own

peti446:
std::condition_variable is used in the awaitable.h file, so it should be included in it.
If you only include dpp::taks or similar header files, it will cause a linker error as it will not be able to link against std::condition_variabl, normally you would include dpp.h or other files that include a lot so It's not really noticeable.

## Code change checklist

### DIFF
--- a/include/dpp/coro/awaitable.h
+++ b/include/dpp/coro/awaitable.h
@@ -39,6 +39,7 @@ struct awaitable_dummy {
 
 // Do not include <coroutine> as coro.h includes <experimental/coroutine> or <coroutine> depending on clang version
 #include <mutex>
+#include <condition_variable>
 #include <optional>
 #include <utility>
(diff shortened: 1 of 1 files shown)


braindigitalis:
thanks for the PR, it would need to go into dev branch (not master) if it is for the upcoming 10.1 release, or in the 10.0 branch if it is for the 10.0.x releases (this is most likely).

braindigitalis (approved):

braindigitalis:
i have reopened this PR and targeted it at `10.0`. The 10.1 `dev` branch already has this fix. Once the CI passes, I shall merge it. Thanks for your contribution!

peti446:
Thanks for taking care of the retarget seems like I forgot to do it last night ^^

PR #1362, opened 2025-01-20
Don't always run php for DPP_CORO

supsm:
If DPP_CORO is configured to be true, the make_struct.php will *always* be run, even if not on UNIX or if PHP isn't found. This PR moves the relevant execute_process to where the other php files are run.

## Code change checklist


supsm:
oops

PR #1380, opened 2025-02-19
fix: oauth2_install_params in current_application_get() could not be parsed

braindigitalis:
* Fix bug with parsing of oauth2_install_params in current_application_get()
* Remove autogenerated mlspp namespace.h (it is already in gitignore)
* Add unit test for current_application_get()
* Change logger errors in unit tests to red prefix

Note: Instead of a map with always two std::optional in application::integration_types_config, the map now only contains keyed values for non-empty integration types instead of using std::optional. e.g. if your bot only has a guild invite, the map will only contain one value keyed to ait_guild_install. This is simpler and matches the way other things in dpp work.

## Code change checklist

### DIFF
--- a/include/dpp/application.h
+++ b/include/dpp/application.h
@@ -216,7 +216,7 @@ public:
  * @brief Configuration object for an app installation
  */
 struct DPP_EXPORT integration_configuration {
-		std::optional<application_install_params> oauth2_install_params;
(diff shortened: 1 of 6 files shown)


Jaskowicz1 (approved):
Glad to see another Unit Test for this 🥳

PR #1385, opened 2025-02-21
fix: deny user ability to set certain headers

### LINKED ISSUE #1382: HTTP request bodies randomly gets cut off in Windows
**Describe the bug**
HTTP request bodies done with `dpp::cluster::request` randomly gets cut off in Windows.

**To Reproduce**
I am currently using the latest version of D++ obtained from `dl.dpp.dev`, version 10.1.0.

```cpp
#include <dpp/dpp.h>

#include <semaphore>
#include <iostream>
#include <cstdlib>
#include <utility>

static std::binary_semaphore g_sem{0};

int main() {
  const auto discord_token{std::getenv("BOT_TOKEN")};

  if (discord_token == nullptr) {
    std::cerr << "error: missing BOT_TOKEN environment variable" << std::endl;
    return 1;
  }

  dpp::cluster bot{discord_token};

  bot.start(dpp::start_type::st_return);

  dpp::http_headers headers{};

  headers.insert(std::pair("Authorization", "top.gg API token"));
  headers.insert(std::pair("Connection", "close"));
  headers.insert(std::pair("Content-Type", "application/json"));
  headers.insert(std::pair("User-Agent", "topgg (https://github.com/top-gg-community/cpp-sdk) D++"));

  bot.request("https://top.gg/api/bots/264811613708746752", dpp::http_method::m_get, [](const auto& response) {
    std::cout << "===BODY START===\n" << response.body << "\n===BODY END===\nSIZE: " << response.body.size() << std::endl;
    g_sem.release();
  }, "", "application/json", headers);

  g_sem.acquire();

  return 0;
}
```

**Expected behavior**
The HTTP request should've retrieved the whole body. This request sometimes only gives less body than expected, and the difference between requests is usually random.

For example, when I ran the code above, I got:
- 2,041 bytes
- 3,746 bytes (the entire body as expected)
- 3,746 bytes (the entire body as expected)
- 3,410 bytes

**Screenshots**
First iteration:
Image

Second and third iteration:
Image

No re-compilations were done between iterations, I simply reran the executable.

**System Details:**
 - OS: Windows 11, compiled with Microsoft Visual C++ under Visual Studio 2022
 - Discord Client: None -- I was using `dpp::cluster` purely to test `dpp::cluster::request`.

**Additional context**
I was testing the [Top.gg C++ SDK](https://github.com/top-gg-community/cpp-sdk) and needed `dpp::cluster::request` to send requests to the Top.gg API.

braindigitalis:
If the user tries to set multiple Content-Length, Host, or Connection HTTP headers, that conflict with those DPP sends, servers can exhibit non-deterministic behaviour. How they react is server specific, and can include closing the connection, returning the wrong page, etc.

We now filter these out.

Fixes #1382

## Code change checklist

### DIFF
--- a/src/dpp/httpsclient.cpp
+++ b/src/dpp/httpsclient.cpp
@@ -54,7 +54,7 @@ void https_client::connect()
 	std::string map_headers;
 	for (auto& [k,v] : request_headers) {
 		std::string lower_header = dpp::lowercase(k);
-		if (lower_header == "connection" && lower_header == "content-length" && lower_header == "host") {
+		if (lower_header == "connection" || lower_header == "content-length" || lower_header == "host") {
 			owner->log(ll_warning, "Detected unnecessary duplicate header '" + k + "' in HTTP request to '" + hostname + "', which has been discarded.");
 		}
 		map_headers += k + ": " + v + "\r\n";
--- a/src/dpp/httpsclient.cpp
+++ b/src/dpp/httpsclient.cpp
@@ -53,7 +53,10 @@ void https_client::connect()
 	state = HTTPS_HEADERS;
 	std::string map_headers;
 	for (auto& [k,v] : request_headers) {
-		map_headers += k + ": " + v + "\r\n";
+		std::string lower_header = dpp::lowercase(k);
+		if (lower_header != "connection" && lower_header != "content-length" && lower_header != "host") {
+			map_headers += k + ": " + v + "\r\n";
+		}
 	}
 
 	if (this->sfd != SOCKET_ERROR) {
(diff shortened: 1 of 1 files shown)


Mishura4 (commented):
Do we really want to silently swallow it? What about throwing an exception instead?

braindigitalis:
> Do we really want to silently swallow it? What about throwing an exception instead?

we cant throw an exception here that the user  can deal with, this is not within the user's ability to catch as it is not within their program flow.

also, swallowing the header is the least breaking change as it wont crash someones bot. This does the opposite, and fixes broken bots.

Jaskowicz1:
Tbh, I think a log to indicate that D++ has removed headers would be nice. Just to indicate to the user running the bot that they don't need to do this and we'll do it for them.

Mishura4 (changes requested):
https://github.com/brainboxdotcc/DPP/commit/0a9159c44c7b34c25fe0691b8605ed21f7acc970#commitcomment-152856605

Jaskowicz1 (approved):
Was gonna comment earlier about the missing continue but seems you've done it between then and now, nice nice

Mishura4 (approved):

PR #1398, opened 2025-03-04
feat: ssl context management

braindigitalis:
Instead of creating a new ssl context for every connection, we only need:

* one SSL context for all client connections
* one for each listening port, with the SSL private key and public key loaded into it

This saves on memory of having to keep copies of the context around, and also massively cuts down on cpu use. Creating an SSL context from scratch means loading the cert store and root certs, generating things, loading private and public keys from disk and more.

## Code change checklist


RealTimeChris (approved):
Looks good to me. Nicely done.

Neko-Life (approved):
lgtm

PR #1399, opened 2025-03-08
fix build process to enable coro in vcpkg

braindigitalis:
Coro would not build in vcpkg as the section to enable it was never copy and pasted over when we made it on by default. vcpkg never had coro.

## Code change checklist

### DIFF
--- a/library-vcpkg/CMakeLists.txt
+++ b/library-vcpkg/CMakeLists.txt
@@ -85,12 +85,84 @@ target_compile_options(
 
 target_compile_features(
 	"${LIB_NAME}" PUBLIC
-	"cxx_std_17" "cxx_constexpr" "cxx_auto_type"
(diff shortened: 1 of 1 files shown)


Jaskowicz1 (commented):
Should we not maybe put the coro stuff in it's own CMake file?

We're repeating stuff that we could potentially change, and if we don't copy it over to either one, the one we forget to update will obviously be missing stuff.

RealTimeChris (approved):
Looks good. As always lol.

braindigitalis:
> Should we not maybe put the coro stuff in it's own CMake file?
> 
> We're repeating stuff that we could potentially change, and if we don't copy it over to either one, the one we forget to update will obviously be missing stuff.

in the case of vcpkg it has its own cmakefile because things work differently there. Most stuff is hard coded to enabled. for example, the target names are also different.

Jaskowicz1 (approved):

PR #1400, opened 2025-03-13
feat: `import dpp;`

Mishura4:
proxy-image

There are still some things to iron out, but it works on clang, and almost works on MSVC...

## Code change checklist


Mishura4:
CONFLICTS? I just pulled the branch bruv

Also remind me to turn modules off by default in CMake before marking as ready

cmake/MSVCSTDModule.cmake
```diff
@@ -0,0 +1,27 @@
+cmake_minimum_required(VERSION 3.30)
```
braindigitalis, line 1: the latest Ubuntu seems to only have 3.22, is this an acceptable version on msvc?

include/dpp/appcommand.h
```diff
@@ -115,7 +121,7 @@ enum command_option_type : uint8_t {
  *
  * You can retrieve them with std::get().
  */
-typedef std::variant<std::monostate, std::string, int64_t, bool, snowflake, double> command_value;
+DPP_EXPORT typedef std::variant<std::monostate, std::string, int64_t, bool, snowflake, double> command_value;
```
braindigitalis, line 124: do we need to export declarations that have no bodies in a cpp file?

include/dpp/appcommand.h
```diff
@@ -185,13 +191,13 @@ struct DPP_EXPORT command_option_choice : public json_interface<command_option_c
  * @param j output json object
  * @param choice command_option_choice to be serialized
  */
-void to_json(nlohmann::json& j, const command_option_choice& choice);
+DPP_EXPORT void to_json(nlohmann::json& j, const command_option_choice& choice);
```
braindigitalis, line 194: I'm not sure these are supposed to be exported to the library user

braindigitalis (changes requested):
I know it's not part of this pr but I also think we should add a new CI entry for static link on Linux

ruslan-ilesik:
First of all. Thank you for your contributions. All stuff u do is just crazy 👍.

I have a question. U specified that it works on clang and almost on MVSC. But what about g++? Is there issues porting it there. Or does g++ 
 not support modules at the moment?

CMakeLists.txt
```diff
@@ -32,6 +32,8 @@ option(DPP_USE_EXTERNAL_JSON "Use an external installation of nlohmann::json" OF
 option(DPP_USE_PCH "Use precompiled headers to speed up compilation" OFF)
 option(AVX_TYPE "Force AVX type for speeding up audio mixing" OFF)
 option(DPP_TEST_VCPKG "Force VCPKG build without VCPKG installed (for development use only!)" OFF)
+option(DPP_MODULES "Support for C++20 modules" ON)
```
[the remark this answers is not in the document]
Mishura4: For sure yeah, that's just for my convenience while developing, I'll leave this unresolved so I don't forget to turn it off

cmake/MSVCSTDModule.cmake
Mishura4: I believe msvc ships with 3.30 yeah, although I can take it down to 3.28, 3.30 is `import std;` experimental support, I'll change it tomorrow

As for package managers, not my problem :kekw: but yeah that does mean OFF by default

include/dpp/appcommand.h
Mishura4: The `export` in modules work a bit differently than Window's `__declspec(dllexport)` in that it's about names rather than definitions, a little bit like `public` and `private` in classes, so yes we need to export the names we introduce to be used as part of the API

include/dpp/appcommand.h
Mishura4: Hmm I'll double check but if they're used by templates or if we want conversions from json etc they need to be visible which means exported

Mishura4:
@braindigitalis for sure I can add that

@ilesik I believe g++14 has them experimental and g++15 (unreleased) has a lot of work in them, I'll test it as well, but I'm ready to ship it just for clang and msvc if it doesn't work on g++

Mishura4:
PR on hold, will reopen later

PR #1401, opened 2025-03-17
Fix: timers stop firing if the reactor has no sockets, only happens on bots that have little activity

braindigitalis:
Fix: timers stop firing if the reactor has no sockets, only happens on bots that have little activity

~~Improvement (unrelated): Bump minimum TLS version to 1.3~~ - rolled back - voice websockets do not support TLS 1.3!

## Code change checklist

### DIFF
--- a/src/dpp/socketengines/epoll.cpp
+++ b/src/dpp/socketengines/epoll.cpp
@@ -77,14 +77,6 @@ struct DPP_EXPORT socket_engine_epoll : public socket_engine_base {
 
 	void process_events() final {
 		const int sleep_length = 1000;
-		if (sockets == 0) {
-			/* epoll_wait() on empty set waits forever (or until another thread inserts a socket into the set.
-			 * We can't trust that this is going to happen, and it may deadlock the cluster, so in the event the
-			 * set is empty, we wait a millisecond (so it isn't a busy-wait) and return.
-			 */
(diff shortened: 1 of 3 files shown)


src/dpp/socketengines/epoll.cpp
```diff
@@ -77,14 +77,6 @@ struct DPP_EXPORT socket_engine_epoll : public socket_engine_base {
 
 	void process_events() final {
 		const int sleep_length = 1000;
-		if (sockets == 0) {
-			/* epoll_wait() on empty set waits forever (or until another thread inserts a socket into the set.
```
braindigitalis, line 81: note: this is an INCORRECT assumption and the cause of the bug. epoll_wait will not wait forever if called with an empty set.

src/dpp/socketengines/kqueue.cpp
```diff
@@ -63,6 +63,7 @@ struct DPP_EXPORT socket_engine_kqueue : public socket_engine_base {
 
 		int i = kevent(kqueue_handle, nullptr, 0, ke_list.data(), static_cast<int>(ke_list.size()), &ts);
 		if (i < 0) {
+			prune();
```
braindigitalis, line 66: in the event kevent fails, we still need to trigger timers

RealTimeChris (approved):
Looks good.

PR #1404, opened 2025-03-24
require c++17

erics118:
clang 16 may bug out on macos if it is not explicitly provided

added a line to require c++17 before linking std::filesystem

## Code change checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -17,7 +17,8 @@
 #
 
 cmake_minimum_required (VERSION 3.16)
-
+set(CMAKE_CXX_STANDARD 17)
+ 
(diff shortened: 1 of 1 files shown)


Mishura4 (dismissed):
Just add it at the top directly after `project` IMO

braindigitalis (changes requested):
we need to use the standard which is set by the detections not hard code to C++17, this could cause conflicts later down the line where the order of these definitions is assumed.

Mishura4:
> > Just add it at the top directly after `project` IMO
> 
> it shouldnt be hard coded to C++17. Our default is C++20 now unless its not supported or DPP_NO_CORO is set. The value for the standard should reflect this setting.

It's not, if you set it to 17 and then to 20, it'll be 20

PR #1419, opened 2025-04-18
feat: Add avatar_url override for webhook struct, resolves #1416

### LINKED ISSUE #1416: No way to arbitrarily override avatar_url for webhooks
**Is your feature request related to a problem? Please describe.**
The discord API for execute_webhook allows you to override the webhook's avatar with an arbitrary URL, but I'm unable to do this with DPP, because no API is provided.

**Describe the solution you'd like**
An avatar_url property inside the webhook struct, or an argument in the execute_webhook method

**Describe alternatives you've considered**
The webhook.avatar property does override the avatar_url field, but it's using an iconhash, which doesn't allow for me to specify urls from other websites.

**Additional context**
cluster/webhook.cppL82

Enovale:
This is a fix for #1416, adding a new property to the webhook struct to allow for custom avatar urls.

If an avatar_url is set, it will be used instead of the custom uploaded avatar.

## Code change checklist

### DIFF
--- a/include/dpp/webhook.h
+++ b/include/dpp/webhook.h
@@ -114,6 +114,7 @@ public:
 	/**
 	 * @brief The default avatar of the webhook.
 	 *
+	 * @note This value will not have any effect when `avatar_url` is set, they are mutually exclusive.
 	 * @note This may be empty.
 	 */
 	utility::iconhash avatar;
@@ -121,6 +122,7 @@ public:
 	/**
 	 * @brief Avatar URL to use instead of the default if it is set.
 	 *
+	 * @note This will override `avatar` if it is set, they are mutually exclusive.
 	 * @note This may be empty.
 	 */
 	std::string avatar_url;
--- a/include/dpp/webhook.h
+++ b/include/dpp/webhook.h
@@ -119,6 +119,13 @@ public:
 	utility::iconhash avatar;
 
 	/**
+	 * @brief Avatar URL to use instead of the default if it is set.
+	 *
+	 * @note This may be empty.
+	 */
+	std::string avatar_url;
+
+	/**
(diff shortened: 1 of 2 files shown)


src/dpp/cluster/webhook.cpp
```diff
@@ -73,12 +73,14 @@ void cluster::execute_webhook(const class webhook &wh, const struct message& m,
 		{"with_components", true},
 	});
 	std::string body;
-	if (!thread_name.empty() || !wh.avatar.to_string().empty() || !wh.name.empty()) { // only use json::parse if thread_name is set
+	if (!thread_name.empty() || !wh.avatar.to_string().empty() || !wh.avatar_url.empty() || !wh.name.empty()) { // only use json::parse if thread_name is set
```
braindigitalis, line 76: it might be nice to tidy up the comment here while we're in this bit of code. It was not correct before, and certaintly isnt now :-)

include/dpp/webhook.h
```diff
@@ -118,6 +118,13 @@ class DPP_EXPORT webhook : public managed, public json_interface<webhook> {
 	 */
 	utility::iconhash avatar;
 
+	/**
+	 * @brief Avatar URL to use instead of the default if it is set.
+	 *
+	 * @note This may be empty.
```
braindigitalis, line 124: you might want to add something in here that says if it is set it supersedes the avatar variable. These two are now mutually exclusive.

braindigitalis (changes requested):

braindigitalis:
just a couple of minor comment changes, and this is good to merge!
Thanks for your contribution!

braindigitalis (approved):

Mishura4 (approved):

PR #1435, opened 2025-05-01
fix(breaking): voice connect doesnt break after a shard reconnect

braindigitalis:
If you attempt to connect a voice connection after the shard has resumed or reconnected, the pointer is invalid. We shouldn't be storing pointers to shards, this was fixed everywhere else for 10.1 and this was *forgor*.

Reported by magistermaks on Discord.

**Breaking change as it adds a mandatory extra parameter for the cluster to the connect_member_voice function for looking up the shard** 

## Code change checklist


RealTimeChris (approved):
Looks good. Nice work.

PR #1451, opened 2025-07-01
Change SOVERSION from full version to major version

benthetechguy:
I'm packaging this library for Debian right now. The only change I need to make is that `SOVERSION` is supposed to just be the major version. The library will still be built with the version in `VERSION`, but with a symlink to the version in `SOVERSION`.

## Code change checklist

### DIFF
--- a/library/CMakeLists.txt
+++ b/library/CMakeLists.txt
@@ -264,7 +264,7 @@ foreach (fullmodname ${subdirlist})
 	set_target_properties(${modname} 
 		PROPERTIES
 		VERSION ${CMAKE_PROJECT_VERSION}
-		SOVERSION ${CMAKE_PROJECT_VERSION}
+		SOVERSION ${DPP_VERSION_MAJOR}
 		POSITION_INDEPENDENT_CODE true
 	)
 	target_include_directories(${modname} PUBLIC 


braindigitalis:
what exactly does this mean? I don't get what you are saying when you say you need to change SOVERSION?

braindigitalis:
I see what you've done but that's wrong and will introduce incompatibility. 10.1.is nothing like 10.0.

the 10 part is discords API major version, when discord release a new major API, we bump to match once we support it. this is maybe once every few years 

the bump from 10.0 to 10.1 introduced c++20 as default instead of c++17, coroutines by default and a new socket reactor model. if people thought they were one and the same, there would be chaos, they are far from binary compatible.

benthetechguy:
The SOVERSION is supposed to keep track of major changes to the ABI. So for example, if I build my bot for D++ 10.1.2, I can send the binary to my friend and they can run it with the D++ 10.1.1 their package manager installed and it should still run fine. However, if they had version 9.x.x it would not work because major things have changed.

Would you say that 10.1.x and 10.0.x would be ABI incompatible? If so, then yes, it does make sense to use major.minor instead of just major.

craigpipeandpiper:
yeah, t hat isnt going to work, we dont ever retain binary abi compatibility even between minor versions, because it means we cant even adjust templates. C++ is finicky this way compared to C.
You wouldnt be able to guarantee binary abi compat between any versions of D++.

PR #1459, opened 2025-07-02
CI: remove vs2019 from release and actions

braindigitalis:
vs2019 is no longer supported on GitHub actions so we are removing it from our CI flow. This will end official support and release zips for vs2019, it may still work for you on this compiler but we don't officially test on it any longer. Now would be a good time to update to vs2022 if you can.

## Code change checklist

- [ ] My code follows the [coding style guide](https://dpp.dev/coding-standards.html).


Mishura4 (approved):
LET'S GOOOOOOOOOOOOOOOOOOOOO

PR #1485, opened 2025-09-14
Fix D++ from stalling for a second before sending REST requests using poll.

voltflake:
Quite a while ago i noticed an issue that makes windows D++ builds delay sending messages, responses, and other REST requests with 1000ms delay. This PR fixes that.

Basically i added a dummy socket into poll implementation so that when poll() is running and waiting for timeout it can be interrupted by event of that dummy socket and refresh its file descriptors on demand.

To properly test this, increase poll timeout to 5000 ms and try simple ping-pong example bot. Before this change bot would hang for 5s and only then reply. Or slightly faster if heartbeat event occurs, which triggers poll activation. Issue persisted on all platforms when using default poll socket engine, both Linux and Windows.

Feel free to pinpoint in flaws that i may missed or request changes.

(they are as documented as they were before me)

### DIFF
--- a/include/dpp/cluster.h
+++ b/include/dpp/cluster.h
@@ -2375,9 +2375,20 @@ public:
 	 * @see https://discord.com/developers/docs/resources/channel#get-pinned-messages
 	 * @param channel_id Channel ID to get pins for
 	 * @param callback Function to call when the API call completes.
-	 * On success the callback will contain a dpp::message_map object in confirmation_callback_t::value. On failure, the value is undefined and confirmation_callback_t::is_error() method will return true. You can obtain full error details with confirmation_callback_t::get_error().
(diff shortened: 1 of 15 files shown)


voltflake:
It inserts that wakeup udp socket into poll fds only once in constructor. Purpose of that socket is to allow poll() to exit before timeout if needed. When other thread updates or registers new socket, and poll() is already running during that process waiting for socket events it can't exit that state until timeout reaches. In other socket engine implementations, like epoll(), early exits are possible and are used by lib. This additional wakeup socket basically adds this interrupt functionality to basic poll(), allowing it to exit early on demand.

Jaskowicz1:
I've approved runners and updated the name, as this is reproduceable on Linux.

braindigitalis (approved):

PR #1487, opened 2025-09-25
refactor: simplify run_once usage via default template parameter

Anohkka:
Defaulting `run_once`'s template parameter to `decltype([]{})` removes the need for its users to provide a unique type, since lambdas are guaranteed to have a unique type:

> [C++11: 5.1.2/3]: The type of the lambda-expression (which is also the type of the closure object) is a unique, unnamed non-union class type — called the closure type — whose properties are described below. This class type is not an aggregate (8.5.1). The closure type is declared in the smallest block scope, class scope, or namespace scope that contains the corresponding lambda-expression.

Example code:

```
#include "dpp/once.h"

#include <iostream>

int main() {
    for (int i = 0; i < 10; ++i) {
        if (dpp::run_once()) {
            std::cout << "This will print once\n";
        }
        if (dpp::run_once()) {
            std::cout << "This, too, will print once\n";
        }
    }
    return 0;
}
```

## Code change checklist

### DIFF
--- a/include/dpp/once.h
+++ b/include/dpp/once.h
@@ -2,7 +2,7 @@
  *
  * D++, A Lightweight C++ library for Discord
  *
- * Copyright 2022 Craig Edwards and D++ contributors 
+ * Copyright 2022 Craig Edwards and D++ contributors
  * (https://github.com/brainboxdotcc/DPP/graphs/contributors)
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


Mishura4:
Fwiw, "works as expected on godbolt" isn't enough for me, that's about as good as "works on my machine", I need a guarantee that it won't change by i.e. updating gcc from 14 to 15, the standard offers some guarantee save for compiler bugs.

After more digging I came to the conclusion that the standard *intends* to make this happen but with current wording doesn't, that's the CWG issue

PR #1504, opened 2025-10-21
fix: dpp::cluster::register_command with coroutine handler on windows

usyless:
register_command with dpp::task was not working for me on windows with msvc, this fixes that, and it makes sure to keep register_command working for non coroutines.

I have not tested the changes on anything other than windows, therefore I'm unsure if im sacraficing other platforms.

~~Also I'm not sure if this is the ideal fix, however it does also let clangd correctly determine the function in use when ctrl+click'ing the register_command, as previously it would just use the non-coro one for all usages of register_command for me. (It is inspired by how dpp::cluster::start_timer works, which works correctly for me)~~

As of the latest commit it now only allows void or dpp::task<void> function return signatures as well.

## Code change checklist

### DIFF
--- a/include/dpp/cluster.h
+++ b/include/dpp/cluster.h
@@ -587,7 +587,12 @@ public:
 	 * @return bool Returns `true` if the command was registered successfully, or `false` if
 	 * the command with the same name already exists
 	 */
-	bool register_command(const std::string& name, const slashcommand_handler_t handler);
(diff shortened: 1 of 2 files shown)


Mishura4:
Tested on WSL g++ 14, clang++ 21, confirmed to work, merging after CI!

PR #1525, opened 2025-12-17
fix: poll vote events and add missing poll intents

ProjectSky:
- Fix message_poll_vote_add/message_poll_vote_remove reading from root JSON instead of "d" object
- Fix message_poll_vote_remove checking wrong event handler (was checking on_message_poll_vote_add)
- Add i_guild_message_polls (1 << 24) intent for guild poll vote events
- Add i_direct_message_polls (1 << 25) intent for DM poll vote events
- Include poll intents in i_default_intents

### DIFF
--- a/include/dpp/intents.h
+++ b/include/dpp/intents.h
@@ -127,6 +127,16 @@ enum intents {
 	i_auto_moderation_execution	= (1 << 21),
 
 	/**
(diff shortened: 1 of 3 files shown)


Jaskowicz1 (approved):

Jaskowicz1:
Good catch!

PR #1530, opened 2025-12-27
feat: Add C++ modules support

mikomikotaishi:
This pull request adds support for C++20 modules.

In order to accomplish this, some constants were changed to using external linkage.

~~I have also updated nlohmann/json to version 3.12.0, as the older version declares some symbols as internal.~~ Updating nlohmann/json has been moved to #1596.

## Code change checklist


Mishura4 (changes requested):
Thanks for the PR! Good stuff. Just gonna prefer the approach with including dpp.h in the module for maintainability, as discussed in https://discord.com/channels/825407338755653642/825411104208977952/1454456223952011427

braindigitalis (changes requested):
some module related unit tests would be good. not repeating a thousand lines of code in dpp.cppm is an absolute must, as @Mishura4 also said.

mikomikotaishi:
I've looked at the `src/unittest/` files but I'm not particularly familiar with how you've set up the unit testing framework. Are you just looking to run those same tests under `import dpp;` rather than `#include <dpp/dpp.h>`?

braindigitalis:
I think that in this case you might want to make a new .cpp under the unittests folder which imports dpp instead of #including it. it can have some test functions that are exported and called via standard approaches from the main unittest file.

mikomikotaishi:
I think this should fix the errors we're encountering. It's with the build system configuration

braindigitalis:
you cant just do -DDPP_MODULES=ON for the entire CI matrix. Some CI's use older g++ (which we support) and some use C++17.

docpages/example_code/using_modules.cpp
```diff
@@ -0,0 +1,29 @@
...
+#include <cstdlib>
+
+import dpp;
+
+using dpp::cluster;
+using dpp::slashcommand;
+using dpp::start_type;
+
+int main() {
+	cluster bot(std::getenv("BOT_TOKEN"));
```
Jaskowicz1: We should avoid `std::getenv` as we don't use this anywhere else, just use `cluster bot("token");` or something.

docpages/example_programs/misc/using_modules.md
```diff
@@ -0,0 +1,9 @@
+\page using_modules Using D++ with C++ modules
+
+D++ is offered as a C++ module, which offers improved compile times over a traditional header.
+
+In order to enable support, you should be using C++20+ with any module-supporting compiler. To activate the feature, pass the `DPP_MODULES` flag to CMake. Ensure that the generated build system supports modules (for CMake, this is usually Ninja).
```
Jaskowicz1: Probably worth changing to `you need to use C++20 (or higher) with any module-supporting compiler`, using `should be` somewhat gives the impression that it's not mandatory.

docpages/example_code/using_modules.cpp
```diff
@@ -0,0 +1,29 @@
...
+		if (dpp::run_once<struct register_bot_commands>()) {
+			bot.global_command_create(
+				slashcommand("ping", "Ping pong!", bot.me.id)
+			);
+		}
+	});
+
+	bot.start(start_type::st_wait);
+	return 0;
+}
```
Jaskowicz1: Needs a new line after the end `}` (line 30 should be blank).

Jaskowicz1 (changes requested):

docpages/example_code/using_modules.cpp
mikomikotaishi: I copied this off the example on the root README but if we're removing it, then I'll remove it from the README too.
```cpp
#include <dpp/dpp.h>
#include <cstdlib>

int main() {
	dpp::cluster bot(std::getenv("BOT_TOKEN"));

	bot.on_slashcommand([](auto event) {
		if (event.command.get_command_name() == "ping") {
			event.reply("Pong!");
		}
	});

	bot.on_ready([&bot](auto event) {
		if (dpp::run_once<struct register_bot_commands>()) {
			bot.global_command_create(
				dpp::slashcommand("ping", "Ping pong!", bot.me.id)
			);
		}
	});

	bot.start(dpp::st_wait);
	return 0;
}
```
@braindigitalis any thoughts?

docpages/example_programs/misc/using_modules.md
mikomikotaishi: OK, makes sense

mikomikotaishi:
Then is it possible to disable the modules unit tests for C++17 CI tests?

mikomikotaishi:
@braindigitalis I think the modules CI should just be in a totally separate CI. It requires Ninja, as CMake can't build modules with Makefiles.

docpages/example_code/using_modules.cpp
mikomikotaishi: Fixed

braindigitalis:
unit tests only run in one g++, which is g++-12. This is the one which is packaged. we cant and dont run unit tests on them all, as the unit tests connect to discord and we cant do this concurrently.

docpages/example_code/using_modules.cpp
Jaskowicz1: Don't change the README example, we can do that in a different PR as that's not the scope of your PR :)

mikomikotaishi: OK, I removed `std::getenv()`.

.github/workflows/test-docs-examples.yml
```diff
@@ -35,17 +35,17 @@ jobs:
...
   
       - name: Generate CMake
-        run: mkdir build && cd build && cmake -DDPP_NO_VCPKG=ON -DAVX_TYPE=T_fallback -DDPP_CORO=ON -DCMAKE_BUILD_TYPE=Debug ..
+        run: mkdir build && cd build && cmake -G Ninja -DDPP_NO_VCPKG=ON -DDPP_MODULES=ON -DAVX_TYPE=T_fallback -DDPP_CORO=ON -DCMAKE_BUILD_TYPE=Debug ..
         env:
           CXX: g++-12
   
       - name: Build Project
-        run: cd build && make -j2 && sudo make install
+        run: cd build && ninja && sudo ninja install
```
Jaskowicz1: This doesn't specify the jobs limit anymore. Also I believe you don't need to do `ninja && sudo ninja install` as you have `-G Ninja`, but need confirmation from someone who knows cmake better than me.

Jaskowicz1 (changes requested):

Jaskowicz1:
Updated Title to match code standards

.github/workflows/test-docs-examples.yml
```diff
@@ -35,17 +35,17 @@ jobs:
...
   
       - name: Install apt packages
-        run: sudo sed -i 's/azure\.//' /etc/apt/sources.list && sudo apt-get update && sudo apt-get install -y g++-12 libopus-dev zlib1g-dev libmpg123-dev liboggz-dev cmake libfmt-dev libopusfile-dev
+        run: sudo sed -i 's/azure\.//' /etc/apt/sources.list && sudo apt-get update && sudo apt-get install -y g++-12 libopus-dev ninja-build zlib1g-dev libmpg123-dev liboggz-dev cmake libfmt-dev libopusfile-dev
   
       - name: Generate CMake
-        run: mkdir build && cd build && cmake -DDPP_NO_VCPKG=ON -DAVX_TYPE=T_fallback -DDPP_CORO=ON -DCMAKE_BUILD_TYPE=Debug ..
+        run: mkdir build && cd build && cmake -G Ninja -DDPP_NO_VCPKG=ON -DDPP_MODULES=ON -DAVX_TYPE=T_fallback -DDPP_CORO=ON -DCMAKE_BUILD_TYPE=Debug ..
         env:
           CXX: g++-12
```
Jaskowicz1: This needs to be bumped up to something like gcc14, CMake will fail scanning for modules otherwise (which it did in https://github.com/brainboxdotcc/DPP/actions/runs/20562247908/job/59070351884)

.github/workflows/test-docs-examples.yml
```diff
@@ -35,17 +35,17 @@ jobs:
...
   
       - name: Build Project
-        run: cd build && make -j2 && sudo make install
+        run: cd build && ninja && sudo ninja install
   
       - name: Test compile examples
-        run: cd docpages/example_code && mkdir build && cd build && cmake .. && make -j2
+        run: cd docpages/example_code && mkdir build && cd build && cmake -G Ninja .. && ninja
         env:
           CXX: g++-12
```
Jaskowicz1: Ditto line 43

src/unittest/test.cpp
```diff
@@ -334,6 +334,9 @@ Markdown lol \\|\\|spoiler\\|\\| \\~\\~strikethrough\\~\\~ \\`small \\*code\\* b
 			set_test(WEBHOOK, false);
 		}
 
+		// Test C++ module import
+		run_module_tests();
```
Jaskowicz1, line 339: This needs to be wrapped with something like `#ifdef DPP_MODULES` otherwise it'll attempt to run when modules aren't enabled (hence the CI fails, windows doesn't build tests seemingly so that's why windows is passing).

Jaskowicz1 (changes requested):

src/unittest/test.cpp
mikomikotaishi: I assume we have to pass this on CMake with target_compile_definitions() too?

Jaskowicz1: You already do https://github.com/brainboxdotcc/DPP/pull/1530/files#diff-9504fe9e5bb5a14156c9bda58abf4040e06b2a1a9debcea7d87b4bb45dfe8772R484

src/unittest/module_test.cpp
```diff
@@ -0,0 +1,124 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ ************************************************************************************/
+
+#include "test.h"
+
+import dpp;
```
Jaskowicz1, line 26: You need to wrap all this in the ifdef too, please make sure you're compiling before committing.

(Lines 25 to 124)

src/unittest/test.h
```diff
@@ -156,6 +156,8 @@ DPP_TEST(FORUM_CHANNEL_GET, "retrieve the created forum channel", tf_online);
 DPP_TEST(FORUM_CHANNEL_DELETE, "delete the created forum channel", tf_online);
 DPP_TEST(ERRORS, "Human readable error translation", tf_offline);
 DPP_TEST(INVALIDUTF8, "Invalid UTF-8 handling", tf_online);
+DPP_TEST(MODULE_IMPORT_BASIC, "C++ module import: basic types", tf_offline);
+DPP_TEST(MODULE_IMPORT_CORO, "C++ module import: coroutines", tf_offline);
```
Jaskowicz1, line 162: ifdef required

src/unittest/test.h
```diff
@@ -570,6 +572,11 @@ inline constexpr auto is_owner = [](auto &&user) noexcept {
 	return get_user_snowflake(user) == TEST_USER_ID;
 };
 
+/**
+ * @brief Main entry point for module tests
+ */
+void run_module_tests();
```
Jaskowicz1, line 582: ifdef required

Jaskowicz1 (changes requested):

src/unittest/module_test.cpp
mikomikotaishi: Apologies. I always run `cmake -S . -B build -G Ninja -DDPP_MODULES=ON -DDPP_BUILD_TEST=ON && cmake --build build` before I commit my changes.

mikomikotaishi:
Perhaps it would be better to do module tests on a later compiler version, like Clang 21 (which there already is a test for), which is known to support modules?

.github/workflows/test-docs-examples.yml
braindigitalis: i think this is important. lack of job limit will slow down jobs on github. do we really NEED to use ninja here?

mikomikotaishi: I restored the -j2 flags.

Also, yes we must use Ninja, because we are building one target which uses modules and Makefiles with CMake do not support modules.

Jaskowicz1:
> Perhaps it would be better to do module tests on a later compiler version, like Clang 21 (which there already is a test for), which is known to support modules?

Yeah we can add it to the CI workflow to allow some to do it, lets get this PR sorted for now as we're getting close to everything working 😄

.github/workflows/test-docs-examples.yml
```diff
@@ -35,17 +35,17 @@ jobs:
           submodules: recursive
   
       - name: Install apt packages
-        run: sudo sed -i 's/azure\.//' /etc/apt/sources.list && sudo apt-get update && sudo apt-get install -y g++-12 libopus-dev zlib1g-dev libmpg123-dev liboggz-dev cmake libfmt-dev libopusfile-dev
+        run: sudo sed -i 's/azure\.//' /etc/apt/sources.list && sudo apt-get update && sudo apt-get install -y g++-12 libopus-dev ninja-build zlib1g-dev libmpg123-dev liboggz-dev cmake libfmt-dev libopusfile-dev
```
Jaskowicz1: Need to change this to install g++-14 since you've now changed the CXX on line 43 and 51 to use g++-14

src/unittest/module_test.cpp
```diff
@@ -0,0 +1,127 @@
...
+	if (test_time != "2022-01-19T17:04:24Z") {
+		set_status(MODULE_IMPORT_BASIC, ts_failed, "timestamp conversion");
+		return;
+	}
+
+	// Message object creation test
+	dpp::message test_msg;
+	test_msg.content = "Test message from module";
+	test_msg.id = 111222333;
+	if (test_msg.content != "Test message from module" || test_msg.id != 111222333) {
```
Jaskowicz1: Just remove these for now since Codacy isn't happy.

src/unittest/module_test.cpp
```diff
@@ -0,0 +1,127 @@
...
+	// Role comparison test
+	dpp::role r1;
+	dpp::role r2;
+	r1.id = 100;
+	r1.position = 1;
+	r1.guild_id = 500;
+	r2.id = 200;
+	r2.position = 2;
+	r2.guild_id = 500;
+	if (!(r1 < r2) || (r1 > r2)) {
```
Jaskowicz1: Just remove these for now since Codacy isn't happy.

Jaskowicz1 (changes requested):

src/unittest/module_test.cpp
mikomikotaishi: How about I replace it with a lambda call instead?

src/unittest/module_test.cpp
mikomikotaishi: How about I replace it with a lambda call instead?

.github/workflows/test-docs-examples.yml
mikomikotaishi: Good catch, sorry I missed this.

mikomikotaishi:
Decided to replace the hard-coded checks with a lambda instead, which should hopefully be accepted by Codacy. If you'd prefer not to let me know, but I do think this is the best way to ensure we can still check the conditions in the unit tests.

Jaskowicz1:
Still seems annoyed about the role check, probably worth just removing it for now and we can address it in a new PR :)

mikomikotaishi:
OK, I just broke the checks into two lambdas, and if it still doesn't pass I'll just comment it out for a later PR.

Looks like it passes now.

.github/workflows/test-docs-examples.yml
Jaskowicz1, line 1: Change this to use ubuntu-24.04 (line 21) and that should fix the issue of it failing to install gcc14, then we should be all good!

Jaskowicz1 (changes requested):

.github/workflows/test-docs-examples.yml
mikomikotaishi: Done, and also rebased.

.github/workflows/test-docs-examples.yml
Jaskowicz1: I'm not sure that's true, cmake added module support in 3.28.

mikomikotaishi: CMake supports module building, yes, but it doesn't support building them when generating Makefiles (you'll get an error like "CMake does not support building modules with Unix Makefiles" or something like that), I saw this as recently as a few days ago on CMake 4.2 in fact.

Jaskowicz1: Ah yeah just tested on an ubuntu24.04 server, that's true. Can we make it still do `cd build && make -j2 && sudo make install` as `-G Ninja` will make it still do Ninja stuff (our CI does this and just confirmed it)

.github/workflows/test-docs-examples.yml
```diff
@@ -35,17 +35,17 @@ jobs:
...
       - name: Generate CMake
-        run: mkdir build && cd build && cmake -DDPP_NO_VCPKG=ON -DAVX_TYPE=T_fallback -DDPP_CORO=ON -DCMAKE_BUILD_TYPE=Debug ..
+        run: mkdir build && cd build && cmake -G Ninja -DDPP_NO_VCPKG=ON -DDPP_MODULES=ON -DAVX_TYPE=T_fallback -DDPP_CORO=ON -DCMAKE_BUILD_TYPE=Debug ..
         env:
-          CXX: g++-12
+          CXX: g++-14
   
       - name: Build Project
-        run: cd build && make -j2 && sudo make install
+        run: cd build && ninja -j2 && sudo ninja install
```
Jaskowicz1: Revert this back to `cd build && make -j2 && sudo make install` as `-G Ninja` sorts this.

.github/workflows/test-docs-examples.yml
```diff
@@ -35,17 +35,17 @@ jobs:
...
   
       - name: Test compile examples
-        run: cd docpages/example_code && mkdir build && cd build && cmake .. && make -j2
+        run: cd docpages/example_code && mkdir build && cd build && cmake -G Ninja .. && ninja -j2
```
Jaskowicz1: Change `ninja -j2` to `make -j2` since `-G Ninja` sorts this

Jaskowicz1 (changes requested):
Should be the last bits 😅

mikomikotaishi:
Hmm, this problem with the standard library has to do with symbols having internal linkage. They fixed most of this in GCC 15, because that's when they shipped module std.

Let me try it with Ubuntu 25.10 and GCC 15

.github/workflows/test-docs-examples.yml
```diff
@@ -18,7 +18,7 @@ permissions:
 jobs:
   test_docs_examples:
     name: Test build examples
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-25.10
```
Jaskowicz1: This is not a valid runner. Change it back to 24.04 please.

Jaskowicz1 (changes requested):

Jaskowicz1:
Since you're having issues with g++-14 (which makes no sense because I can build g++-14 modules fine locally), just use a clang version that's supported (pretty sure clang15 has modules, but use clang18).

Make sure you set the CXX for the env on both builds and then the apt install too 🫡

mikomikotaishi:
I have no idea why this current error is happening seeing as it never happened on my end, though I built mine in Clang 21.

Jaskowicz1:
Seems it's a clang18 issue 🙃 , change to clang20

CMakeLists.txt
```diff
@@ -33,6 +33,7 @@ option(DPP_USE_EXTERNAL_JSON "Use an external installation of nlohmann::json" OF
 option(DPP_USE_PCH "Use precompiled headers to speed up compilation" OFF)
 option(AVX_TYPE "Force AVX type for speeding up audio mixing" OFF)
 option(DPP_TEST_VCPKG "Force VCPKG build without VCPKG installed (for development use only!)" OFF)
+option(DPP_MODULES "Support for C++20 modules" OFF)
```
Jaskowicz1: Could we get this changed to `Support for C++20 modules (Experimental)` ? Needs a lot of testing and there's not a lot of gcc/clang versions that support modules properly

mikomikotaishi: OK

docpages/example_code/using_modules.cpp
```diff
@@ -0,0 +1,28 @@
+import dpp;
+
+using dpp::cluster;
+using std::ready_t;
+using dpp::slashcommand;
+using dpp::slashcommand_t;
+using dpp::start_type;
```
Mishura4: Why this? We don't do this for any other example. Also what is `std::ready_t`?

library/CMakeLists.txt
```diff
@@ -459,25 +459,70 @@ if (NOT BUILD_SHARED_LIBS)
...
 				target_compile_features(${testname} PRIVATE cxx_std_20)
 			else()
 				target_compile_features(${testname} PRIVATE cxx_std_17)
 			endif()
 			if (MSVC)
 				target_compile_options(${testname} PRIVATE /utf-8)
 			endif()
-			if(BUILD_SHARED_LIBS)
+			if (DPP_MODULES)
+				target_link_libraries(${testname} PUBLIC dpp_module)
```
Mishura4: Can't we set it up so that `target_link_libraries(${testname} PUBLIC dpp)` brings the module, as discussed on the discord?

Mishura4 (changes requested):

docpages/example_code/CMakeLists.txt
```diff
@@ -33,7 +33,7 @@ project(documentation_tests)
 
 string(ASCII 27 Esc)
 
-set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DDPP_CORO -std=c++20 -pthread -O0 -fPIC -rdynamic -DFMT_HEADER_ONLY -Wall -Wextra -Wpedantic -Werror -Wno-unused-parameter -Wno-deprecated-declarations")
+set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DDPP_CORO -DPP_MODULES -std=c++20 -pthread -O0 -fPIC -DFMT_HEADER_ONLY -Wall -Wextra -Wpedantic -Werror -Wno-unused-parameter -Wno-deprecated-declarations")
```
talisein: Shouldn't this be `-DDPP_MODULES`

Jaskowicz1: yes

mikomikotaishi: Fixed

library/CMakeLists.txt
mikomikotaishi: I think my current revision fixes this, but please review and let me know if it is acceptable.

docpages/example_code/using_modules.cpp
mikomikotaishi: It's habit, but fine, I don't mind changing it back. I just frankly find the namespace qualification to be useless and verbose.

.github/workflows/test-docs-examples.yml
```diff
@@ -35,17 +35,19 @@ jobs:
...
   
       - name: Build Project
-        run: cd build && make -j2 && sudo make install
+        run: cd build && cmake --build . -j2 && sudo ninja install
+        env:
+          CXX: clang++-20
   
       - name: Test compile examples
-        run: cd docpages/example_code && mkdir build && cd build && cmake .. && make -j2
+        run: cd docpages/example_code && mkdir build && cd build && cmake -G Ninja .. && cmake --build . -j2
```
Jaskowicz1: You need to add `-DDPP_MODULES=ON` after `cmake -G Ninja`, like the Generate CMake step.

Jaskowicz1 (changes requested):

.github/workflows/test-docs-examples.yml
mikomikotaishi: Done

Jaskowicz1:
Im rather confused why the CI keeps saying

```
Manually-specified variables were not used by the project:

    DPP_MODULES
```

cmake/CPackSetup.cmake
```diff
@@ -11,20 +11,40 @@ set(DPP_INSTALL_LIBRARY_DIR ${CMAKE_INSTALL_LIBDIR}/${DPP_VERSIONED})
...
-		RUNTIME DESTINATION  ${CMAKE_INSTALL_BINDIR}
-		INCLUDES DESTINATION ${CMAKE_INSTALL_INCLUDE_DIR})
+	if (DPP_MODULES)
+		install(TARGETS dpp
+			EXPORT ${DPP_EXPORT_NAME}
+			LIBRARY DESTINATION  ${CMAKE_INSTALL_LIBRARY_DIR}
+			ARCHIVE DESTINATION  ${CMAKE_INSTALL_LIBRARY_DIR}
+			RUNTIME DESTINATION  ${CMAKE_INSTALL_BINDIR}
+			INCLUDES DESTINATION ${CMAKE_INSTALL_INCLUDE_DIR}
+			FILE_SET CXX_MODULES DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/dpp/include)
```
Jaskowicz1: As mentioned on Discord, I'm pretty sure this is leading to the cppm being installed at `/usr/local/include/dpp/include/dpp/dpp.cppm`, I would change this to be `${CMAKE_INSTALL_INCLUDE_DIR}` and see if that works.

Jaskowicz1 (changes requested):

cmake/CPackSetup.cmake
mikomikotaishi: OK, let me try again.

mikomikotaishi:
Looks like we finally got things to pass. Could we move forward with merging?

mikomikotaishi:
@Jaskowicz1 could you re-review the changes again now that the tests pass

mikomikotaishi:
Hi, are there still any plans to merge this?

Mishura4:
bump

mikomikotaishi:
Apologies for the silence.

Some criticisms and requests were made earlier in the library development channel. Could those be re-posted here for PR clarity?

braindigitalis:
any chance we can trim it down so that its easier to review? perhaps by not introducing ninja as a dependency of docs build, not changing its compiler to clang, and keeping changes to a minimum in the cmakelists etc? right now there are 53 changed files.

docpages/example_code/CMakeLists.txt
```diff
@@ -28,19 +28,41 @@
 # libmpg123-dev
 # dpp latest master with -DDPP_CORO=ON installed sytemwide
 
-cmake_minimum_required (VERSION 3.16)
+cmake_minimum_required (VERSION 3.28)
 project(documentation_tests)
 
 include("${CMAKE_CURRENT_SOURCE_DIR}/../../cmake/colour.cmake")
 
-set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DDPP_CORO -std=c++20 -pthread -O0 -fPIC -rdynamic -DFMT_HEADER_ONLY -Wall -Wextra -Wpedantic -Werror -Wno-unused-parameter -Wno-deprecated-declarations")
```
braindigitalis, line 36: removal of -rdynamic breaks the eval.cpp example. it would compile, but wouldnt run, so doesnt serve as a good example.

docpages/example_code/CMakeLists.txt
```diff
@@ -28,19 +28,41 @@
 # libmpg123-dev
 # dpp latest master with -DDPP_CORO=ON installed sytemwide
 
-cmake_minimum_required (VERSION 3.16)
+cmake_minimum_required (VERSION 3.28)
```
braindigitalis, line 31: still not keen on this version bump. many of our users dont have this version of cmake.

docpages/example_code/autocomplete.cpp
```diff
@@ -23,7 +23,7 @@ int main()
 	});
 
 	/* The interaction create event is fired when someone issues your commands */
-	bot.on_slashcommand([&bot](const dpp::slashcommand_t & event) {
```
braindigitalis: do we need to change all the examples, it makes the PR much harder to review, so many files to go through.

library/CMakeLists.txt
```diff
@@ -330,20 +330,20 @@ if (HAVE_VOICE)
 	# Private statically linked dependencies
 	if(NOT BUILD_SHARED_LIBS AND NOT APPLE)
 		target_link_libraries(dpp PRIVATE
-				mlspp.a
-				mls_vectors.a
-				bytes.a
-				tls_syntax.a
-				hpke.a
+				$<BUILD_INTERFACE:mlspp.a>
```
braindigitalis, line 342: really really not liking that we changed this. it took a LOT of effort to make this always build everywhere. with static and dynamic linking, on a ton of different systems. some of which CI doesnt test (e.g. static builds)

src/unittest/module_test.cpp
```diff
@@ -0,0 +1,141 @@
+/************************************************************************************
```
braindigitalis, line 1: this doesnt actually test modules, just verifies they can be linked. but, i guess, theres nothing else we can do to test that, as it is a compile-time thing.

braindigitalis (changes requested):
left some thoughts :-) thanks for the continued effort on this!

mikomikotaishi:
OK; will look in to this.

braindigitalis:
all builds failed. nlohmann doesnt need touching in this pr any more.

docpages/example_code/CMakeLists.txt
mikomikotaishi: CMake only supports modules on 3.28 onwards, but this is just for building the examples with modules, it doesn't force this on the consumer.

docpages/example_code/autocomplete.cpp
mikomikotaishi: I've moved these changes into a separate PR, #1604

include/dpp/nlohmann/json.hpp
braindigitalis, line 1: this file should not be picked up as changed at all in this pr. please merge master into the branch

braindigitalis (approved):

include/dpp/dpp.cppm
```diff
@@ -0,0 +1,67 @@
...
+#include <dpp/export.h>
+#include <dpp/compat.h>
+
+export module dpp;
+
+export extern "C++" {
+    #include <dpp/dpp.h>
+    #include <dpp/dns.h>
+    #include <dpp/restrequest.h>
+    #include <dpp/unicode_emoji.h>
```
braindigitalis, line 78: I don't think unicode emoji should be in the module as default.
it is optional to include as it is massive and might just be something where we say "for this bit, you have to still #include it"

library/CMakeLists.txt
```diff
@@ -511,7 +537,6 @@ if(DPP_INSTALL)
 		# Installation
 
 		include(GNUInstallDirs)
-		install(TARGETS dpp LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR})
```
braindigitalis, line 520: why did this get removed, is it somewhere else now?

README.md
```diff
@@ -33,6 +33,7 @@ D++ is a lightweight and efficient library for **Discord** written in **modern C
 * Stable [Windows support](https://dpp.dev/buildwindows.html)
 * Ready-made compiled packages for Windows, Raspberry Pi (ARM64/ARM7/ARMv6), Debian x86/x64, and RPM based distributions
 * Highly scalable for large amounts of guilds and users
+* Support for [C++ modules](https://dpp.dev/using_modules.html)
```
braindigitalis: might want to read as experimental support

braindigitalis (changes requested):

include/dpp/dpp.cppm
mikomikotaishi: Modules only need to be compiled once, so even if we include this massive file, the price is paid only once. If we force users of modules to include the header, then they'd just be re-paying the price over and over anyway, just to make the price of compiling the module a little lower.

In fact, even compiling an extremely large file that only has declarations of compile-time constants like
```cpp
inline constexpr char32_t WHITE_HEAVY_CHECK_MARK = U'✅';
```
is still very cheap to compile.

If the concern is forcing people who don't want to use it at all to pay that price, we could hide it behind a macro; and, enabling `#include <dpp/unicode_emoji.h>` into the module could even be opt-in, but I definitely think users should be able to have the ability to access it from the module if they are sure they will use it.

README.md
mikomikotaishi: OK; I'll label it experimental clearly

mikomikotaishi: Labelled as experimental now

include/dpp/nlohmann/json.hpp
mikomikotaishi: Removed

library/CMakeLists.txt
mikomikotaishi: This install does still take place; it's done in `cmake/CPackSetup.cmake`, which has a richer install rule anyway; keeping it `library/CMakeLists.txt` just reinstalls the library a second time (with less information)

docpages/example_code/CMakeLists.txt
mikomikotaishi: I'm restoring it but it has to appear in `CMAKE_EXE_LINKER_FLAGS`; `-Wunused-command-line-argument` causes it to fail if it appears in `CMAKE_CXX_FLAGS`, because it is a link-only flag, and can't appear at the compile command.

library/CMakeLists.txt
mikomikotaishi: `$<BUILD_INTERFACE:>` is build-time transparent, during build it expands into `mlspp.a` and produces the same linker command.

.github/workflows/ci.yml
```diff
@@ -272,3 +272,89 @@ jobs:
         with:
           name: "libdpp - RPM Package ${{matrix.cfg.name}}"
           path: "${{github.workspace}}/build/*.rpm"
+
+  # Static library build (BUILD_SHARED_LIBS=OFF)
```
braindigitalis, line 276: this should be a separate pr, rather than adding it into here. we are finally starting to simplify the modules pr let's not sneak more features into it

braindigitalis (approved):

docpages/example_code/CMakeLists.txt
braindigitalis: this is part of the reason I didn't want to change to clang. the -Wall is a different set of warnings.

library/CMakeLists.txt
braindigitalis: then if it does the same why change it?

.github/workflows/ci.yml
mikomikotaishi: OK; broke off to #1612.

docpages/example_code/CMakeLists.txt
```diff
@@ -28,19 +28,43 @@
...
 
-set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DDPP_CORO -std=c++20 -pthread -O0 -fPIC -rdynamic -DFMT_HEADER_ONLY -Wall -Wextra -Wpedantic -Werror -Wno-unused-parameter -Wno-deprecated-declarations")
+option(DPP_MODULES "Build examples with C++20 modules support" ON)
+
+set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DDPP_CORO -std=c++20 -pthread -O0 -fPIC -DFMT_HEADER_ONLY -Wall -Wextra -Wpedantic -Werror -Wno-unused-parameter -Wno-deprecated-declarations")
 set (CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -O0")
 
+set (CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -rdynamic")
+
+# Create gcm.cache directory and symlink to the main build's dpp.gcm
```
Mishura4: Yes I can see that, but why?

docpages/example_code/CMakeLists.txt
```diff
@@ -28,19 +28,43 @@
...
+	
+	target_compile_definitions(${examplename}_out PRIVATE DPP_MODULES)
+	set_target_properties(${examplename}_out PROPERTIES CXX_SCAN_FOR_MODULES ON)
+	
+	if (CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
+		target_compile_options(${examplename}_out PRIVATE -fmodules-ts)
+	elseif(CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
+		target_compile_options(${examplename}_out PRIVATE 
+			-fmodules
+			-mavx2
```
Mishura4: Why AVX2 for an example?

library/CMakeLists.txt
Mishura4: @braindigitalis BUILD_INTERFACE only expands when building the library artifacts ; it's usually as a counterpart to INSTALL_INTERFACE which only expands when bringing the library from an install directory

Up until modules that was functionally the_ same as PRIVATE / PUBLIC but with modules actually PRIVATE *is* included when bringing module files (as they are translation units that need to be built privately into a public binary artifact, like .cpp files), even when linking an installed library. So BUILD_INTERFACE prevents this and is only used when building the library itself.

So adding BUILD_INTERFACE is correct here, even if I am not sure the author knew why.

Mishura4 (changes requested):
Example doesn't seem to build on Windows. Can we add it to CI somehow?
Also, please fix merge conflicts.

docpages/example_code/using_modules.cpp
```diff
@@ -0,0 +1,25 @@
+#include <string>
+#include <typeinfo>
```
Jaskowicz1: Is this include required?

docpages/example_programs/misc/using_modules.md
```diff
@@ -0,0 +1,9 @@
+\page using_modules Using D++ with C++ modules
+
```
Jaskowicz1, line 2: We should probably add a `\warning <message>` here to note that this is experimental.

We did this with coro a while back, would be good for this!

See https://dpp.dev/10.0.26/using-coroutines.html for ref

include/dpp/etf.h
```diff
@@ -42,7 +42,7 @@ namespace dpp {
 /**
  * @brief Current ETF format version in use
  */
-const uint8_t FORMAT_VERSION = 131;
+inline const uint8_t FORMAT_VERSION = 131;
```
Jaskowicz1: reason we can't change this to constexpr?

Jaskowicz1 (commented):
A couple comments, nothing to really reject for though

docpages/example_code/using_modules.cpp
mikomikotaishi: ~~Probably not. I had them here a long time ago but I can't see what they're used for now and I don't think they are used.~~

From a distance it wasn't clear, but it turns out it `<string>` really is needed for the `== "ping";` comparison.

docpages/example_code/CMakeLists.txt
mikomikotaishi: This wasn't needed, thanks for catching this.

docpages/example_programs/misc/using_modules.md
mikomikotaishi: OK, I'll add a `modules_warn.dox` file as well for it and copy the exact same format.

include/dpp/etf.h
mikomikotaishi: I'll make it `constexpr`

docpages/example_code/CMakeLists.txt
mikomikotaishi: Clang obtains the BMI paths through `-fprebuilt-module-path` which indicates where to scan, but GCC has no such option, and places its BMIs in `gcm.cache/`, which is why the lookup location is populated with the symlink to the actual interface.

Mishura4: Why is this necessary? I've not had to do this with any of my other projects

docpages/example_code/using_modules.cpp
Jaskowicz1: Yeah `<string>` is fine, was just the `<typeinfo>` I was curious about :)

docpages/example_programs/misc/using_modules.md
Jaskowicz1: Thanks!

docpages/include/modules_warn.dox
```diff
@@ -0,0 +1 @@
+\warning D++ modules are currently only supported by D++ on g++ 15, clang/LLVM 17, and MSVC 17.6 or above. Additionally, your program has to support C++20.
```
Jaskowicz1: Sorry to nitpick, could we word this to be `D++ modules are currently experimental and are only supported by D++ on g++ 15, clang/LLVM 17, and MSVC 17.6 or above. Additionally, your program has to support C++20.`

docpages/include/coro_warn.dox
```diff
@@ -1 +1 @@
-\warning D++ Coroutines are currently only supported by D++ on g++ 13, clang/LLVM 14, and MSVC 19.37 or above. Additionally, your program has to support C++20.
+\warning D++ coroutines are currently only supported by D++ on g++ 13, clang/LLVM 14, and MSVC 19.37 or above. Additionally, your program has to support C++20.
```
Jaskowicz1: I would remove this change, doesn't really match the capitalisation across the board but also better to keep changes to a minimum

.github/workflows/test-docs-examples.yml
```diff
@@ -35,17 +35,49 @@ jobs:
...
+      - name: Add MSBuild to PATH
+        uses: microsoft/setup-msbuild@30375c66a4eea26614e0d39710365f22f8b0af57 # v3.0.0
+
+      - name: Generate CMake
+        run: cmake -B build -G "Visual Studio 17 2022" -DDPP_NO_VCPKG=ON -DDPP_MODULES=ON -DDPP_BUILD_TEST=OFF -DAVX_TYPE=AVX0
+        env:
+          DONT_RUN_VCPKG: true
+
+      - name: Build module example
+        run: cmake --build build --target using_modules_example --config Release --parallel 2
```
Jaskowicz1, line 81: does this target exist? I can't see it anywhere in this PR. If it does exist, it's not being built for Windows.

This is causing CI to fail

Jaskowicz1 (changes requested):

.github/workflows/test-docs-examples.yml
Mishura4: I asked for examples to be built in CI but yes the CI needs to be fixed

Jaskowicz1: Yeah i'm happy with that, it's just the fact that the `using_modules_example` target doesn't look like it exists

docpages/example_code/CMakeLists.txt
```diff
@@ -28,19 +28,29 @@
...
 project(documentation_tests)
 
 include("${CMAKE_CURRENT_SOURCE_DIR}/../../cmake/colour.cmake")
 
-set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DDPP_CORO -std=c++20 -pthread -O0 -fPIC -rdynamic -DFMT_HEADER_ONLY -Wall -Wextra -Wpedantic -Werror -Wno-unused-parameter -Wno-deprecated-declarations")
+option(DPP_MODULES "Build examples with C++20 modules support" ON)
+
+find_package(dpp REQUIRED)
+
+set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DDPP_CORO -std=c++20 -pthread -O0 -fPIC -DFMT_HEADER_ONLY -Wall -Wextra -Wpedantic -Werror -Wno-unused-parameter -Wno-deprecated-declarations -Wno-unused-command-line-argument")
```
Jaskowicz1, line 40: This is gaining `-std=gnu++20` during build, do we want to have `CMAKE_CXX_EXTENSIONS` set to false?

It seems to be causing a build fail due to `error: GNU extensions was disabled in AST file 'CMakeFiles/dpp__dpp@synth_c8bec728562e.dir/e782fe423ed1.bmi' but is currently enabled`

Jaskowicz1 (changes requested):

docpages/example_code/CMakeLists.txt
mikomikotaishi: I was considering this, but I wasn't sure if disabling this would be too intrusive an option for consumers. I do think it's the most portable option though, so it's a change I'd support making.

Jaskowicz1: Yeah I'm just a bit confused on where the error has come from 😅 It seems to have started from your recent commit

mikomikotaishi: I planned to have it enabled for tests and leave it untouched on the main build itself so that it could remain agnostic for users, but seeing as it's an experimental feature anyway I think we'd probably just best keep it disabled for simplicity for the time being.

Jaskowicz1:
ngl i'm pretty happy with moving to clang-cl for windows lol

docpages/example_code/CMakeLists.txt
```diff
@@ -28,19 +28,31 @@
...
 
+set (CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -rdynamic")
+
 file(GLOB example_list ./*.cpp)
 foreach (example ${example_list})
 	get_filename_component(examplename ${example} NAME)
 	message(STATUS "Found example '${BoldBlue}${examplename}${ColourReset}'")
 	add_executable(${examplename}_out ${example})
-	target_link_libraries(${examplename}_out dl dpp mpg123 oggz ogg opusfile opus)
+	target_link_libraries(${examplename}_out dl dpp::dpp mpg123 oggz ogg opusfile opus)
```
Jaskowicz1, line 52: More of a question really, do we want to point out that you need to link `dpp` like this (`dpp::dpp`) in cmake via the `using_modules.md` page?

mikomikotaishi: Good catch, I agree. I'll add it to the documentation along with some of the settings we chose to clarify usage to consumers

Jaskowicz1 (approved):

PR #1533, opened 2025-12-30
ci: Added offline Unit Tests to Windows, CI now depends on a pre-build.

Jaskowicz1:
This PR now lets Windows build Unit Tests. CI now also has a pre-build which will just generate CMake. In the event that there is any issue in that step, CI will prevent any other runners, meaning we don't throw too many resources.

This PR also reduces the amount of clang runners for linux. We now only do the lowest and highest version for an ubuntu LTS version.

No Checklist applies for this PR.


Ashthetik:
As per the discord convo, I'll be looking at picking up on this shortly

Jaskowicz1:
Going to close this PR since you're taking it over @Ashthetik

PR #1543, opened 2026-01-04
Fix hang when sending HTTP request with empty token and non-empty logger

osipyonok:
Fix infinite loop when sending HTTP request with empty token and non-empty logger
A simple bot to reproduce the issue: [hang_test.txt](https://github.com/user-attachments/files/24424056/hang_test.txt)

## Code change checklist

### DIFF
--- a/src/dpp/cluster.cpp
+++ b/src/dpp/cluster.cpp
@@ -195,10 +195,12 @@ void cluster::log(dpp::loglevel severity, const std::string &msg) const {
 		dpp::log_t logmsg(nullptr, 0, msg);
 		logmsg.severity = severity;
 		logmsg.message = msg;
-		size_t pos{0};
-		while ((pos = logmsg.message.find(token, pos)) != std::string::npos) {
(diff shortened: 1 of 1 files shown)


braindigitalis:
wait, why would you do this?

you cant do a .request before you start the cluster, its not to do with the lack of a token, the reactor hasnt started.

braindigitalis (approved):

braindigitalis:
thanks for the PR regardless :+1: 
the issue is a real one, the token should be checked for non-empty before we attempt to log. however, as i said this wont actually ever make a request, because it relies on the cluster being started first, in the real world you'd do that request in on_ready.

osipyonok:
> wait, why would you do this?
> 
> you cant do a .request before you start the cluster, its not to do with the lack of a token, the reactor hasnt started.

Ah, my bad! Thanks for the clarification. I was aiming for a super minimal demo bot, but I ended up oversimplifying it :sweat_smile:

braindigitalis:
makes sense :) Happy to merge this in when the CI finishes its job

PR #1552, opened 2026-02-20
[feat] Added get_path to http_server_request so that the user can get the request path

Henonicks:
Apparently `dpp::http_server_request::path` is private and there is no way to even get its value so this fixes it

## Code change checklist


include/dpp/http_server_request.h
```diff
@@ -129,6 +129,12 @@ class DPP_EXPORT http_server_request : public ssl_connection {
 	 */
 	bool timed_out;
 
+	/**
+	 * Get request path
+	 * @return request path
+	 */
+	std::string get_path() const;
```
braindigitalis, line 136: we can return a const reference here:

const std::string& 

instead of std::string

braindigitalis (changes requested):

include/dpp/http_server_request.h
Henonicks: The reason I'm not returning one is to comply with the signatures of the rest of the strings - all of them return copies.

braindigitalis (approved):

PR #1575, opened 2026-04-13
docs(example): Add minimal slash command (ping) example.

kaedeek:
## Summary
Add a minimal slash command (ping) example.

## Additional context
While similar functionality is shown in the README, this example provides a minimal, runnable version for beginners.

### DIFF
--- a/example/simple_example.cpp
+++ b/example/simple_example.cpp
@@ -1,0 +1,23 @@
+#include <dpp/dpp.h>
+#include <cstdlib>
+
+int main() {
+	dpp::cluster bot(std::getenv("BOT_TOKEN"));
+
+	bot.on_slashcommand([](auto event) {
+		if (event.command.get_command_name() == "ping") {
+			event.reply("Pong!");
+		}
+	});
+
+	bot.on_ready([&bot](auto event) {
+		if (dpp::run_once<struct register_bot_commands>()) {
+			bot.global_command_create(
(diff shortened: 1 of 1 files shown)


ruslan-ilesik:
Hey, thanks for PR, while I am not the main reviewer I would like to give a suggestion.

Would be nice if you could also put usage of

https://dpp.dev/classdpp_1_1cluster.html#acb7b73536f793da2a2955d70baa991f1

Into the guide, as better alternative to if else in on_slashcommand.

Ashthetik:
I'm not much of a contributor, but what would be the purpose from moving the ping example from https://dpp.dev/firstbot.html to the example programs?

ruslan-ilesik:
Also, would be great if you had comments in code itself quickly explaining what each of lines do, for new people to c++ it may be very helpful

ruslan-ilesik:
> I'm not a contributor, but what would be the purpose from moving the ping example from https://dpp.dev/firstbot.html to the example programs? 

Also a good point, but we have it as first bot AND kinda advertisement like "look, it is not harder than js", so I am not sure about that. @braindigitalis should be judge of it.

But in any case, either keep it in examples or update main page to more "fresh" implementation may be beneficial

Jaskowicz1:
This pr doesn't make any sense, it's adding this to a whole new folder? We already have docs about a minimal slash command example (https://dpp.dev/firstbot.html)

PR #1578, opened 2026-04-29
fix: 100% busy main thread while bot is not playing audio in vc

usyless:
I am unsure whether the fix is valid, considering it goes against the comment to always set WANT_WRITE.
I have tested with epoll in wsl, poll under windows and whatever my raspberry pi 3b+ on debian 12 uses and it seems to work
However my testing has not been extensive, and therefore I would like for some more confirmation as to whether this is valid.

I'm also unsure on the needs stop frames part, then again, the voice stuff is too much for me to all look at
And im unsure on whether the send_now branch should also need the same treatment with the socket update
I also want to look into whether update socket can be made more efficient by preventing the copy of the entire object each time

## Code change checklist

### DIFF
--- a/src/dpp/voice/enabled/read_write.cpp
+++ b/src/dpp/voice/enabled/read_write.cpp
@@ -34,8 +34,17 @@ void discord_voice_client::send(const char* packet, size_t len, uint64_t duratio
 		frame.packet.assign(packet, packet + len);
 		frame.duration = duration;
 
-		std::lock_guard<std::mutex> lock(this->stream_mutex);
-		outbuf.emplace_back(frame);
+		bool was_empty = false;
+		{
+			std::lock_guard<std::mutex> lock(this->stream_mutex);
+			was_empty = outbuf.empty();
+			outbuf.emplace_back(frame);
+		}
+
+		if (was_empty) {
+			udp_events.flags = WANT_READ | WANT_WRITE | WANT_ERROR;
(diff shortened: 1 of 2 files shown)


braindigitalis:
can someone on bsd or osx test this under kqueue please?

braindigitalis:
@usyless is this pr ok to move out of draft

usyless:
I haven't really tested enough, so it'd be nice for others to test too to ensure it doesn't break random parts of voice, but sure that's fine as I've been using it with my two bots and not noticed any issue there

Jaskowicz1:
If you can whip up a quick test program, I can get it running on OSX

usyless:
Oh awesome, I can do that tonight or tomorrow morning.

usyless:
@Jaskowicz1 Here is a basic example [main.cpp](https://github.com/user-attachments/files/27440096/main.cpp), its a bit cooked since i re-used the code from my other bot with a single server queue-based system but it has joining, leaving and playing (can also leave while playing reliably with no crash - im just using the pausing as an indicator of that here)

braindigitalis:
can't we just use src/davetest?

usyless:
Oh idk, I just made a quick program as asked but yea built in tests would make sense

Jaskowicz1:
ngl i forgot about davetest

braindigitalis (approved):

Issue #1476, opened 2025-08-14
Possible issue with thread management (Helgrind)


fclivaz42:
**Git commit reference**
d46bd98396609c83defff2da3227be383935d5ea (V10.1.3)

**Describe the bug**
Running the basic ping/pong bot example through helgrind (valgrind --tool=helgrind) creates a gigantic log of data races and mutex/rwlock issues. 

**To Reproduce**
1. Compile the base program shown at https://dpp.dev/firstbot.html, without `bot.on_log()`. (I personally used the flags `-ldpp -std=c++23 -g3`)
2. Run `valgrind --tool=helgrind ./a.out`
3. Observe!

**Expected behavior**
Helgrind should not be outputting so many "possible data race", "write lock granted on mutex/rwlock which is currently wr-held by a different thread", and so on.

**Screenshots**
Not applicable, however I can attach both a log file as well as the sample code (without the token heh) that produced this error.

**System Details:**
 - OS: Arch Linux
 - Discord Client used for testing: N/A

**Additional context**
This seems to happen regardless of the OS. I also attempted it on Alpine Linux and i could recreate it as well.
Oddly enough, -fsanitize=thread does not complain at all.

Here's the program that I used to recreate this issue:

```c++
#include <dpp/dpp.h>

const std::string BOT_TOKEN = "insert token here";

int main() {
	dpp::cluster bot(BOT_TOKEN);

	bot.on_slashcommand([&bot](const dpp::slashcommand_t& event) {
		if (event.command.get_command_name() == "stop") {
			event.reply("Pong!");
			bot.shutdown(); // shutting the bot down also seems to throw a million issues!
		}
	});

	bot.on_ready([&bot](const dpp::ready_t& event) {
		(void)event;
		if (dpp::run_once<struct register_bot_commands>()) {
			bot.global_command_create(dpp::slashcommand("stop", "stop!", bot.me.id));
		}
	});

	bot.start(dpp::st_wait);
}
```

And the attached Valgrind log:

[helgrind.log](https://github.com/user-attachments/files/21778888/helgrind.log)

Sorry if this seems a bit hectic? I could not find another mention of helgrind in the issues, so I decided to open one myself.

Have a nice day!

braindigitalis:
is this related to an actual stability issue?

it runs fine for me in production with 217,000 servers, and the issues don't seem to be in our remit?

```
 
[...]

PR #1585, opened 2026-05-12
fix: addition of a few protections to prevent data races

fclivaz42:
This pull request aims to fix a few recurring data races that keep happening, which I mentionned in #1476 . I have run the unit tests and they pass with 100% completion.

However, I have found a few issues with that:
- `guild_get_ban()` fails if there has not been anyone who has been banned. This does not seem to be affected by the changes in this PR directly since it fails on a clean build too.
- `channel_pins_get()` also fails if there are no pinned messages at all. Same as above.

There still is a data race happening whith a write in `dpp::request_concurrency_queue::~request_concurrency_queue()` while `dpp::thread_pool::thread_pool` is active (offending line seems to be the `task.function()` in the `try` block) which I would still like to fix, but have no clue how just yet.

I was only able to test `epoll()` and there are no more races when `fds.erase(fd)` gets called. the kqueue call does not have any thread protection, but I assume the fix is the same as `epoll()`. Since I cannot test it myself, i have left it out of this PR.

## Code change checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,3 +32,4 @@ config.json
 compile_commands.json
 src/dpp/dpp.rc
 .DS_STORE
+.env
--- a/src/dpp/discordevents.cpp
+++ b/src/dpp/discordevents.cpp
@@ -272,6 +272,8 @@ time_t ts_not_null(const json* j, const char* keyname)
 	 * Note that discord timestamps contain a decimal seconds part, which time_t and struct tm
 	 * can't handle. We strip these out.
 	 */
+	static	std::mutex	ts_time;
+	std::lock_guard<std::mutex>	lg_time(ts_time);
 	time_t retval = 0;
 	if (j->contains(keyname) && !(*j)[keyname].is_null() && (*j)[keyname].is_string()) {
 		tm timestamp = {};
--- a/src/dpp/queues.cpp
+++ b/src/dpp/queues.cpp
@@ -153,6 +153,8 @@ void populate_result(const std::string &url, cluster* owner, http_request_comple
 /* Returns true if the request has been made */
 bool http_request::is_completed()
 {
+	// For some odd reason, BSD requires this to have a lock to prevent data races even though `completed` is std::atomic.
+	std::unique_lock<std::mutex> lock(this_captured_mutex);
 	return completed;
 }
 
@@ -281,8 +283,8 @@ http_request_completion_t http_request::run(request_concurrency_queue* processor
 					{
 						std::lock_guard<std::mutex> lock(this_captured_mutex);
 						this_captured = false;
+						this_captured_signal.notify_all();
 					}
-					this_captured_signal.notify_all();
 				});
 			}
 		);
@@ -317,7 +319,7 @@ request_concurrency_queue::request_concurrency_queue(class cluster* owner, class
 		tick_and_deliver_requests(in_index);
 		/* Clear pending removals in the removals queue */
 		if (time(nullptr) % 90 == 0) {
-			std::scoped_lock lock1{in_mutex};
+			std::scoped_lock lock1{rem_mutex};
 			for (auto it = removals.cbegin(); it != removals.cend();) {
 				if ((*it)->is_completed()) {
 					it = removals.erase(it);
@@ -409,6 +411,7 @@ void request_concurrency_queue::tick_and_deliver_requests(uint32_t index)
 			{
 				/* Find the owned pointer in requests_in */
 				std::scoped_lock lock1{in_mutex};
+				std::scoped_lock lock2{rem_mutex};
 
 				const std::string &key = request_view->endpoint;
 				auto [begin, end] = std::equal_range(requests_in.begin(), requests_in.end(), key, compare_request{});
--- a/src/dpp/queues.cpp
+++ b/src/dpp/queues.cpp
@@ -262,7 +262,9 @@ http_request_completion_t http_request::run(request_concurrency_queue* processor
 					/* We are globally rate limited - user up to shenanigans */
 					processor->requests->globally_limited_until = (newbucket.retry_after ? newbucket.retry_after : newbucket.reset_after) + newbucket.timestamp;
 				}
+				processor->buck_mutex.lock();
 				processor->buckets[this->endpoint] = newbucket;
+				processor->buck_mutex.unlock();
 
 				/* Transfer it to completed requests */
 				{
@@ -333,6 +335,8 @@ request_concurrency_queue::request_concurrency_queue(class cluster* owner, class
 
 request_concurrency_queue::~request_concurrency_queue()
 {
+	std::scoped_lock lock1{in_mutex};
+	std::lock_guard lock2{rem_mutex};
 	terminate();
 	creator->stop_timer(in_timer);
 }
@@ -379,31 +383,38 @@ void request_concurrency_queue::tick_and_deliver_requests(uint32_t index)
 		for (auto& request_view : requests_view) {
 			const std::string &key = request_view->endpoint;
 			http_request_completion_t rv;
-			auto currbucket = buckets.find(key);
 
-			if (currbucket != buckets.end()) {
-				/* There's a bucket for this request. Check its status. If the bucket says to wait,
-				 * skip all requests until the timer value indicates the rate limit won't be hit
-				 */
-				if (currbucket->second.remaining < 1) {
-					uint64_t wait = (currbucket->second.retry_after ? currbucket->second.retry_after : currbucket->second.reset_after);
-					if ((uint64_t)time(nullptr) > currbucket->second.timestamp + wait) {
-						/* Time has passed, we can process this bucket again. send its request. */
-						request_view->run(this, creator);
-					} else {
-						if (!request_view->waiting) {
-							request_view->waiting = true;
+			{
+				std::unique_lock<std::mutex>	unilock{buck_mutex};
+				auto currbucket = buckets.find(key);
+
+				if (currbucket != buckets.end()) {
+					/* There's a bucket for this request. Check its status. If the bucket says to wait,
+					 * skip all requests until the timer value indicates the rate limit won't be hit
+					 */
+					if (currbucket->second.remaining < 1) {
+						uint64_t wait = (currbucket->second.retry_after ? currbucket->second.retry_after : currbucket->second.reset_after);
+						if ((uint64_t)time(nullptr) > currbucket->second.timestamp + wait) {
+							/* Time has passed, we can process this bucket again. send its request. */
+							unilock.unlock();
+							request_view->run(this, creator);
+						} else {
+							if (!request_view->waiting) {
+								request_view->waiting = true;
+							}
+							/* Time not up yet, wait more */
+							break;
 						}
-						/* Time not up yet, wait more */
-						break;
+					} else {
+						/* We aren't at the limit, so we can just run the request */
+						unilock.unlock();
+						request_view->run(this, creator);
 					}
 				} else {
-					/* We aren't at the limit, so we can just run the request */
+					/* No bucket for this endpoint yet. Just send it, and make one from its reply */
+					unilock.unlock();
 					request_view->run(this, creator);
 				}
-			} else {
-				/* No bucket for this endpoint yet. Just send it, and make one from its reply */
-				request_view->run(this, creator);
 			}
 
 			/* Remove from inbound requests */
@@ -411,7 +422,7 @@ void request_concurrency_queue::tick_and_deliver_requests(uint32_t index)
 			{
 				/* Find the owned pointer in requests_in */
 				std::scoped_lock lock1{in_mutex};
-				std::scoped_lock lock2{rem_mutex};
+				std::lock_guard lock2{rem_mutex};
 
 				const std::string &key = request_view->endpoint;
 				auto [begin, end] = std::equal_range(requests_in.begin(), requests_in.end(), key, compare_request{});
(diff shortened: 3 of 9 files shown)


braindigitalis:
when unit testing, specific setup is expected. specifically some messages should be pinned and a user should be banned. on my unit test server I ban Dyno, because the id is well known.
if the tests generally pass this is good, and ok to be merged to dev.

braindigitalis:
thank you lots for the contribution!

fclivaz42:
No problem! I am spinning up a FreeBSD machine/VM to test if kqueue could also benefit from a mutex, and if it does i'll drop an extra commit if that's fine by you.

braindigitalis:
yeah that's fine if you want to try that this can wait till you say it's ready.

fclivaz42:
Testing on BSD proved that the mutex in kqueue was also needed.

I also *seem* to have fixed the other data races I was talking about. Unit tests passed, but i have only lightly tested in real-world to check if there were any more data races. If I find anything else, I'll be making another PR!

Edit: I cannot guarantee anything on windows since I have no idea how to build any of my testing or how to look for data races on windows.

fclivaz42:
> thank you lots for the contribution!

And thank you a whole lot for this library! thorougly loving using it, least i could do is help the project in some way ;)

braindigitalis (approved):

fclivaz42:
update: I have found more race conditions in the SSL connection handler, looking into it now

fclivaz42:
Hello again! Fixed a few more things, most notably a few races in `ts_not_null`, `epoll` and the ratelimit queue.

src/dpp/discordevents.cpp
```diff
@@ -272,6 +272,8 @@ time_t ts_not_null(const json* j, const char* keyname)
 	 * Note that discord timestamps contain a decimal seconds part, which time_t and struct tm
 	 * can't handle. We strip these out.
 	 */
+	static	std::mutex	ts_time;
```
braindigitalis, line 275: why this?

this doesn't act on shared state.

.gitignore
```diff
@@ -32,3 +32,4 @@ config.json
 compile_commands.json
 src/dpp/dpp.rc
 .DS_STORE
+.env
```
braindigitalis, line 35: why is this here, dpp never has a .env

src/dpp/thread_pool.cpp
```diff
@@ -66,20 +66,18 @@ thread_pool::~thread_pool() {
...
 	}
 }
 
 void thread_pool::enqueue(thread_pool_task task) {
-	{
-		std::unique_lock<std::mutex> lock(queue_mutex);
-		tasks.emplace(std::move(task));
-	}
+	std::unique_lock<std::mutex> lock(queue_mutex);
+	tasks.emplace(std::move(task));
```
braindigitalis, line 79: not feeling this.
you're now holding the lock while the notify fires so it remains locked while the event runs, it is heavily locking, will be slower and risks deadlocks in production.

braindigitalis (changes requested):

src/dpp/discordevents.cpp
fclivaz42: This was racing on BSD, and I should have commented it. Putting this mutex solved the racing

braindigitalis:
where are you finding these race conditions, as they aren't a problem in production even on huge bots. I am a bit concerned that this is going to end up causing deadlocks as we have no way to test this apart from  throwing it into a prod bot. unit tests don't test soak, and actual production heavy usage over months.

.gitignore
fclivaz42: Slipped, but in retrospect can be useful to source a .env for eg. unit testing where you can put the required environment variables

src/dpp/thread_pool.cpp
fclivaz42: As weird as locking a notify sounds, I've had a better track record with firing a notify while still holding the lock since the waiting thread(s) will contend the lock and then check the predicate, assuring that there is no race condition on the predicate itself. Once the lock is released, the check on `tasks` can be made (after the emplace) so there is no race condition on `tasks`.

fclivaz42:
I'm mostly finding them while idling, launching multiple slash commands in parallel and with reconnections. The way I've been testing this is by having a few slashcommands be launched in different channels that take a while to complete and that's mostly how I found these races, having even encountered segmentation faults. Those locks made the commands be able to complete without an issue. 
I have to admit that I wasn't *personally* able to test other cases (voice chat, guild administration, mutliple shards for multiple guilds, etc), but since I will be running my own bots on my own fork/branch, what we can do is not merge what sounds odd to you and I can report later if I've had any issues on my side?

src/dpp/thread_pool.cpp
Mishura4: `you're now holding the lock while the notify fires so it remains locked while the event runs`
What do you mean? @braindigitalis 
The lock should be immediately released after the cv is notified

src/dpp/queues.cpp
```diff
@@ -260,7 +262,9 @@ http_request_completion_t http_request::run(request_concurrency_queue* processor
 					/* We are globally rate limited - user up to shenanigans */
 					processor->requests->globally_limited_until = (newbucket.retry_after ? newbucket.retry_after : newbucket.reset_after) + newbucket.timestamp;
 				}
+				processor->buck_mutex.lock();
 				processor->buckets[this->endpoint] = newbucket;
+				processor->buck_mutex.unlock();
```
Mishura4, line 267: Can this be an RAII lock?

fclivaz42: Could be but i found it to be overkill for just one line

ruslan-ilesik: for reference, a lot of codebase uses stuff like

code
{
   std::unique_lock lk(mutext);
   ...code...
} //here is will be automatically unlocked.
code

and we would like to keep it consistenst

fclivaz42: Agreed, that can be changed with no issue

Mishura4: It's just good practice, it's safety against exceptions. `processor->buckets[this->endpoint] = newbucket;` will only throw from a std::bad_alloc and that's probably unrecoverable anyways, but we should still release the lock just in case

src/dpp/queues.cpp
```diff
@@ -153,6 +153,8 @@ void populate_result(const std::string &url, cluster* owner, http_request_comple
 /* Returns true if the request has been made */
 bool http_request::is_completed()
 {
+	// For some odd reason, BSD requires this to have a lock to prevent data races even though `completed` is std::atomic.
```
Mishura4, line 156: What do you mean "requires"?

fclivaz42: Was racing on the variable if the lock was absent

src/dpp/discordevents.cpp
braindigitalis: but how do you know?

braindigitalis (changes requested):

src/dpp/queues.cpp
braindigitalis: how was it racing? did you have an actual issue in production or are you using tsan or something, trying to figure out how you're finding them as this pr is getting big and will end up hard to test

braindigitalis (changes requested):

src/dpp/discordevents.cpp
fclivaz42: Because TSAN was pointing at `crossplatform_strptime` performing a write while `ts_not_null` was performing a read

braindigitalis:
doesn't build in arm 6 32 bit, the compilation fails due to missing 64 bit wide atomic integer.

src/dpp/queues.cpp
fclivaz42: No issue in production (at least personally) but while using debugging tools for my own stuff, this popped up. Since it still kept racing on BSD, I dropped the extra lock there.

braindigitalis: in that case the right fix here at least is to use the _r version that is thread safe, not just stick a mutex around it, libc provides a thread safe form

fclivaz42: by _r you mean the load() function? because if that's the case yeah that is my bad for not thinking of it

braindigitalis: no I meant timegm_r. but... only freebsd is not thread safe here.
as such this is adding contention on the most popular platforms (Linux, windows) to benefit one with a bug. the lock only applies to bsd.
see: 
https://man7.org/linux/man-pages/man3/timegm.3.html
attributes section, MT-safe.

fclivaz42: I'm confused, what does this have to do with returning a boolean?

fclivaz42:
Regarding the arm6 32-bit, i am not entirely sure it is my direct fault but i will have to admit that it's odd that it fails here since I haven't touched any variables that the error stack is mentionning. The only thing that comes to my mind would be the change in `sslconnection`, but that is in reagard to `last_unique_id` which doesnt't have anything to do withe time tracking.

edit: my bad i cannot read, is it about the linking issues? if thats the case wouldnt a define switch do the trick to use a uint32 instead of uint64 on 32-bit?

src/dpp/queues.cpp
braindigitalis: timegm is the actual issue, not a boolean, just realised I was taking about the other section.

but you shouldn't need a mutex here at all. something is wrong with how you're approaching that. can't just stick a mutex on it.

src/dpp/thread_pool.cpp
braindigitalis: but why hold it while we notify at all?

braindigitalis (changes requested):

braindigitalis:
> Regarding the arm6 32-bit, i am not entirely sure it is my direct fault but i will have to admit that it's odd that it fails here since I haven't touched any variables that the error stack is mentionning. The only thing that comes to my mind would be the change in `sslconnection`, but that is in reagard to `last_unique_id` which doesnt't have anything to do withe time tracking.
> 
> edit: my bad i cannot read, is it about the linking issues? if thats the case wouldnt a define switch do the trick to use a uint32 instead of uint64 on 32-bit?

no the fix isn't to shrink the type, need to figure out the right way to do what you're trying to do on that platform

braindigitalis:
my main concern here is too many changes to properly test all at once. you keep adding a bit more and a bit more. was going to merge the original.

it started off as a couple of locks we could easily test in a production bot to now dozens of changes, that isn't so easy to test.

I suggest we take this back to  the simpler change and do it a bit at a time. adding tons of locks all at once with no way to test it properly (unit tests can't) is a recipe for deadlocks.

ruslan-ilesik:
> my main concern here is too many changes to properly test all at once. you keep adding a bit more and a bit more. was going to merge the original.
> 
> it started off as a couple of locks we could easily test in a production bot to now dozens of changes, that isn't so easy to test.
> 
> I suggest we take this back to  the simpler change and do it a bit at a time. adding tons of locks all at once with no way to test it properly (unit tests can't) is a recipe for deadlocks.

Once it will be ready for testing tell me, i can run my bot under it. I don't use vouce stuff so if there is changes there I would not catch it. Also i am on Debian now not freebsd

braindigitalis:
> > my main concern here is too many changes to properly test all at once. you keep adding a bit more and a bit more. was going to merge the original.
> > it started off as a couple of locks we could easily test in a production bot to now dozens of changes, that isn't so easy to test.
> > I suggest we take this back to  the simpler change and do it a bit at a time. adding tons of locks all at once with no way to test it properly (unit tests can't) is a recipe for deadlocks.
> 
> Once it will be ready for testing tell me, i can run my bot under it. I don't use vouce stuff so if there is changes there I would not catch it. Also i am on Debian now not freebsd

we still need to simplify 

there are far too many locking semantic changes all at once. too much risk of deadlock.

testing on your bot would be a start but this needs splitting into multiple prs. the first of which is what was initially in this pr at time of submission 

the freebsd only platform strftime tying for example would throttle performance everywhere else including Linux and windows that dont need a lock there

ruslan-ilesik:
> > > my main concern here is too many changes to properly test all at once. you keep adding a bit more and a bit more. was going to merge the original.
> > > it started off as a couple of locks we could easily test in a production bot to now dozens of changes, that isn't so easy to test.
> > > I suggest we take this back to  the simpler change and do it a bit at a time. adding tons of locks all at once with no way to test it properly (unit tests can't) is a recipe for deadlocks.
> > 
> > Once it will be ready for testing tell me, i can run my bot under it. I don't use vouce stuff so if there is changes there I would not catch it. Also i am on Debian now not freebsd
> 
> we still need to simplify 
> 
> there are far too many locking semantic changes all at once. too much risk of deadlock.
> 
> testing on your bot would be a start but this needs splitting into multiple prs. the first of which is what was initially in this pr at time of submission 
> 
> the freebsd only platform strftime tying for example would throttle performance everywhere else including Linux and windows that dont need a lock there 

Sure, I am only implying that I am willing to volunteer as "test in production" :)

fclivaz42:
I'm perfectly fine with multiple PR's! no worries with that, do you want me to delete this one and make a new one with only the first batch of changes (with the fixes to epoll from the third batch)?

src/dpp/queues.cpp
fclivaz42: After a bit more testing I think that yeah it was an error on my end while testing the first time, probably was running tests without the actual fix and went overboard with a "solution". removing this and re-testing today yielded expected results so yeah i'll be dropping this. Thanks for the comment!

fclivaz42:
Closing since this is not getting merged, will re-open one with less changes.

PR #1586, opened 2026-05-15
fix: add protections against dataraces in epoll engine, cluster timer and queues

fclivaz42:
As discussed, here is a smaller PR with less risky changes.
I have changed the following:
- `epoll`, where I removed an unused variable and added a lock where elements were removed from the fds vector
- `kqueue`, where I have added a lock similar to `epoll`
- `queues`, where I have changed two booleans to `std::atomic` booleans, added a safeguard for the `cli` variable and guards for the `removals` vector.
- `timer`, where I added a guard to an unguarded if statement.

I believe these changes are light enough to not cause any issues whatsoever, and have kept the riskier changes out of this PR for now while I do more testing on my end.

## Code change checklist

### DIFF
--- a/include/dpp/queues.h
+++ b/include/dpp/queues.h
@@ -229,12 +229,17 @@ class DPP_EXPORT http_request {
 	/**
 	 * @brief True if request has been made.
 	 */
-	bool completed;
+	std::atomic<bool> completed;
 
 	/**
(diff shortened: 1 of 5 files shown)


fclivaz42:
this is definitely me when i dont pay attention to where i am merging

PR #1615, opened 2026-07-19
ci: sitemap now uses a pinned version

Jaskowicz1:
This PR changes the sitemap.yml to be a pinned version for security reasons (see the OpenSSF Scorecard)

Templates are not applicable for this PR.


braindigitalis (commented):
why remove -g, does it still run? note this action takes literally hours

Jaskowicz1:
I can’t test this action due to the requirements of the action, but I
couldn’t see a reason as to why it had to be -g.

Thought, if we are worried, the scorecard may be okay with us only adding
the @7.5.0

Mishura4 (approved):

braindigitalis:
merged but perhaps we need to find a better way to do this. it takes so long now.

PR #1616, opened 2026-07-21
declare heavy special members out of line, and proper move semantics

erics118:
moves special member functions out of line for faster builds, so these constructors will no longer be made once per translation unit

this is especially important for `confirmation_callback_t` and the `confirmable_t` variant

We define copy, move, copy assignment, move assignment, destructor operations out of line for:

we add proper move semantics for:
- user
- message
- command_option
- channel
- guild
- welcome_screen (part of guild)
- component
- embed
- interaction
- interaction_response
- slashcommand

change instances of `virtual ~class_name()` to `~class_name() override`

these are excluded from doxygen b/c these functions are quite verbose and don't add anything to the documentation

for my bot, with unity disabled, builds went from 18sec to 8sec

## Code change checklist


src/dpp/restrequest.cpp
```diff
@@ -0,0 +1,14 @@
+#include <dpp/restrequest.h>
+
+namespace dpp {
+
+#ifndef _DOXYGEN_
```
braindigitalis: i dont think these ifndef are required. doxygen is not configured to parse the .cpp files only headers. even then, why do you want to hide them from doxygen?

include/dpp/channel.h
```diff
@@ -480,6 +480,11 @@ class DPP_EXPORT channel : public managed, public json_interface<channel> {
 	/** Constructor */
 	channel();
 
+#ifndef _DOXYGEN_
```
braindigitalis: why are we hiding the ctors from docs?

braindigitalis (changes requested):

PR #1620, opened 2026-07-23
move confirmation_callback_t constructors out of line and fix move semantics

erics118:
TLDR: user-declared dtors were suppressing implicit moves, so the "moves" would copy data. this pr restores proper move semantics without removing the virtual dtors, and defines certain big class special members out of line for a much faster build (for me, 41.8s->12.6s)

---

previously, almost every data class declared a destructor (usually of the form `virtual ~T() = default`). the existence of this user-declared dtor suppresses the implicit move constructor/move assignment, so these "moves" have been copies.

this PR fixes it to have proper move semantics, and moves the definitions out of line to reduce codegen (they are now generated once, instead of in every TU).

all classes still keep their virtual destructors, so deriving from them and deleting through a base pointer is the same as before.

specifically:

- `json_interface` only types: ban, invite, prune, dtemplate, voiceregion, voicestate, auditlog, guild_command_permissions, onboarding, application_role_connection and its nested metadata, automod_metadata
  - the virtual dtor stays, and we declare the full rule of 5 with `default` to get proper moves back
- `managed` derived types: application, integration, stage_instance, sticker, user_identified, automod_rule
  - they already have a virtual dtor via `managed`, so the extra dtor decl is redundant. removing it re-enables the implicit moves.

by doing this, every `confirmable_t` alternative is nothrow moveable (except on msvc, see below), so `confirmation_callback_t` is also nothrow moveable, and the REST results and coro plumbing properly move instead of copy.

for the noexcept specs, two c++17 facts collide:

- an out-of-line defaulted member does not get a computed exception specification: callers only see what the header declares, so the noexcept must be spelled on the declaration, and it must match what the compiler would have computed or the defaulted definition is ill-formed.
- msvc's `std::map` move can throw, unlike libstdc++/libc++, so a type holding a map can't promise unconditional noexcept.

so:

- big types: message, embed, component, interaction, interaction_response, guild, confirmation_callback_t
  - define their special members out of line for the consumer compile speedup
  - the ones with only nothrow-movable members (embed, component) declare plain `noexcept`
  - the ones holding `std::map` directly or transitively (message via `attached_poll`, interaction via `resolved` its integration-owners map and its `msg`, guild via `voice_members`, interaction_response via its `message`, confirmation_callback_t via `confirmable_t`) declare a conditional spec mirroring what the compiler would compute
    - for example, we have the line `message(message&&) noexcept(std::is_nothrow_move_constructible_v<std::optional<poll>>))`: nothrow on gcc/clang, potentially throwing on msvc, valid everywhere
- smaller types: command_option, slashcommand, channel, user, welcome_screen, and all the small rule-of-5 types
  - keep in-class `= default` copy/move members. in-class defaulting gets the computed exception specification automatically, so the compiler produces the right spec per platform with nothing spelled out
  - virtual destructors for command_option, slashcommand, channel, user and welcome_screen still live out of line, since an out-of-line dtor is the class's key function and pins the vtable and typeinfo to a single TU.

I also ran some benchmarks by compiling my bot, to see whether this out-of-lining had benefits (avg of 3 runs):
- before this pr: 41.8s
- this pr, but with no out-of-lining: 38.4s
- this pr: 12.6s

## Code change checklist

### DIFF
--- a/include/dpp/appcommand.h
+++ b/include/dpp/appcommand.h
@@ -1156,16 +1156,16 @@ public:
 	interaction();
 
 	/** Copy constructor */
-	interaction(const interaction&) = default;
+	interaction(const interaction&);
 
 	/** Move constructor */
-	interaction(interaction&&) = default;
+	interaction(interaction&&) noexcept(std::is_nothrow_move_constructible_v<command_resolved> && std::is_nothrow_move_constructible_v<message> && std::is_nothrow_move_constructible_v<std::map<application_integration_types, snowflake>>);
 
 	/** Copy assignment operator */
-	interaction& operator=(const interaction&) = default;
+	interaction& operator=(const interaction&);
 
 	/** Move assignment operator */
-	interaction& operator=(interaction&&) = default;
+	interaction& operator=(interaction&&) noexcept(std::is_nothrow_move_assignable_v<command_resolved> && std::is_nothrow_move_assignable_v<message> && std::is_nothrow_move_assignable_v<std::map<application_integration_types, snowflake>>);
 
 	/**
 	 * @brief Destroy the interaction object
--- a/include/dpp/appcommand.h
+++ b/include/dpp/appcommand.h
@@ -30,6 +30,7 @@
 #include <dpp/entitlement.h>
 #include <variant>
 #include <map>
+#include <type_traits>
 #include <dpp/json_fwd.h>
 #include <dpp/json_interface.h>
 
@@ -525,9 +526,9 @@ public:
 	 */
 	interaction_response() = default;
 	interaction_response(const interaction_response&);
-	interaction_response(interaction_response&&) noexcept;
+	interaction_response(interaction_response&&) noexcept(std::is_nothrow_move_constructible_v<message>);
 	interaction_response& operator=(const interaction_response&);
-	interaction_response& operator=(interaction_response&&) noexcept;
+	interaction_response& operator=(interaction_response&&) noexcept(std::is_nothrow_move_assignable_v<message>);
 
 	/**
 	 * @brief Construct a new interaction response object
--- a/include/dpp/appcommand.h
+++ b/include/dpp/appcommand.h
@@ -1442,6 +1442,11 @@ public:
 	 */
 	guild_command_permissions();
 
+	guild_command_permissions(const guild_command_permissions&) = default;
+	guild_command_permissions(guild_command_permissions&&) = default;
+	guild_command_permissions& operator=(const guild_command_permissions&) = default;
+	guild_command_permissions& operator=(guild_command_permissions&&) = default;
+
 	virtual ~guild_command_permissions() = default;
 
 };
--- a/include/dpp/appcommand.h
+++ b/include/dpp/appcommand.h
@@ -302,10 +302,15 @@ public:
 	 */
 	command_option() = default;
 
+	command_option(const command_option&);
+	command_option(command_option&&) noexcept;
+	command_option& operator=(const command_option&);
+	command_option& operator=(command_option&&) noexcept;
+
 	/**
 	 * @brief Destroy the command option object
 	 */
-	virtual ~command_option() = default;
+	virtual ~command_option();
 
 	/**
 	 * @brief Add a file type (dpp::co_attachment) to filter for.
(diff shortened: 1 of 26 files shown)


include/dpp/appcommand.h
```diff
@@ -506,6 +511,10 @@ struct DPP_EXPORT interaction_response : public json_interface<interaction_respo
 	 * @brief Construct a new interaction response object
 	 */
 	interaction_response() = default;
+	interaction_response(const interaction_response&);
+	interaction_response(interaction_response&&);
+	interaction_response& operator=(const interaction_response&);
```
braindigitalis, line 543: documentation needed

include/dpp/appcommand.h
```diff
@@ -1117,10 +1126,15 @@ class DPP_EXPORT interaction : public managed, public json_interface<interaction
 	 */
 	interaction();
 
+	interaction(const interaction&) = default;
```
braindigitalis: documentation needed

include/dpp/appcommand.h
```diff
@@ -1415,6 +1429,11 @@ class DPP_EXPORT guild_command_permissions : public json_interface<guild_command
 	 */
 	guild_command_permissions();
```
braindigitalis, line 1467: documentation needed

include/dpp/appcommand.h
```diff
@@ -1561,10 +1580,15 @@ class DPP_EXPORT slashcommand : public managed, public json_interface<slashcomma
 	 */
 	slashcommand(const std::string &_name, const slashcommand_contextmenu_type _type, const dpp::snowflake _application_id);
 
+	slashcommand(const slashcommand&) = default;
```
braindigitalis, line 1627: documentation needed

include/dpp/automod.h
```diff
@@ -247,6 +247,11 @@ struct DPP_EXPORT automod_metadata : public json_interface<automod_metadata> {
 	 */
 	automod_metadata();
 
+	automod_metadata(const automod_metadata&) = default;
```
braindigitalis, line 251: documentation needed

include/dpp/ban.h
```diff
@@ -57,6 +57,11 @@ class DPP_EXPORT ban : public json_interface<ban> {
 	/** Constructor */
 	ban();
 
+	ban(const ban&) = default;
```
braindigitalis, line 61: documentation needed

include/dpp/integration.h
```diff
@@ -234,9 +234,6 @@ class DPP_EXPORT integration : public managed, public json_interface<integration
 	/** Default constructor */
 	integration();
 
-	/** Default destructor */
-	~integration() = default;
```
braindigitalis, line 238: if it is managed a user could choose to put it in a cache. without the virtual dtor, that becomes ub.

include/dpp/message.h
```diff
@@ -1617,8 +1627,6 @@ struct DPP_EXPORT sticker : public managed, public json_interface<sticker> {
 	 */
 	sticker();
 
-	virtual ~sticker() = default;
```
braindigitalis, line 1633: same here user could choose to cache it with a custom cache. that's the point of `managed`

include/dpp/prune.h
```diff
@@ -47,6 +47,13 @@ struct DPP_EXPORT prune : public json_interface<prune> {
 	virtual json to_json_impl(bool with_prune_count = false) const;
 
 public:
+	prune() = default;
+
+	prune(const prune&) = default;
```
braindigitalis, line 55: documentation needed

include/dpp/restresults.h
```diff
@@ -278,7 +279,13 @@ struct DPP_EXPORT confirmation_callback_t {
 	/**
 	 * @brief Construct a new confirmation callback t object.
 	 */
-	confirmation_callback_t() = default;
+	confirmation_callback_t();
+
+	confirmation_callback_t(const confirmation_callback_t&);
```
braindigitalis, line 285: documentation needed

braindigitalis (changes requested):
needs comments, and we can't really just get rid of the destructors as it breaks the expectations of the managed base class.

erics118:
ill fix docs and other stuff once i figure out what actually works

include/dpp/integration.h
erics118: the virtual dtor is still there, just inherited because this class derives from `managed`. I removed the decl so that we don't need to declare the rule of 5 functions for the move semantics that we want

include/dpp/message.h
erics118: the virtual dtor is still there, just inherited because this class derives from `managed`. I removed the decl so that we don't need to declare the rule of 5 functions for the move semantics that we want

braindigitalis (approved):

Jaskowicz1 (approved):

PR #1622, opened 2026-08-01
fix: Bound decode_compressed to the length zlib actually wrote

Mounika2456:
decode_compressed trusts the uncompressed size declared in the term header instead of the length uncompress() reports back, so a term that declares more than its zlib stream produces has the remainder decoded out of heap that was never written.

An ett_compressed term declaring 4096 bytes whose stream deflates to five (an ett_binary header announcing 4080 bytes of payload) comes back as a 4080 character string assembled from that memory; under ASAN the same input trips container-overflow in read_8_bits, since reserve() leaves the vector empty. zlibcontext::decompress already sizes its buffer and counts what inflate wrote, so this brings the term decoder in line with it. Covered by a new offline test that fails on master.

## Code change checklist

### DIFF
--- a/src/dpp/etf.cpp
+++ b/src/dpp/etf.cpp
@@ -492,21 +492,22 @@ json etf_parser::decode_tuple_large() {
 
 json etf_parser::decode_compressed() {
 	const uint32_t uncompressedSize = read_32_bits();
-	unsigned long sourceSize = uncompressedSize;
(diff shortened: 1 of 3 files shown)


Mounika2456:
Superseded by #1623, which is this same branch raised against `dev`. GitHub won't let me reopen this one while that is open, so leaving it closed.

PR #1631, opened 2026-08-01
fix: race condition when destroying voice client

Neko-Life:
# Changes in this PR

In small bot where concurrent reconnection never goes over just a few reconnection the probability of a crash is very tiny.
Tested in Musicat with exactly 41 concurrent reconnection.

- Fix intermittent crash when voiceconn decides to reset voiceclient

## Code change checklist

### DIFF
--- a/src/dpp/discordclient.cpp
+++ b/src/dpp/discordclient.cpp
@@ -596,15 +596,22 @@ bool voiceconn::is_active() const {
 	return voiceclient != nullptr;
 }
 
-voiceconn& voiceconn::disconnect() {
+voiceconn& voiceconn::disconnect(bool close) {
 	if (this->is_active()) {
-		voiceclient.reset();
+		this->voiceclient->terminating = true;
+		this->voiceclient->pause_audio (true);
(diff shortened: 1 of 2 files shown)


src/dpp/discordclient.cpp
```diff
@@ -596,15 +596,22 @@ bool voiceconn::is_active() const {
 	return voiceclient != nullptr;
 }
 
-voiceconn& voiceconn::disconnect() {
+voiceconn& voiceconn::disconnect(bool close) {
 	if (this->is_active()) {
-		voiceclient.reset();
+		this->voiceclient->terminating = true;
+		this->voiceclient->pause_audio (true);
```
braindigitalis: only thing I would change here is the weird styling where the prams have a space before the brackets

braindigitalis (changes requested):

src/dpp/discordclient.cpp
Neko-Life: whoops right, it's gnu style btw

braindigitalis (approved):

PR #1632, opened 2026-08-05
fix: tighten up shell escaping to handle some edge cases

braindigitalis:
When passing command line parameters to a bourne style shell using `dpp::utility::exec()`, the user could be misled to believe DPP does escaping of \` symbols, `$()`, etc (it did not).

This PR adds support for escaping that properly blocks those bourne behaviours in Linux and other similar operating systems, so that it is a bit more friendly.

Note, this does not absolve the library user of NOT passing user crafted data into `dpp::utility::exec()`. It is not our job to protect the library users from stupidity.

Windows doesn't need this patch, as `std::quoted()` will suffice there.

## Code change checklist

### DIFF
--- a/src/dpp/utility.cpp
+++ b/src/dpp/utility.cpp
@@ -451,6 +451,18 @@ uint32_t hsl(double h, double s, double l) {
 	return rgb(r, g, b);
 }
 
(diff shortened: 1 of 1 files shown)


Neko-Life (approved):
lgmt

PR #1635, opened 2026-08-06
fix: bound RTP header offsets in voice_courier_loop to packet size

Mounika2456:
ASan, replaying the courier's header block against a 44 byte datagram whose first byte is 0x1F (extension bit set, CSRC count 15), the shortest packet `read_ready()` forwards:

    ciphertext_len = 18446744073709551584

    ERROR: AddressSanitizer: heap-buffer-overflow
    READ of size 2 at 0x60400000041a thread T0
        #0 __asan_memcpy
        #1 in <courier header block> ... memcpy(&ext_len_in_words, &ciphertext[2], sizeof(uint16_t))
    0x60400000041a is located 30 bytes after 44-byte region [0x6040000003d0,0x6040000003fc)

`offset_to_data` is `12 + 4 * csrc_count` and reaches 72, while `read_ready()` only guarantees 44 bytes, so `ciphertext_len` wraps, the extension length is read off the end of the vector holding the datagram, and `total_header_len` walks the AEAD past that same allocation for its AAD. The voice UDP socket is bound without a `connect()` and read with `recv()`, so the sender is any host that can reach the port, and none of this sits behind the Poly1305 check. Drop the packet when the header its first byte describes does not fit in what actually arrived.

## Code change checklist

### DIFF
--- a/src/dpp/voice/enabled/courier_loop.cpp
+++ b/src/dpp/voice/enabled/courier_loop.cpp
@@ -186,8 +186,12 @@ void discord_voice_client::voice_courier_loop(discord_voice_client& client, cour
 						 * trailing nonce, otherwise ciphertext_len below underflows.
 						 */
 						if (packet_size < static_cast<size_t>(offset_to_data) + nonce_size) {
-							/* Invalid Discord RTP payload. */
-							return;
+							/*
+							 * Invalid Discord RTP payload. Consume it and move on to the
+							 * next parked payload, like the pop at the end of this iteration.
+							 */
+							d.parked_payloads.pop();
+							continue;
 						}
 
 						size_t total_header_len = offset_to_data;
@@ -202,8 +206,12 @@ void discord_voice_client::voice_courier_loop(discord_voice_client& client, cour
 							 * it is present before reading it and subtracting it from ciphertext_len.
 							 */
 							if (ciphertext_len < sizeof(uint16_t) * 2) {
-								/* Invalid Discord RTP payload. */
-								return;
+								/*
+								 * Invalid Discord RTP payload. Consume it and move on to the
+								 * next parked payload, like the pop at the end of this iteration.
+								 */
+								d.parked_payloads.pop();
+								continue;
 							}
 							/**
 							 * Get the RTP Extensions size, we only get the size here because
--- a/src/dpp/voice/enabled/courier_loop.cpp
+++ b/src/dpp/voice/enabled/courier_loop.cpp
@@ -179,6 +179,17 @@ void discord_voice_client::voice_courier_loop(discord_voice_client& client, cour
 						const size_t csrc_count = buffer[0] & 0b0000'1111;
 						/* Skip to the encrypted voice data */
 						const ptrdiff_t offset_to_data = header_size + sizeof(uint32_t) * csrc_count;
+
+						/*
+						 * csrc_count comes from the first byte of the datagram, so the packet
+						 * has to actually be long enough to hold the header it describes plus the
+						 * trailing nonce, otherwise ciphertext_len below underflows.
+						 */
+						if (packet_size < static_cast<size_t>(offset_to_data) + nonce_size) {
+							/* Invalid Discord RTP payload. */
+							return;
+						}
+
 						size_t total_header_len = offset_to_data;
 
 						uint8_t *ciphertext = buffer + offset_to_data;
@@ -186,6 +197,14 @@ void discord_voice_client::voice_courier_loop(discord_voice_client& client, cour
 
 						size_t ext_len = 0;
 						if ([[maybe_unused]] const bool uses_extension = (buffer[0] >> 4) & 0b0001) {
+							/*
+							 * The 4 byte extension header sits past offset_to_data, so make sure
+							 * it is present before reading it and subtracting it from ciphertext_len.
+							 */
+							if (ciphertext_len < sizeof(uint16_t) * 2) {
+								/* Invalid Discord RTP payload. */
+								return;
+							}
 							/**
 							 * Get the RTP Extensions size, we only get the size here because
 							 * the extension itself is encrypted along with the opus packet


braindigitalis:
are these the result of real usage? or a theoretical output?
I'm struggling to make sense of what you wrote and how that relates to the fix, does it still send  and receive voice output correctly and work with DAVE?

braindigitalis:
I’m not accepting this in its current form. The stated issue only requires a small bounds check, but this patch changes types, moves unrelated logic and introduces a new parked_payloads.pop() path without explaining the queue semantics. Please reduce it to the minimal validation necessary, preserve the existing control flow, and provide evidence that normal voice receive and DAVE receive were tested.

Mounika2456:
Fair, that was too invasive. I've cut it down to just the two length checks and left everything else alone: `offset_to_data` is back to `ptrdiff_t`, the `uses_extension`/`ext_header_len` declarations are untouched, and there's no new `pop()`. Both guards `return;` the same way the existing "Invalid Discord RTP payload" path does, so the control flow doesn't change. The diff is now 19 added lines with no deletions.

The ASan numbers aren't from live traffic. They're from a small harness that runs this function's header block verbatim against the shortest datagram `read_ready()` forwards: 44 bytes with the first byte 0x1F, so csrc_count is 15 and the extension bit is set. `offset_to_data` then works out to 72 while only 44 bytes arrived, so `ciphertext_len = packet_size - offset_to_data - nonce_size` wraps to ~2^64 and the extension-length read at `ciphertext[2]` lands 30 bytes past the vector. The guard drops that packet before any of those offsets are used.

A well-formed packet takes the identical path, so normal receive is unaffected, and DAVE runs after the AEAD step on the decrypted `opus_packet`, which for a valid packet still gets the same `ciphertext`/`ciphertext_len`/`total_header_len` as before. For evidence: the library builds clean with `-DDPP_NO_VCPKG=ON -DCMAKE_BUILD_TYPE=Release -DAVX_TYPE=AVX0` (no new warnings) and the offline unittest is Failed: 0 Passed: 64. Under ASan the crafted packet is now dropped by the guard with no report, and a normal packet (csrc 0, no extension) still passes through with `offset_to_data=12`, `ciphertext_len=28`, the same values it computed before. I don't have a live channel with a second speaker and a DAVE ratchet to exercise offline, so an end to end confirmation on a real voice session is the one thing I can't reproduce here, but the valid-packet path isn't changed by this.

braindigitalis:
the unit tests don't test and can't test voice receive.

but if discord isn't sending this, and can't send this, how could we test it properly? making arbitrary shims to test data discord never sends (always one csrc) seems like it isn't going to get us much, you're testing things the library doesn't support.

Mounika2456:
Agreed on the test side, a shim that fakes csrc counts isn't worth adding and I'm not asking to add one. The unittest can't reach voice receive, so that's not where this belongs.

Where I think the framing slips is "discord isn't sending this, and can't send this." The bytes that drive offset_to_data don't have to come from Discord. The voice UDP socket is created unconnected (bound to INADDR_ANY, no connect()) and read with recv(), and recv() on an unconnected socket takes a datagram from any source that can reach the port. read_ready() only rejects packet_size < 44 and the RTCP payload types, so a crafted 44 byte datagram with the first byte set to 0x1F reaches the courier, and all of this header math runs before the Poly1305 tag is checked. So the csrc_count of 15 isn't Discord sending something odd, it's an unauthenticated sender setting four bits.

The guard isn't trying to support multi-csrc packets. For a real Discord packet (one csrc, header fits) the check is always true and the path is byte for byte identical, nothing changes. It only fires when the header the first byte declares doesn't fit in the bytes that actually arrived, which is exactly the case that underflows ciphertext_len and reads the extension length off the end of the vector today. It's the same shape as the existing packet_size < 44 reject, just covering the offset the csrc bits can push it to.

braindigitalis:
and how is the unauthenticated sender going to know which port or your bots ip? what about nat? and even then, what will they do? I dont see a crash here.
but bot ips aren't supposed to be publicly known, if they were that's a huge ddos risk.

Mounika2456:
On the crash: it's a heap-buffer-overflow read, not just odd numbers. With csrc_count 15 the extension-length read at `ciphertext[2]` lands 30 bytes past the 44 byte vector, and `total_header_len` (72) is handed to the AEAD as the AAD length over that same 44 byte allocation, so OpenSSL reads about 28 bytes past the end too. ASan catches the first read; the second is silent since OpenSSL isn't instrumented. So it's a genuine out-of-bounds read driven by attacker-influenced length math, it just doesn't reliably segfault.

You're right on reachability though. An attacker needs the bot's public IP and the current ephemeral UDP port, and has to land a datagram through NAT, so this isn't spray-the-internet and I won't pretend it's high severity. Where I'd still place it is robustness: it's the same shape as the existing `packet_size < 44` reject, just covering the offset the csrc bits can push the header to, so a short or malformed packet gets dropped instead of underflowing `ciphertext_len`. The guard is cheap and doesn't touch the valid-packet path, but if you'd rather not carry it given how hard it is to reach, that's a fair call.

braindigitalis:
sure but as before, we can't merge anything until the boxes are ticked in the pr description and the cla contributor covenant is done.

Mounika2456:
Makes sense. I'll get the CLA signed through cla-assistant today so that check clears. On the checklist, four of the five boxes have been ticked since the PR went up; the one I left open is the tooling attestation, since I did use ASan and a bit of light tool assistance for research and sanity checks while running this down, and I didn't want to tick it as an absolute. The diff itself was written and tested by hand, and the guard's behavior on valid packets is as described above.

braindigitalis (approved):

braindigitalis (changes requested):
changes needed, see comments

src/dpp/voice/enabled/courier_loop.cpp
```diff
@@ -179,13 +179,32 @@ void discord_voice_client::voice_courier_loop(discord_voice_client& client, cour
...
 
 						size_t ext_len = 0;
 						if ([[maybe_unused]] const bool uses_extension = (buffer[0] >> 4) & 0b0001) {
+							/*
+							 * The 4 byte extension header sits past offset_to_data, so make sure
+							 * it is present before reading it and subtracting it from ciphertext_len.
+							 */
+							if (ciphertext_len < sizeof(uint16_t) * 2) {
+								/* Invalid Discord RTP payload. */
+								return;
```
braindigitalis: this actually causes the whole courier thread to terminate! this hasn't been tested.

src/dpp/voice/enabled/courier_loop.cpp
```diff
@@ -179,13 +179,32 @@ void discord_voice_client::voice_courier_loop(discord_voice_client& client, cour
...
 						const ptrdiff_t offset_to_data = header_size + sizeof(uint32_t) * csrc_count;
+
+						/*
+						 * csrc_count comes from the first byte of the datagram, so the packet
+						 * has to actually be long enough to hold the header it describes plus the
+						 * trailing nonce, otherwise ciphertext_len below underflows.
+						 */
+						if (packet_size < static_cast<size_t>(offset_to_data) + nonce_size) {
+							/* Invalid Discord RTP payload. */
+							return;
```
braindigitalis: causes thread termination.

Mounika2456: Good catch, that return was wrong. Swapped both guards to pop the offending payload and continue, so the malformed packet is dropped and the loop moves on to the next one instead of tearing down the thread. It matches the pop() at the end of the normal iteration.

src/dpp/voice/enabled/courier_loop.cpp
Mounika2456: Same fix here, this one now pops and continues too rather than returning. So a packet whose extension header doesn't fit gets dropped and the courier keeps draining the rest of the queue.

Mounika2456:
Pushed the fix for both guards. To confirm the thread doesn't die, I ran the courier's inner drain loop with a two-entry queue: seq 0 is the malformed 44-byte packet (first byte 0x1F, so offset_to_data works out to 72) and seq 1 is a well-formed one. Under ASan seq 0 is now popped and skipped by the first guard with no out-of-bounds read, the loop keeps going, and seq 1 is processed with the same offsets as before. Library builds clean with the same flags and the offline unittest is still Failed: 0 Passed: 64.

src/dpp/voice/enabled/courier_loop.cpp
braindigitalis: have you any way of testing this with real discord clients

PR #1639, opened 2026-08-09
[fix] Made shardless cluster not cause an assertion error when getting destroyed

Henonicks:
After `cluster.cpp:162` happens, executing `cluster.cpp:163` causes the assertion fail. This also works the other way around.

Shutting down the cluster (`dpp::cluster::shutdown`) before `cluster.cpp:162` or the destruction of the object makes the assertion fail go away.

The issue isn't caused by `dpp::request_queue::~request_queue` directly. This destructor marks queues as terminating which doesn't let new requests come in. Yet after `dpp::cluster::tick_timers` has already been called so it needs to, well, tick all the timers. The timer queue can end up depleting before `tick_timers` finishes its job since in the `while (true)` loop (matching `do` at `timer.cpp:66`) we pop the queue and never check if it's empty. And if it's supposed to be unable to, well, that's exactly what caused this issue in the first place. The only place in that loop's function where we do check is above the `do`, at `timer.cpp:62`. My guess is that these checks happen to be quick enough when there are no shards to cause that assertion fail before the cluster gets to shut down, though I'm not that familiar with the codebase to be able to assert a confident verdict.

Either way, this PR fixes it by simply checking for the `next_timer`'s emptiness before accessing its `.top()`. Since the original check virtually instantly exits `tick_timers` if it succeeds and otherwise accesses the queue anyway, my addition should have no effect on a bot's functionality. Regardless, I tested it with mine for a few minutes and it ran fine.

I decided to maintain the order of the calls in `dpp::cluster`'s destructor in case it was important for graceful exiting.

## Code change checklist

### DIFF
--- a/src/dpp/cluster/timer.cpp
+++ b/src/dpp/cluster/timer.cpp
@@ -67,7 +67,7 @@ void cluster::tick_timers() {
 		timer_t cur_timer;
 		{
 			std::lock_guard<std::mutex> l(timer_guard);
-			if (next_timer.top().next_tick > now) {
+			if (next_timer.empty() || next_timer.top().next_tick > now) {
 				/* Nothing to do */
 				break;
 			}


Jaskowicz1:
Im wonder if it's worth just changing the condition of the while to be `!next_timer.empty()` but unsure on the safety of that as we couldn't capture lock.

Henonicks:
> Im wonder if it's worth just changing the condition of the while to be `!next_timer.empty()` but unsure on the safety of that as we couldn't capture lock.

Since `std::priority_queue::empty` is really just `<underlying_container>::empty()`, it's as safe as "If not currently empty, acquire lock and do stuff with it (might be empty when we do so)"

Is what I just mentioned safe? At the moment, the only places where we pop/empty the queue are:
- In `dpp::cluster::shutdown`
- Below the `if` that I modified (only executes if the check fails)

I modified the code by undoing my change and replacing the `while (true)` with `while (!next_timer.empty())`.
Then I ran this:
```cpp
dpp::cluster bot("", 0, dpp::NO_SHARDS);
bot.start(dpp::st_return);
std::this_thread::sleep_for(std::chrono::seconds(5));
bot.shutdown();
```
With tsan, a few times, and it showed nothing.

So with the whole picture in mind, it seems safe, but I'm not too sure about that.

Jaskowicz1:
What you've done in this pr is safe and fine yeah, i'd be happy to approve as is, i'm just thinking about the while thing because it means we could probs remove https://github.com/brainboxdotcc/DPP/pull/1639/changes#diff-f71e736e2abd27d3494bf5ddaaa251a462df6a05e1405a4e53993c1363b6e0c4R60-R65

but I would prefer the changes already proposed by you in this PR

Jaskowicz1 (approved):

braindigitalis (approved):

PR #1643, opened 2026-08-28
bound ip discovery address to buffer size before string conversion

Mounika2456:
discover_ip() reads the IP discovery response off the voice UDP socket into a 74 byte buffer, copies the 64 byte address field into ip_discovery_packet::address, then returns it with std::string(inbound_packet.address). That constructor runs strlen across the buffer, but nothing in the discovery wire format guarantees a NUL inside those 64 bytes. A response whose whole address region is non-zero makes strlen walk past the end of the fixed buffer; ASAN reports a heap-buffer-overflow read of size 65 one byte past the 64 byte region, and whatever it reads lands in the external IP we hand back in the select_protocol payload.

Bound the read with strnlen(address, ADDRESS_BUFFER_SIZE) so the returned string can never span more than the 64 bytes actually received. Building the bound where the string is constructed leaves the packet struct untouched and stops the over-read at the single place the buffer is treated as a C string; a well formed, NUL-terminated address decodes exactly as before.

## Code change checklist

### DIFF
--- a/src/dpp/voice/enabled/discover_ip.cpp
+++ b/src/dpp/voice/enabled/discover_ip.cpp
@@ -168,7 +168,7 @@ std::string discord_voice_client::discover_ip() {
 					return "";
 				}
 				ip_discovery_packet inbound_packet(buffer);
-				return {inbound_packet.address};
+				return std::string(inbound_packet.address, strnlen(inbound_packet.address, ip_discovery_packet::ADDRESS_BUFFER_SIZE));
 		}
 	}
 	return {};


braindigitalis:
hold up this doesn't make sense.

char address[ADDRESS_BUFFER_SIZE]{0}; // NOLINT

the buffer is bigger than any possible returned address, and discord does NOT return a buffer with garbage on the end, in short it is always null terminated.

braindigitalis:
"nothing in the discovery wire format guarantees a NUL inside those 64 bytes"

where do you get this? without a content length field how else would it work?

braindigitalis:
https://docs.discord.com/developers/topics/voice-connections#ip-discovery

```
Address	    |      Null-terminated string in response	|        64 bytes
```

discord is not an adversarial service that will send junk trying to attack bots. this isn't like irc or a random website.

PR #1650, opened 2026-09-09
cast to unsigned char before tolower/toupper/isdigit calls

Mounika2456:
`dpp::lowercase`/`dpp::uppercase` hand the C `tolower`/`toupper` straight to `std::transform` over a `std::string`, so every element is a `char` widened to `int`:

    std::transform(s2.begin(), s2.end(), s2.begin(), tolower);

`tolower`/`toupper` are only defined for arguments representable as `unsigned char` or `EOF`. Where `char` is signed, any byte >= 0x80 arrives as a negative value, which is out of that contract. These helpers normalize untrusted input: prefix command names in `commandhandler`/`slashcommand`, and the request line plus header keys parsed by the interactions `http_server`. A UTF-8 name like `café` or a header key with a high byte is enough to reach the call with a negative argument. In practice glibc, musl and Apple libc all tolerate that (glibc's table covers -128..255, the others range check), so this is a conformance cleanup rather than a crash. The one documented failure is the MSVC debug CRT, which asserts `c >= -1 && c <= 255` in the `is*` functions.

Two `isdigit(*str.c_str())` calls in `cluster/confirmation.cpp` (error field-name handling) have the same problem.

The `isdigit` sites cast to `unsigned char`, matching the fix already applied to the `tolower` lambdas in `httpsclient.cpp`. `lowercase`/`uppercase` are templated on the character type, so their lambdas take `T` and convert through `std::make_unsigned_t<T>`, which gives the same result for `char` without narrowing wider types. Valid input is unaffected; bytes >= 0x80 still pass through unchanged. Built against `dev` and the offline unit tests pass.

## Code change checklist


braindigitalis:
have you tested this in real world conditions, not in a fabricated shim?

braindigitalis:
Depending on the libc that indexes outside the ctype table, and the MSVC debug CRT asserts on it.

depending how? this feels like an assumption.

include/dpp/stringops.h
```diff
@@ -42,7 +43,7 @@ namespace dpp {
 template <typename T> std::basic_string<T> lowercase(const std::basic_string<T>& s)
 {
 	std::basic_string<T> s2 = s;
-	std::transform(s2.begin(), s2.end(), s2.begin(), tolower);
+	std::transform(s2.begin(), s2.end(), s2.begin(), [](unsigned char c){ return std::tolower(c); });
```
braindigitalis: this should be `T c`, not `unsigned char c`.

I know this pees in the cornflakes of the pr intent somewhat but lowercase is templated to take the contained char type of basic_string. this would break it.

include/dpp/stringops.h
```diff
@@ -56,7 +57,7 @@ template <typename T> std::basic_string<T> lowercase(const std::basic_string<T>&
 template <typename T> std::basic_string<T> uppercase(const std::basic_string<T>& s)
 {
 	std::basic_string<T> s2 = s;
-	std::transform(s2.begin(), s2.end(), s2.begin(), toupper);
+	std::transform(s2.begin(), s2.end(), s2.begin(), [](unsigned char c){ return std::toupper(c); });
```
braindigitalis: same here as above

braindigitalis (changes requested):

include/dpp/stringops.h
Mounika2456: Good catch, that did break it. I checked with a `std::wstring` holding U+0141 and U+0100 and the `unsigned char` parameter turned them into `a` and NUL.

The lambda takes `T c` now and converts through `std::make_unsigned_t<T>` before the call. For `char` a byte like 0xC3 still reaches `tolower` as 195 rather than -61, and wider types are no longer narrowed. `wstring`, `u16string` and `u32string` output matches dev again.

include/dpp/stringops.h
Mounika2456: Same change in `uppercase`, it takes `T c` as well.

Mounika2456:
On the two earlier questions, which I should have answered sooner.

Real world testing: not against a live gateway, no. What I've run is the library built from this branch with a `dpp::http_server` on a shardless cluster, and curl sending a header key containing bytes 0xC3 0xA9 (`X-CAFÉ`) plus a lowercase `get` verb. The header is stored and found under the lowercased key and the verb is accepted, same as dev. The offline unittest is Failed: 0 Passed: 64.

"Depending how": fair, that sentence claimed more than I verified. What I actually confirmed is that where `char` is signed the byte 0xC3 reaches `tolower` as -61, which is outside the range the C standard defines it for. I haven't seen that misbehave on any libc I could check. glibc's table deliberately covers -128..255, musl does a range compare, and Apple's returns the negative value unchanged (ran that one here). The only documented failure I know of is the MSVC debug CRT asserting `c >= -1 && c <= 255` in the `is*` functions. That would apply to the two `isdigit` sites, but those only see Discord's error field names, which are ASCII, and I haven't run it on MSVC myself.

So this is a conformance cleanup, not a crash anyone can hit today, and the PR description overstated it. I've corrected the wording there. If that isn't worth carrying, no objection to closing it.

braindigitalis (approved):

PR #1652, opened 2026-09-12
Added Mingw support

TracerDS:
[ Please outline your change here. Be sure to check the checkboxes below. ]
The current version of DPP does not support mingw environments (ucrt64, clang64).
This PR adds this support.

## Code change checklist


include/dpp/nlohmann/json.hpp
```diff
@@ -20277,23 +20277,25 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
         result["version"]["minor"] = NLOHMANN_JSON_VERSION_MINOR;
         result["version"]["patch"] = NLOHMANN_JSON_VERSION_PATCH;
 
-#ifdef _WIN32
+#if defined(DPP_USE_MINGW)
+        result["platform"] = "mingw";
+#elif defined(DPP_USE_WINDOWS)
```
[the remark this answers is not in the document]
braindigitalis: who the fuck invited you to this party copilot, fuck off

braindigitalis:
this pr is way too big. it touches so much stuff. how can we know it doesn't break something, it's diving in changing functionality outside the build system, renaming variables, moving includes.

is it tested? does it have working ci?

secondly, who is going to support it after it's merged? I for one don't really like mingw and it is another target to support, test in ci, run unit tests on.

thirdly who added all these ai reviews, copilot is NOT an authorised reviewer of dpp!

braindigitalis (changes requested):
this pr is far too big in its current form and changes too many things outside the build system.

we already support proper native windows builds so mingw as a crutch is not needed.

include/dpp/nlohmann/json.hpp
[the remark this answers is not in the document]
braindigitalis: actually what.... why have you changed a third party dependency?

braindigitalis (changes requested):

src/dpp/socket.cpp
```diff
@@ -74,7 +74,6 @@ template <typename T> bool raii_socket::set_option(int level, int name, T value)
 
 template bool raii_socket::set_option<int>(int, int, int);
 template bool raii_socket::set_option<bool>(int, int, bool);
-template bool raii_socket::set_option<socklen_t>(int, int, socklen_t);
```
braindigitalis, line 77: why remove it?

src/dpp/utility.cpp
```diff
@@ -127,13 +127,12 @@ bool is_coro_enabled() {
...
-#ifdef _WIN32
-	std::time_t curr_time = time(nullptr);
+	auto curr_time = std::time(nullptr);
+#ifdef DPP_USE_WINDOWS
 	return trim(std::ctime(&curr_time));
 #else
-	auto t = std::time(nullptr);
 	struct tm timedata;
-	localtime_r(&t, &timedata);
+	localtime_r(&curr_time, &timedata);
```
braindigitalis: why is this refactor part of the pr

src/dpp/httpsclient.cpp
```diff
@@ -61,7 +61,7 @@ void https_client::connect()
 		map_headers += k + ": " + v + "\r\n";
 	}
 
-	if (this->sfd != SOCKET_ERROR) {
+	if (this->sfd != INVALID_SOCKET) {
```
braindigitalis, line 64: why?

include/dpp/export.h
```diff
@@ -29,13 +31,13 @@
  * which is hugely non-standard, but apparently "it broke stuff" so they dont ever change it
  * from C++98. Ugh.
  */
-#if (!defined(__cplusplus) || (defined(_MSC_VER) && (!defined(_MSVC_LANG) || _MSVC_LANG < 201703L)) || (!defined(_MSC_VER) && __cplusplus < 201703L))
+#if !defined(DPP_USE_CPP) || DPP_USE_CPP_VERSION < 201703L
```
braindigitalis, line 34: are you absolutely sure this is doing the same thing?

include/dpp/nlohmann/json.hpp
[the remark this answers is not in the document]
TracerDS: > actually what.... why have you changed a third party dependency?

Whoopsie, my mistake. I went too bad with replacing `_WIN32` stuff. Will revert it

TracerDS:
> this pr is way too big. it touches so much stuff. how can we know it doesn't break something, it's diving in changing functionality outside the build system, renaming variables, moving includes.

It doesnt change any funcitonality, only CMakeLists is modified and made definitions rely on compiler macros. Basically the same thing as it was before but with more detections.

> is it tested? does it have working ci?

Its working fine.

> secondly, who is going to support it after it's merged? I for one don't really like mingw and it is another target to support, test in ci, run unit tests on.

You dont run unit tests to check compilers. You write unit tests to check the code. It will naturally be supported as no changes touched the actual codebase.

> thirdly who added all these ai reviews, copilot is NOT an authorised reviewer of dpp!

Its been added by default. Your repo, your settings 🙄

> we already support proper native windows builds so mingw as a crutch is not needed.

MSVC only is not a good support choice. Clang on windows works dubiously if at all.
Non MSVC clang (`clang++`) doesn't work at all,
`clang-cl` not only doesnt recognize windows syntax flags (`/Zc:preprocessor`) but it also as above doesnt compile even with MSVC compatible mode.

include/dpp/export.h
TracerDS: Yes. In `system.h` I already check for the `MSVC` inconsistency. `DPP_USE_CPP_VERSION` will always result in either `__cplusplus` (if non MSVC) or `_MSVC_LANG` if on MSVC

src/dpp/httpsclient.cpp
TracerDS: `SOCKET_ERROR` is not a socket type. Since `INVALID_SOCKET` is just `(SOCKET)SOCKET_ERROR` (not directly but its always true) then it should be compared with the same type instead of silent (actually verbose) cast.

src/dpp/socket.cpp
TracerDS: It causes issues as `socklen_t` = `int` on mingw toolchain.
So it causes redefinition
```cpp
template bool raii_socket::set_option<int>(int, int, int);
template bool raii_socket::set_option<bool>(int, int, bool);
template bool raii_socket::set_option<int>(int, int, int);
```

src/dpp/utility.cpp
TracerDS: I had some issues with the `current_date_time` function (the codeblock this change is in) but its been resolved and it looks like I forgot to revert these changes. Will revert them in the next commit

Jaskowicz1:
@braindigitalis See discord for discussion on Copilot, seems they've broken it again, not the fault of this review (from what I understand).

Mishura4:
> MSVC only is not a good support choice. Clang on windows works dubiously if at all.
Non MSVC clang (clang++) doesn't work at all,
clang-cl not only doesnt recognize windows syntax flags (/Zc:preprocessor) but it also as above doesnt compile even with MSVC compatible mode.

What are you talking about? We literally test clang-cl as part of our CI/CD pipeline.

src/dpp/socket.cpp
Mishura4: So you need to remove the int specialization then, not the socklen_t specialization.

Mishura4 (changes requested):
We will not **officially** support MinGW. None of the maintainers use it as there is almost no reason to with MSVC and clang-cl. We would have to be careful that none of our changes break it, and that's not something we can or are willing to do.

With that said I don't mind a PR that fixes it for now, but then we cannot have a tutorial that explains how to build it in MinGW, as we cannot officially support it. If it breaks in the future, then it breaks, and we can't care.

src/dpp/socket.cpp
TracerDS: ~~sure, either one will work.~~

The codebase uses explicit `int` template, e.g.
[include/dpp/socket_listener.h](https://github.com/brainboxdotcc/DPP/tree/master/include/dpp/socket_listener.h#L101)
```cpp
fd.set_option<int>(SOL_SOCKET, SO_REUSEADDR, 1);
```
[include/dpp/socket.h](https://github.com/brainboxdotcc/DPP/tree/master/include/dpp/socket.h#L184)
```cpp
extern template DPP_EXPORT bool raii_socket::set_option<int>(int level, int name, int value);
```

TracerDS:
> We will not officially support MinGW. None of the maintainers use it as there is almost no reason to with MSVC and clang-cl. We would have to be careful that none of our changes break it, and that's not something we can or are willing to do.

As stated above there is nothing to actually support. As long as the include checks compare the system macros (`DPP_USE_...`) then nothing will break. MSVC is actually not a good compiler for the actual use. Which leaves the other option of clang-cl but it has issues above. And you're obviously only limited to the VS toolchain.
Supporting mingw will also bring new contributors who dont rely on VS tools. The build process would also be simplified as its still just gcc/clang under the hood but with a different shell.

Mishura4:
You have been given reasons as to why we will not officially support MinGW, and you have chosen to ignore them. You have been given a proposition for a compromise, and you have chosen to ignore it as well.

Well, good luck.

TracerDS:
> You have been given reasons as to why we will not officially support MinGW, and you have chosen to ignore them. You have been given a proposition for a compromise, and you have chosen to ignore it as well.
> 
> Well, good luck.

??? What are you taking about? I've given reasons to support it and the only valid reason you said you won't is because current collaborators sit on msvc only. 

It is not my role to start supporting it or not, there is no compromise as I'm not the one that's making the decision.

Ashthetik:
> > We will not officially support MinGW. None of the maintainers use it as there is almost no reason to with MSVC and clang-cl. We would have to be careful that none of our changes break it, and that's not something we can or are willing to do.
> 
> As stated above there is nothing to actually support. As long as the include checks compare the system macros (`DPP_USE_...`) then nothing will break. MSVC is actually not a good compiler for the actual use. Which leaves the other option of clang-cl but it has issues above. And you're obviously only limited to the VS toolchain.
> Supporting mingw will also bring new contributors who dont rely on VS tools. The build process would also be simplified as its still just gcc/clang under the hood but with a different shell.

In what way is MSVC/clang-cl not a good compiler for windows? It's compliant to the WinAPI ABI, and has no issues with optimisations, linking, or building in general

I've personally found MinGW to hold extensive compatibility issues with Windows especially around networking since the Windows TCP stack is different to Unix/Linux. MinGW tends to also get flagged as suspicious/PUP because they don't follow the standard PE architecture without having to add rs and rec files to the project

TracerDS:
> In what way is MSVC/clang-cl not a good compiler for windows? It's compliant to the WinAPI ABI, and has no issues with optimisations, linking, or building in general
> 
> I've personally found MinGW to hold extensive compatibility issues with Windows especially around networking since the Windows TCP stack is different to Unix/Linux. MinGW tends to also get flagged as suspicious/PUP because they don't follow the standard PE architecture without having to add rs and rec files to the project

mingw is nowadays fully compatible with windows. It also has no issues with optimisations, linking or building in general. In fact, mingw is a better candidate for building stuff because it is basically a posix shell on windows so backporting anything non-windows to windows is trivial. It also compiles a fully native windows PE exec. So if they get flagged, we got one reason out of the picture. Resource files are also part of the windows resource mechanism, not a required thing to be conformant with PE execs.

MSVC is notoriously known for not following the standard, it requires you to have the exact same toolchain to work, you cant crosscompile.
`clang-cl` is just msvc in disguise.
You also have the deprecation by windows of using C functions and going over to their "safe variants" (`..._s`). The "has no issues with optimisations" is also a stretch. One example of such inconsistency can be found [here](https://www.reddit.com/r/cpp/comments/100vctp/msvc_vs_clang_performance_has_anyone_tested/).

The networking stack isnt what you say it is really. Windows exposes a bsd api for sockets (and thus, networking) to be more or less compatible with earlier iterations of berkeley's sockets (the api all networking on linux relies on). It has its own winapi for networking but the general backbone was still dependent on the old sockets behaviour. Sure, the differences in how each system does networking is different. But its trivial to separate these issues and fix them especially with the knowledge and libraries we have in 2026.

Don't take it as an attack for MSVC and saying how MSVC is a terrible compiler and to never ever ever use it.
I use it all the time, it does its job. But mingw is also a solid choice to compile nowadays especially with how easy it is to crosscompile with it.

PR #1655, opened 2026-09-18
feat: implement new modal components

nyukuru:
There's 3 new modal components not yet in DPP that I added:

- Radio Group
- Checkbox Group
- Checkbox

Radio group and checkbox group are very similar to the selection menus and similarly take an `options` field. I decided to make the options field a `std::variant` because of this which slightly breaks the api for those using the field directly. 

With the checkbox group, the value of `value` in the response can now be a boolean type so that logic was also added

## Code change checklist


include/dpp/message.h
```diff
@@ -584,11 +684,9 @@ class DPP_EXPORT component : public json_interface<component> {
 	int32_t max_length;
 
 	/**
-	 * @brief Select options for select menus.
-	 *
-	 * @warning Only required and available for select menus of type dpp::cot_selectmenu
+	 * @brief Options to pick for select menus or group menus.
 	 */
-	std::vector<select_option> options;
+	std::variant<std::vector<select_option>, std::vector<group_option>> options;
```
braindigitalis: this is a breaking change for existing bots. is there a way we can represent it without breaking the API, perhaps by using the same type to represent both? I'll leave that one for you to think about.

braindigitalis (changes requested):

braindigitalis:
thanks for your prs! this and your last one are looking good. just a few small changes to consider :)

braindigitalis:
merging your other pr has introduced a conflict you'll have to manually resolve

nyukuru:
I collapsed `select_option` and `group_option` into one `choice_option` now (I figured leaving it as `select_option` might be confusing going forward when used in the context of a group option). And just kept a `select_option` as an alias so it doesn't break the API anymore. I hope that implementation is fine.

braindigitalis:
> I collapsed `select_option` and `group_option` into one `choice_option` now (I figured leaving it as `select_option` might be confusing going forward when used in the context of a group option). And just kept a `select_option` as an alias so it doesn't break the API anymore. I hope that implementation is fine.

yeah that will be fine, I'll approve it in a few, t hanks!

braindigitalis (approved):

